PR #44, opened 2026-06-02
Add Codex Computer Use runtime setup

HomunculusLabs:
## Summary
- add first-class Codex Computer Use setup/status UX and runtime launchability checks
- resolve Codex app-managed Computer Use plugin `.mcp.json` into one-turn app-server MCP overrides
- add opt-in Codex app-server diagnostics and method-aware timeout messages

## Validation
- `.agents/skills/rpce-contribution-check/scripts/preflight.sh push` against `upstream/main`
- `make dev-lint`
- `make dev-test`
- `make dev-swift-build PRODUCT=RepoPrompt`

Note: non-disruptive `make dev-smoke` could not run locally because RepoPrompt CE was not running/MCP-connectable.

PR #45, opened 2026-06-02
Harden signed test build workflow

baron:
## Summary
- add a signed-test source ref verifier that rejects PR/fork/malformed refs and only allows commits reachable from canonical upstream branches or tags
- generate and upload signed-test provenance JSON binding requested ref, resolved source commit, trusted tooling commit, workflow run URL, signing mode, reachable refs, and artifact hashes
- strengthen signed-test smoke validation, release-tooling regression tests, and release docs

## Validation
- `python3 Scripts/test_release_tooling.py ReleaseToolingTests.test_signed_test_build_ref_helper_requires_upstream_reachability ReleaseToolingTests.test_signed_test_build_ref_helper_rejects_pr_and_fork_refs ReleaseToolingTests.test_signed_test_build_docs_describe_reachability_and_provenance ReleaseToolingTests.test_release_workflows_isolate_executable_helper_smoke_and_harden_p12_cleanup`
- `bash -n Scripts/release.sh Scripts/verify_signed_test_build_ref.sh`
- `python3 -m py_compile Scripts/test_release_tooling.py`
- `actionlint .github/workflows/signed-test-build.yml`

## Notes
- Full `Scripts/test_release_tooling.py` currently hangs in existing `test_legacy_sparkle_key_export_is_rejected`; targeted coverage for this change passes.

### DIFF
--- a/.github/workflows/signed-test-build.yml
+++ b/.github/workflows/signed-test-build.yml
@@ -1,0 +1,368 @@
+name: Signed Test Build
+
+on:
+  workflow_dispatch:
+    inputs:
+      source_ref:
+        description: Upstream branch, tag, or upstream-reachable SHA to build and sign
+        required: true
+        default: main
+        type: string
+
+concurrency:
+  group: signed-test-build
+  cancel-in-progress: false
+
+permissions:
+  contents: read
+
+jobs:
+  validate-ref:
+    name: Validate Test Build Ref
+    if: github.ref == 'refs/heads/main'
+    runs-on: macos-26
+    outputs:
+      commit: ${{ steps.source-ref.outputs.commit }}
+      reachable_refs: ${{ steps.source-ref.outputs.reachable_refs }}
+      tooling-commit: ${{ steps.tooling-ref.outputs.commit }}
+    steps:
+      - name: Require trusted workflow dispatch
+        env:
+          SOURCE_REF: ${{ inputs.source_ref }}
+        run: |
+          set -euo pipefail
+          [[ -n "$SOURCE_REF" ]] || { echo "source_ref is required" >&2; exit 1; }
+          [[ "$SOURCE_REF" != -* ]] || { echo "source_ref must not start with '-'" >&2; exit 1; }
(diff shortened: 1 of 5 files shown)


najibninaba:
I ran a RepoPrompt review pass on this PR, then checked the flagged areas manually against the workflow and a couple of small git fixtures. I think the overall shape is good, especially the no-secrets staging job and the protected signing job, but I’d tighten a few ref/input edges before merge.

First, reachability is only checked in `validate-ref`. The protected `sign` job may wait for environment approval, then sign `needs.validate-ref.outputs.commit` later. If the branch or tag that made that commit reachable is deleted or force-pushed during that gap, the job can still sign the old SHA. I’d re-run `verify_signed_test_build_ref.sh` in the protected job before signing, compare the fresh commit to `needs.validate-ref.outputs.commit`, and use the refreshed reachable refs in provenance.

Second, `verify_signed_test_build_ref.sh` mostly validates whatever `actions/checkout` left at `HEAD`. I reproduced a local case where `source_ref=main~1` is accepted when HEAD is already checked out to that commit. That means a requester can enter a commit-ish that is not actually a branch, tag, or full SHA, and the helper will still accept it if checkout already landed on a reachable commit. I’d resolve `SOURCE_REF` explicitly as an exact upstream branch, exact tag, or full SHA; reject revspec syntax and branch/tag ambiguity; then compare the resolved commit to HEAD.

One smaller workflow issue: the smoke job interpolates `inputs.source_ref` directly into inline Python. GitHub’s Actions security docs recommend passing untrusted input through env vars or action inputs instead, and Git branch names can contain quotes. Passing the expected ref through env/argv would avoid syntax breakage or injection-shaped surprises.

I also ran the focused release tooling tests locally on the PR branch: `python3 -m unittest Scripts/test_release_tooling.py` passes. I did not run the workflow itself in GitHub Actions, since I don’t think I have access to dispatch this protected signing path. So this is based on the PR diff, RepoPrompt review, local git fixtures, and the focused release-tooling tests.

baron:
@najibninaba thank you for the feedback and review. I've addressed it in [216e272](https://github.com/repoprompt/repoprompt-ce/pull/45/commits/216e2726e7c321c38857b054b134816586a5e7c7). 

If it's still lacking feel free to address this with a better PR or modify the associated branch.

@provencher if you have time please review it as well. The ability to create test builds will unblock PR #44

PR #48, opened 2026-06-03
Fix Markdown table border clipping

### LINKED ISSUE #28: Fix clipped Markdown table borders in Agent Mode and Oracle chat
## Problem
Some Markdown tables can render with clipped or incomplete borders in chat transcripts. The issue appears in both Agent Mode and Oracle chat because both use the shared TextKit-backed Markdown renderer.

## Existing foundation
- `EnhancedMarkdownCompiler.renderTextTable(...)` builds `NSTextTable` / `NSTextTableBlock` output with collapsed `0.5pt` borders.
- Large tables already fall back to plain text.
- `EnhancedMarkdownView` applies a small table-specific height allowance and invalidates cached measurements after width changes.

## Likely gap
The fixed measurement allowance can still be insufficient when border painting and width-sensitive cell wrapping change the final layout height.

## Acceptance criteria
- [x] Multi-row tables keep all outer and inner borders visible in Agent Mode and Oracle chat.
- [x] Narrow/wide resize cycles, wrapped cells, larger font presets, and neighboring text do not clip the final row, bottom edge, or right edge.
- [x] Large-table plain-text fallback continues to work.
- [x] Add a regression fixture or focused rendering test for the shared renderer.

HomunculusLabs:
## Summary
- Fix TextKit-backed Markdown table measurement so table borders have enough horizontal/vertical drawable space in shared chat renderers.
- Keep large Markdown tables on the existing plain-text fallback path.
- Add focused Markdown/TextKit regression coverage for wrapped table geometry, table/non-table reuse, drawing bounds, and fallback thresholds.

## Validation
- `make dev-test FILTER=MarkdownTextTableMeasurementTests` — 7 tests passed
- `make dev-format`
- `make dev-lint`
- `.agents/skills/rpce-contribution-check/scripts/preflight.sh push` — original outgoing validation included guardrails, secret scan, `make dev-lint`, full `make dev-test`, and `make dev-swift-build PRODUCT=RepoPrompt`; rerun before upstream push also passed on the clean branch
- Manual CE debug app validation: Oracle/chat tables, Agent Mode tables, resize cycles, larger content wrapping, and large-table fallback (300 columns & 50001 characters verified)


HomunculusLabs:
This fixes [Issue #28](https://github.com/repoprompt/repoprompt-ce/issues/28 )

baron:
Thanks for tackling this. I did a local pass with the table stress case from the screenshot and wanted to leave the notes here in case they help guide the final shape of the fix.

The border fix itself is real, but I want to echo @provencher’s caution about the scroll view / VStack / markdown render-area layering. In a local experiment I was able to make the clipped borders go away while also making transcript scrolling noticeably choppy, which lines up with the concern that measurement or invalidation in the wrong place can feed back through SwiftUI layout.

The lower-churn direction that looked safest locally was:

- Keep a small visible inset/height allowance so collapsed `NSTextTable` outer borders have room to draw.
- Avoid doing height measurement from `setFrameSize`; that path can run during scroll/layout reconciliation. It should only keep geometry/cache state coherent and let `sizeThatFits` own height measurement.
- Preserve the content-version + width cache fast path so unchanged markdown at an unchanged width does not repeatedly call TextKit layout during scrolling.
- If table measurement needs to be isolated from the live rendered view, reuse a detached TextKit measurement stack keyed by content/version/width rather than allocating a fresh `NSTextView` or touching the live layout manager on every miss.

Local validation only: focused markdown table measurement tests passed, the actual debug build looked good at wide and narrow widths on the repro text, and an 8s sample during scripted scrolling showed table measurement was not a dominant stack and `setFrameSize` was not a hot measurement path.

I have not pushed any code because I want to respect the authorship of this PR. Happy to submit a follow-up patch or share the local diff if you think that would be useful.

moonray:
Visually the fix seems to work. But some automated code review found some edge cases and a test that needs to be improved:

1. The headline border-clip test can’t fail. testTextTableDrawingRectIncludesRightAndBottomBorderAllowanceInsideViewBounds asserts drawingRect.maxY + 0.5 ≤ bounds.maxY + 0.5 and the maxX analog. aBut bounds.height = ceil(usedRect.height) + 6 and the container is set to width − 2, so by construction usedRect.maxY ≤ ceil(usedRect.height) and usedRect.maxX ≤ width − 2 < width − 1. Both assertions hold regardless of whether the border is actually painted correctly — they’re invariants of the chosen constants, not the painted output. The test should assert against the actual lowest NSTextTableBlock enclosing-rect maxY + 0.5pt border, independent of the inset/allowance constants.

2. Constant-tuned, not measured. 1pt inset + 6pt allowance work for the tested widths (220–520) and fonts (16/22), and the author verified larger presets manually — but there’s no guard against an extreme font preset where the border/padding-to-glyph ratio shifts.

3. Rapid-resize transient. The stale-cache-on-shrink path is mitigated (geometry now resyncs inside measuredHeight) but not eliminated — a one-frame clip during a fast narrow↔wide drag is still possible because SwiftUI proposes sizes asynchronously. The issue’s “resize cycles do not clip” criterion is met in practice, not provably in code.

baron (changes requested):
Blocking on exact head `ef7ead24b9dcce58e0193acc1aa4728b3c8f825b`.

The branch conflicts with the current Markdown renderer, and the headline drawing-rect regression is tautological: its bounds are derived from the same fixed border allowances being asserted, rather than measuring the painted table block at the right/bottom boundary. Rebase onto the current renderer and replace the constant-derived assertion with a rendering/table-block-boundary regression that fails on the original clipping behavior, then rerun CI.

morluto:
### Audit disposition — **refresh/rebase required** (2026-08-14)

The underlying rendering fix is reasonable, but this branch is stale against the current Markdown/TextKit implementation. I would preserve the work rather than close it, then rebase and revalidate it on the current renderer.

The main review risk is that table-specific drawable insets can affect the containing mixed Markdown view, not only the table. Please cover prose + table content, multiple tables, wrapping/resizing, scrolling and selection, accessibility, and the existing large-table fallback on the refreshed head. I do not see a reason to merge the conflicted branch as-is.

morluto:

**Disposition: refresh and revalidate before merge.** The table-specific geometry change is plausible and the focused/manual coverage is useful, but this is a shared Markdown/TextKit renderer. Please rebase onto current `main` and verify that mixed prose+table content, multiple tables, wrapping, selection, scrolling, and accessibility do not inherit unintended container padding or measurement changes from the table path. I did not find a fundamental objection to the fix once that current-head integration evidence is green.

PR #58, opened 2026-06-03
Add Dock New Window menu item

### LINKED ISSUE #57: Add “New Window” action to Dock menu
## Summary

RepoPrompt CE should offer a Dock right-click menu action for opening a new main window.

## Problem

Users can currently open a new window through File → New Window, but the same action is not available from the macOS Dock menu. This makes the app feel less native and adds friction when users are interacting with RepoPrompt CE primarily from the Dock, especially when working with multiple virtual desktops.

## Desired Behavior

Right-clicking the RepoPrompt CE Dock icon should show a `New Window` item that behaves the same as File → New Window.

## Acceptance Criteria

- [ ] Dock right-click menu includes `New Window`
- [ ] Selecting it opens a new RepoPrompt CE main window
- [ ] Behavior matches File → New Window by using the existing SwiftUI main-window path
- [ ] The action does not introduce a separate AppKit window lifecycle
- [ ] Focused test coverage verifies the Dock menu item wiring

## Validation

- Unit coverage for the Dock menu item
- Manual verification: File → New Window and Dock → New Window both open equivalent main windows

dah:
Closes #57

### DIFF
--- a/Sources/RepoPrompt/App/AppDelegate.swift
+++ b/Sources/RepoPrompt/App/AppDelegate.swift
@@ -18,6 +18,7 @@ import SwiftUI
 class AppDelegate: NSObject, ObservableObject, NSApplicationDelegate {
     /// Prevents re-entrant termination (Cmd+Q twice, menu + dock quit, etc.)
     private var terminationInProgress = false
+    private let dockMenuController = DockMenuController()
 
     // New global routing/settings services (kept alive by the AppDelegate)
     private var windowRoutingService: WindowRoutingService?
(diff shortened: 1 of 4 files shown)


baron (commented):
I found one behavior gap in the Dock menu flow: the new-window action works, but should also bring RepoPrompt forward when invoked while the app is inactive.

Sources/RepoPrompt/App/AppDelegate.swift
```diff
@@ -164,6 +176,16 @@ class AppDelegate: NSObject, ObservableObject, NSApplicationDelegate {
         return .terminateLater
     }
 
+    @objc private func openNewWindowFromDockMenu(_ sender: NSMenuItem) {
+        do {
+            try AppWindowOpener.shared.openMainWindow()
```
baron: When this Dock menu action is invoked while RepoPrompt is inactive, `openMainWindow()` can create/request the SwiftUI window without activating the app, so the new window may remain behind the currently focused app. Native Dock menu commands like Safari's `New Window` and Finder's `New Finder Window` bring the app/new window forward, and choosing a Dock menu item is a foreground user action. Please activate RepoPrompt after opening, and ideally make the created window key/order front via the existing window-opening path if available. The current test only verifies that the menu item exists/enabled, so it would not catch this focus regression.

dah: good catch; will try to push this soon

provencher (dismissed):
Rebased onto current main, reviewed through Agent Mode, hardened startup and nil-sender Dock action routing, and validated with formatter, strict lint, full tests, RepoPrompt build, guardrails, secret scans, and green hosted CI.

provencher (approved):
Addressed the outstanding foreground-activation review thread, preserved queued startup handling, reran full local validation, and confirmed hosted CI passes on the final head.

PR #73, opened 2026-06-05
Fix Keyboard Shortcuts settings crash

### LINKED ISSUE #71: Settings > Keyboard Shortcuts crashes the app
Open the settings, go to keyboard shortcuts and the app crashes ... for me it was 100% of the time.

strickvl:
## Summary
- patch KeyboardShortcuts 2.3.0 during packaging so localized recorder resources resolve from the signed `Contents/Resources` bundle
- add a shared SwiftPM resource-bundle validator that checks bundle markers, app-root cleanliness, and the packaged executable patch marker
- wire the validator into package/staged-release checks and add release-tooling regression tests

Closes #71

## Validation
- `bash -n Scripts/package_app.sh Scripts/validate_staged_release.sh Scripts/validate_required_swiftpm_resource_bundles.sh Scripts/patch_keyboard_shortcuts_resource_lookup.sh Scripts/release.sh`
- focused release-tooling unit tests via `uv run --python 3.12 -m unittest ...` — 9 tests OK
- `RELEASE_ALLOW_ADHOC_SIGNING=1 ./Scripts/package_app.sh release`
- `codesign --verify --deep --strict --verbose=2` passed during packaging
- embedded MCP helper layout + `--version` smoke passed
- Oracle review loop reached “happy / no remaining blockers” after hardening the patch marker and dependency revision guard

## Manual verification
- ran the app locally
- confirmed the packaged app keeps `KeyboardShortcuts_KeyboardShortcuts.bundle` under `Contents/Resources`
- confirmed no `KeyboardShortcuts_KeyboardShortcuts.bundle` is present at the app bundle root
- opened Settings → Keyboard Shortcuts
- confirmed the recorder controls render and the app no longer crashes


baron (commented):
Non-blocking follow-up: the packaged debug app path is working and the crash is resolved, so I would not hold this PR for this.

One small hardening gap: the patch helper and packaged-app validator both key off the `RepoPromptKeyboardShortcutsResourceLookupV1` marker. I verified with a fixture that a stale/partial `KeyboardShortcuts` checkout containing that marker but still using `NSLocalizedString(..., bundle: .module, ...)` would be treated as already patched, and a marker-only packaged executable would pass the validator.

Could be worth strengthening the guardrail in a follow-up by checking for the actual patched lookup, e.g. the `.keyboardShortcutsResourceBundle` `NSLocalizedString` call, rather than only the marker.

baron (approved):
Approved. I verified the packaged debug app launches and the Keyboard Shortcuts settings crash is resolved. The remaining marker-check hardening note is non-blocking follow-up.

PR #74, opened 2026-06-06
Fix stalls in long Agent Mode and Context Builder runs

provencher:
## What this fixes

Long Agent Mode and Context Builder runs could occasionally stop making visible progress, finish with missing output, or leave an MCP caller waiting indefinitely.

This PR fixes the underlying lifecycle and routing races instead of adding timeouts. Long-running work can continue normally, while cancellation still returns promptly when requested.

## What changed

- MCP no longer silently drops messages when a connection is overloaded. The affected connection now fails clearly without disrupting other connections.
- Agent runs now carry generation and ownership information so events from an older run cannot update a newer one.
- Completion is published only after buffered assistant text, tool state, persistence, and MCP results agree on the same final revision.
- Session routing fails safely when the same MCP session appears to map to multiple live tabs.
- Context Builder cancellation no longer waits for slow provider cleanup, and large streamed results are accumulated without repeatedly rebuilding the full preview.
- Long conversations use narrower UI refreshes and less expensive replay serialization.
- The ask-user wizard behavior from Classic is included because it participates in the repaired run lifecycle.

## User impact

- Fewer Agent Mode and Context Builder stalls during long or tool-heavy runs.
- No missing assistant tail when a provider finishes quickly after its final streamed output.
- More reliable cancellation and MCP wait behavior.
- Better responsiveness in long conversations.
- More useful diagnostics when a transport or lifecycle failure does occur.

## Validation

Passed:

- repository guardrails and secret scanning
- SwiftFormat and strict SwiftLint
- full test suite
- provider package tests
- `RepoPrompt` and `repoprompt-mcp` builds
- live CE MCP smoke test


baron (dismissed):
**Summary**: Reviewed PR #74 (`codex/harden-agent-mode-mcp-lifecycle` vs `main`) with RepoPrompt's `rp-review` workflow and follow-up source checks. The lifecycle hardening direction looks good, but I found three must-fix issues around cleanup semantics, attachment reservations, and stale Codex turn completion.

**Must-Fix**
- `Sources/RepoPrompt/Features/AgentMode/Runtime/AgentModeRunService.swift:935`: `waitForCleanup` is discarded, while `Sources/RepoPrompt/Features/AgentMode/Runtime/AgentRunTerminalCommitBarrier.swift:208` always runs provider teardown in a detached task. Callers like MCP `agent_run` cancellation still pass `waitForCleanup: true`, but can now proceed before provider/controller shutdown completes. Preserve the contract by awaiting teardown when requested or returning a teardown handle.
- `Sources/RepoPrompt/Features/AgentMode/Runtime/Codex/CodexAgentModeCoordinator.swift:3267`: terminalized Codex send failures/cancellations call `finalizeCodexRun` without forwarding `attachmentReservationID`; `finalizeCodexRun` then builds the barrier request without a reservation. Reserved attachments can be left un-restored/un-cleaned after failed startup/send/cancel paths. Thread `attachmentReservationID` through `finalizeCodexRun` into the terminal barrier.
- `Sources/RepoPrompt/Features/AgentMode/Runtime/Codex/CodexAgentModeCoordinator.swift:4427`: stale non-nil `turnCompleted` events are accepted whenever `session.codexCurrentTurnID` is nil. A delayed completion from an older Codex turn can finalize a newer active run before its `turnStarted` arrives. Correlate completions to the current run/turn generation, or reject unknown non-nil turn IDs for active runs.

**Suggestion**
- `Tests/RepoPromptTests/AgentMode/AgentModeRunServiceLifecycleTests.swift:165`: add regression coverage for awaited cleanup, stale `turnCompleted`, and Codex failed-send attachment reservation cleanup.

I did not run the Swift test suite; this was a review pass only.

provencher:
Thanks for the review, @baron. I pushed **f99dc65** to address the two confirmed lifecycle issues.

### What changed

- **Cancellation cleanup semantics are now explicit.** The old `waitForCleanup` Boolean was replaced with named completion modes:
  - `.terminalPublished` returns after canonical terminal publication and synchronous provider detachment, keeping UI and MCP cancellation responsive.
  - `.terminalTeardownCompleted` additionally waits for the exactly-once provider/attempt teardown and is reserved for cleanup-sensitive paths.
- **Terminal teardown is tracked and awaitable.** A caller can await teardown even if another caller already published the terminal state, without creating a second teardown owner.
- **Codex turn completion now requires observed turn correlation.** Stale or mismatched non-nil `turnCompleted` events can no longer terminalize a newer run or consume its pending turn kind. Nullable turn IDs remain supported when a corresponding start was observed.

### Deliberately unchanged

I did not add attachment-reservation threading in this patch. The existing wildcard behavior already restores attachments correctly, and tightening that contract would require broader state changes unrelated to the two confirmed defects.

### Validation

- Full coordinated test suite passed
- Lifecycle tests: 15/15
- Codex liveness tests: 15/15
- Stop-target tests: 13/13
- Workspace cleanup tests: 4/4
- Worktree tests: 18/18
- SwiftFormat/SwiftLint passed
- RepoPrompt product build passed
- Push preflight, guardrails, and secret scan passed

No visible app lifecycle actions were performed.

baron (approved):
Approved after checking the follow-up lifecycle fixes and green CI.

PR #85, opened 2026-06-07
Fix warm Agent Mode tab activation

provencher:
## Summary
- consume compose-tab activation from the post-storage, window-filtered notification instead of the pre-storage `@Published` emission
- preserve the existing explicit Agent Mode bootstrap and strict transcript ownership/generation guards
- add real `WindowState` regression coverage for immediate warm A→B, A→B→A, and cross-window isolation

## Root cause
`AgentModeViewModel` handled `$activeComposeTabID` during `willSet`, so the destination activation ran while `currentTabID` still identified the source. Strict ownership correctly rejected publication, and deduplication prevented a retry.

## Validation
- `make dev-test FILTER=AgentModeChatSwitchActivationTests` (3 tests)
- `make dev-lint`
- `make dev-test`
- contribution preflight for commit and push
- `make dev-smoke` attempted non-disruptively; blocked because the CE app was not running or MCP was disabled

### DIFF
--- a/Sources/RepoPrompt/Features/AgentMode/ViewModels/AgentModeViewModel.swift
+++ b/Sources/RepoPrompt/Features/AgentMode/ViewModels/AgentModeViewModel.swift
@@ -2034,11 +2034,17 @@ final class AgentModeViewModel: ObservableObject {
         guard let promptManager else { return }
         installPromptManagerCascadeResolvers(promptManager)
 
-        // Observe tab changes
-        promptManager.$activeComposeTabID
-            .removeDuplicates()
-            .sink { [weak self] tabID in
-                self?.onTabChanged(tabID)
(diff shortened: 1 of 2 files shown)


baron (approved):
Review complete. No blocking findings.

Validation:
- `git diff --check origin/main...HEAD` passed.
- `make dev-test FILTER=AgentModeChatSwitchActivationTests` passed (3 tests, 0 failures).
- RepoPrompt review CLI/context builder pass completed with no must-fix findings.
- Repo contribution preflight commit-mode guardrails passed before this GitHub-visible review action.

Non-blocking suggestion:
- Consider follow-up coverage for the non-replaying notification edge cases: activating Agent Mode after `activeComposeTabID` is already set, and cold persisted-session loading after a tab switch. The PR already covers warm switches, back-to-back warm switches, and window-scoped notification behavior.

PR #89, opened 2026-06-07
Fix branch popover scroll containment

strickvl:
## Summary
- Give the Agent Mode branch switch popover a single scaled root height so loaded branch rows stay inside the popover boundary.
- Let the branch row list scroll within the remaining popover space instead of using an independently estimated scroll height.
- Use a regular `VStack` inside the finite branch list scroll view so the scrollbar range is measured up front and does not shrink/recalibrate after the first scroll.

Additional notes:
- `make dev-test` initially had one unrelated failure in `UnixSocketMCPTerminalCleanupTests`; rerunning `make dev-test FILTER=UnixSocketMCPTerminalCleanupTests` passed, and push preflight's full root test run passed afterward.


provencher (approved):
Rebased onto current main, reviewed with independent Agent Mode passes, hardened worktree classification and popover containment edge cases, and validated with formatter, strict lint, full tests, RepoPrompt build, guardrails, secret scans, and green hosted CI.

PR #93, opened 2026-06-07
Fix onboarding Skip All routing

fitchmultz:
## Summary
- Fix **Skip All** on the first-launch onboarding screen.
- Keep the existing behavior for users who open the setup guide from an active workspace.

## Why
On first launch, RepoPrompt is still in the system fallback workspace. The old **Skip All** route tried to continue into the main app, but that route is intentionally blocked when no real workspace is active, so the onboarding screen appeared to do nothing.

This changes onboarding completion to do the obvious thing in both cases:

- first launch / no workspace yet: show the workspace chooser
- real workspace already active: return to the main app

## Validation
- `make guardrails`
- `make dev-lint`
- Manual: clicked **Skip All** from the onboarding screen and confirmed it leaves onboarding.

### DIFF
--- a/Sources/RepoPrompt/App/ViewModels/ContentViewModel.swift
+++ b/Sources/RepoPrompt/App/ViewModels/ContentViewModel.swift
@@ -124,6 +124,24 @@ class ContentViewModel: ObservableObject {
         }
     }
 
+    /// Completes onboarding from either first launch or the manually opened setup guide.
+    ///
+    /// First launch starts in the system fallback workspace, so there is no main workspace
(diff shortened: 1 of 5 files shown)


provencher (dismissed):
Reviewed the onboarding routing change. Local guardrails, lint, full tests, and RepoPrompt product build pass; GitHub Style and Secret Scan are green.

provencher (approved):
Validated locally and in CI. Onboarding exits consistently record completion before routing, the overflow test is stabilized without weakening ownership assertions, and all required checks pass.

PR #95, opened 2026-06-07
Add expanded file mention picker style

### LINKED ISSUE #90: Add Compact/Expanded style for file @ mention pickers
## What to build

Add a global `@ File Picker Style` setting that preserves today’s compact file-mention picker by default and offers an opt-in expanded, VS Code-like file picker experience.

The current compact picker is good for quick, low-noise file tagging, but its 5-result limit makes it hard to choose the right file when many paths match the same query. Expanded mode should solve that discovery problem without replacing the compact default for users who prefer the existing minimal popup.

This should be delivered as one end-to-end vertical slice: JSON-backed setting, Settings UI, shared picker configuration, both file `@` picker implementations, overlay sizing/readability, and focused tests.

## User story

As a RepoPrompt user tagging files with `@`, I want an optional expanded file picker that shows many more matching files in a wider, scrollable list, so I can quickly find the right file when many paths match my query without losing the current compact default behavior.

## Acceptance criteria

- [ ] Settings exposes an always-visible Text Editing control labeled `@ File Picker Style` with `Compact` and `Expanded` options.
- [ ] `Compact` is the default for missing/invalid settings and preserves current behavior: 5 returned file results, 5 visible rows, current file-picker width, and current row content.
- [ ] `Expanded` returns up to 99 file results and shows up to 15 visible rows before scrolling.
- [ ] `Expanded` widens file `@` picker overlays to about 480pt for readability.
- [ ] `Expanded` shows parent/root path subtitles for file rows; Compact preserves current subtitle/content behavior.
- [ ] The setting applies to both file `@` picker paths: the older prompt/instructions mention picker and the newer Agent Mode file-tag picker.
- [ ] Slash-skill suggestions are unchanged.
- [ ] Existing insertion/attachment semantics are unchanged; subtitle/path display changes are display-only.
- [ ] Existing keyboard navigation behavior is unchanged, including wrap-around and auto-scroll-to-highlight.
- [ ] Existing bare-`@` behavior is unchanged; Expanded does not eagerly list arbitrary workspace files before the current picker would have candidates/results.
- [ ] In the older prompt picker, the synthetic `Selected` folder continues returning all selected files; the mode only controls viewport sizing/readability.
- [ ] In the older prompt picker, Expanded applies consistently to root, search, direct folder children, and folder drill-down child overlay levels.
- [ ] Agent Mode keeps its existing candidate breadth formula, so Expanded may search roughly `99 * 8` candidates before scoring.
- [ ] Settings changes are required to apply on next picker refresh/open; instant active-popup update is optional, not required.
- [ ] Tests cover settings persistence/defaulting/invalid raw values and focused mode-to-config behavior.

## Out of scope

- Changing file search or ranking algorithms.
- Changing slash-skill suggestion behavior.
- Changing insertion, attachment, or token semantics for committed mentions.
- Changing bare-`@` behavior to eagerly list arbitrary workspace files.
- Live-updating an already-open picker immediately when the setting changes; applying on the next picker refresh/open is sufficient.
- Adding standalone documentation beyond Settings helper copy.

## Implementation notes

- Add a small shared file-mention picker mode/configuration type, colocated with mention UI/model code, as the single source of truth for mode-derived values:
  - Compact: result cap 5, visible rows 5, current file-picker width, no new path subtitles.
  - Expanded: result cap 99, visible rows 15, width about 480pt, parent-path subtitles for files.
- Persist only the mode raw value in global JSON settings, likely under `GlobalScalarPreferences.UISettings`.
- Thread the derived config into:
  - `MentionSuggestionService` / `MentionCoordinator` / `AttributedTextKitView` for the older prompt picker.
  - `AgentFileTagSuggestionService` / `FileTagMentionHelper` / `ResizableTextFieldFeatures` / `AgentInputBar` for Agent Mode file tags.
  - `MentionOverlayController` and `MentionSuggestionView` for visible row cap, scroll indicators, and width.
- Keep result cap, visible row cap, width, and subtitle mode separate concepts in code, even though they are derived from one mode setting.

## Validation suggestions

- Run focused settings/config tests after implementation.
- Run the smallest relevant coordinated app build/test command after Swift/UI changes.
- If manual app validation is safe, verify:
  - Compact still shows the current 5-result behavior.
  - Expanded returns many matching file results, shows 15 visible rows, scrolls beyond that, and is wider/readable.
  - Both file picker paths honor the setting.
  - Slash-skill suggestions are unchanged.

aAtila:
## Summary

Users can now opt into an expanded file mention picker that shows more results and includes path subtitles to disambiguate files with the same name. The compact picker remains the default.

## Changes

- New `FileMentionPickerStyle` setting (compact / expanded) persisted in global settings and surfaced in Appearance preferences.
- Expanded mode widens the picker, increases visible results, and adds path-based subtitles.
- Mention popup frames clamped to visible screen bounds so the wider picker doesn't overflow off-screen.
- Native NSWindow shadow disabled on the rounded picker popup to eliminate rectangular halo artifacts.
- Active mention sessions reset when picker configuration changes.

Fixes https://github.com/repoprompt/repoprompt-ce/issues/90

### DIFF
--- a/Sources/RepoPrompt/Features/AgentMode/Views/AgentInputBar.swift
+++ b/Sources/RepoPrompt/Features/AgentMode/Views/AgentInputBar.swift
@@ -280,6 +280,7 @@ struct AgentComposerView: View, Equatable {
     @State private var modelMenuSnapshotReleaseTask: Task<Void, Never>? = nil
 
     @ObservedObject private var fontScale = FontScaleManager.shared
+    @ObservedObject private var globalSettings = GlobalSettingsStore.shared
     private var fontPreset: FontScalePreset {
         fontScale.preset
     }
@@ -570,6 +571,7 @@ struct AgentComposerView: View, Equatable {
                     fileTagLookupContextProvider: { [tabID = props.currentTabID] in
                         await actions.agentWorkspaceLookupContext(tabID)
                     },
+                    fileMentionPickerConfiguration: globalSettings.fileMentionPickerConfiguration(),
                     onFileTagCommitted: handleFileTagCommitted(_:),
                     slashSkillSuggestionsProvider: { query in
                         await actions.slashSkillSuggestions(query)
--- a/Sources/RepoPrompt/Features/Settings/Models/GlobalSettingsDocument.swift
+++ b/Sources/RepoPrompt/Features/Settings/Models/GlobalSettingsDocument.swift
@@ -165,6 +165,7 @@ struct GlobalScalarPreferences: Codable, Equatable {
         var collapseLatestFileChanges: Bool?
         var showTooltips: Bool?
         var experimentalAttributedTextEditor: Bool?
+        var fileMentionPickerStyle: String?
         var enableKeyboardShortcuts: Bool?
         var fontScaleBodySize: Double?
 
@@ -174,6 +175,7 @@ struct GlobalScalarPreferences: Codable, Equatable {
             collapseLatestFileChanges: Bool? = nil,
             showTooltips: Bool? = nil,
             experimentalAttributedTextEditor: Bool? = nil,
+            fileMentionPickerStyle: String? = nil,
             enableKeyboardShortcuts: Bool? = nil,
             fontScaleBodySize: Double? = nil
         ) {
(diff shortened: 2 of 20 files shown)


provencher (approved):
Maintainer review complete. Rebased, hardened, pair-cleaned, and validated locally and in hosted CI.

PR #116, opened 2026-06-09
Model picker missing Z.ai options at startup

### LINKED ISSUE #115: Model picker doesn't show CC Z.ai options when key is preconfigured
## Bug

When CC Z.ai is already configured and the API key is saved in Settings, the Z.ai models do not appear as options in the model picker after app startup. A manual state change (e.g., toggling the key off and on in Settings) is required before the models show up.

## Root Cause

`APISettingsViewModel.publishClaudeCodeGLMAvailability()` was short-circuiting with `guard didChange else { return }` when the persisted `configured` mirror flag was already `true`. During startup, `loadStoredData()` loads the ZAI key into memory and updates `compatibleBackendSecretPresence`, but the persisted flag does not change — so the notification was never posted, and model-picker view models never refreshed.

## Fix

Always post `.claudeCodeGLMAvailabilityChanged` after recomputing the in-memory GLM/Z.ai secret state, regardless of whether the persisted configured mirror changed.

### Files Changed
- `Sources/RepoPrompt/Features/Settings/ViewModels/APISettingsViewModel.swift` — remove early return guard, always post notification
- `Tests/RepoPromptTests/Prompt/PromptAgentAvailabilityRefreshTests.swift` — new regression test

## Reproduction
1. Configure CC Z.ai with a valid API key in Settings
2. Quit and relaunch the app
3. Z.ai models are missing from the model picker
4. Expected: Z.ai models should be available immediately after startup

moonray:
## Summary

Fixes #115

When CC Z.ai was preconfigured with a saved API key, the model picker in IDE mode did not show Z.ai models after app launch. The notification `.claudeCodeGLMAvailabilityChanged` was only posted when the persisted `configured` flag changed value. At startup, this flag was already `true`, so the notification was skipped and model pickers never refreshed.

## Changes

- **`APISettingsViewModel.swift`** — Always post `.claudeCodeGLMAvailabilityChanged` after recomputing the in-memory GLM/Z.ai secret state, regardless of whether the persisted configured mirror changed.
- **`PromptAgentAvailabilityRefreshTests.swift`** — New regression test that verifies `PromptViewModel.availableAgentKinds` includes `.claudeCodeGLM` after a preconfigured ZAI key loads during startup.

## Testing

```
make dev-test FILTER=PromptAgentAvailabilityRefreshTests
```


baron:
Non-blocking follow-up from a local review of this fix: the startup bug is real and this PR fixes the missing refresh, but `always post` looks a bit broader than necessary.

The more targeted pattern would be to move change detection into `APISettingsViewModel.publishClaudeCodeGLMAvailability()` using the in-memory secret-presence mirror, not just the persisted configured flag:

```swift
let previousSecretPresent = compatibleBackendSecretPresence[.glmZAI] ?? false
let configuredDidChange = ClaudeCodeGLMIntegration.setConfigured(hasStoredZAIKey)
compatibleBackendSecretPresence[.glmZAI] = hasStoredZAIKey
guard configuredDidChange || previousSecretPresent != hasStoredZAIKey else { return false }
NotificationCenter.default.post(name: .claudeCodeGLMAvailabilityChanged, object: nil)
return true
```

That still fixes launch-time staleness because startup is `nil/false -> true` in memory even when the persisted mirror was already true, while avoiding no-op observer churn on revalidation / unchanged saves. It also gives the `setConfigured` return value a useful role again.

Related cleanup suggestion: the `.glmZAI` `saveCompatibleBackendSecret` / `deleteCompatibleBackendSecret` paths currently can post `.claudeCodeGLMAvailabilityChanged` via the GLM helper and then immediately post the same notification again through `postCompatibleBackendAvailabilityChanged()`. If the helper returns whether it posted, the GLM paths can skip the generic post and avoid a back-to-back duplicate refresh.

Test suggestion: keep the startup regression, but make it assert “at least one refresh” rather than exactly one during load, and add a small save/delete-path regression that a GLM secret availability transition emits a single refresh. If adding a fake secure store for this, consider extracting the existing test double instead of adding another file-local copy.

I validated this direction locally with focused tests (`PromptAgentAvailabilityRefreshTests`, `SecureStorageRepairServiceTests`) and `make dev-lint`, but treating this as non-blocking since the current PR already addresses the user-visible startup miss.

baron:
Pushed a maintainer update on top of your fix (merged latest main + `5a504d7`). A review of the change surfaced that the unconditional post traded the startup-staleness bug for redundant observer refreshes, so the update keeps your fix's behavior on launch but restores deduplication:

- `publishClaudeCodeGLMAvailability()` now gates on in-memory secret presence in addition to the persisted flag — first load after launch still fires (presence goes nil → true), while no-op reloads/re-saves/failed validations stay quiet.
- `saveCompatibleBackendSecret`/`deleteCompatibleBackendSecret` use the publish result so GLM operations post exactly one `.claudeCodeGLMAvailabilityChanged` instead of two.
- Kept your startup-refresh test (with `assertForOverFulfill = false` so it asserts 'at least one refresh') and added a single-post-on-save test.
- Extracted the secure-store fake into a shared `TestSecureStorageBackend` test helper.

Both focused suites pass on the updated branch (`PromptAgentAvailabilityRefreshTests` 2/2, `SecureStorageRepairServiceTests` 7/7) plus SwiftFormat/SwiftLint clean. Thanks for the fix and the test!

baron (approved):
Reviewed in depth: original fix audited (7-angle review), notification logic tightened to in-memory change detection, double-post eliminated, tests extended (startup refresh + single-post save), secure-store fake shared. Focused suites and full local test suite pass on the merged head; CI green (two earlier failures were the known flaky concurrency tests, #161).

PR #128, opened 2026-06-09
Add ~/.opencode/bin to OpenCode CLI path hints

### LINKED ISSUE #127: OpenCode ACP connection fails: CLI not found (missing ~/.opencode/bin path hint)
## Summary

OpenCode ACP connections fail because the CLI binary at `~/.opencode/bin/opencode` cannot be found. The standard OpenCode install directory is missing from the universal fallback PATH hints.

## Reproduction

1. Open Settings → MCP → CLI Providers
2. Click "Test Connection" for OpenCode
3. Observe: `OpenCode CLI was not found as a valid executable regular file for 'opencode'. Install OpenCode or configure its absolute path.`

## Environment

- OpenCode CLI installed at `~/.opencode/bin/opencode` (standard install location via `curl -fsSL ... | bash`)
- RepoPrompt CE debug app launched via `open -n`

## Expected Result

OpenCode ACP session bootstrap completes successfully, model options are discovered.

## Actual Result

```
OpenCode CLI was not found as a valid executable regular file for `opencode`. 
Install OpenCode or configure its absolute path.
```

## Root Cause

`~/.opencode/bin` is not in `CLINativePathDefaults.userToolBins` — the universal fallback directory list used when building the enriched PATH. The directory is added to PATH by the user's login shell config (`.zshrc`/`.zprofile`), so `which opencode` works in a terminal. But when RepoPrompt resolves CLI paths:

1. The inherited PATH (from `open -n` via LaunchServices) is the minimal launchd PATH — no `~/.opencode/bin`
2. The login shell cache captures the full PATH, but `CommandPathResolver`'s interactive shell lookup can fail in the app context
3. `locateInSearchPaths()` falls back to the enriched PATH, which includes `userToolBins` directories like `~/.cargo/bin`, `~/go/bin`, `~/.bun/bin` — but NOT `~/.opencode/bin`

The fix is to add `~/.opencode/bin` to `userToolBins`, which is the same category as other standard CLI tool install directories (`~/.cargo/bin`, `~/go/bin`, `~/.bun/bin`).

Additionally, the `ACPAgentSessionController.bootstrap()` is missing the `notifications/initialized` notification required by the ACP protocol after the `initialize` handshake. Without it, OpenCode (and other MCP-derived ACP servers) silently ignores `session/new` requests.

moonray:
## Summary

Fixes #127 — OpenCode ACP connections fail due to two independent issues.

### 1. Add `~/.opencode/bin` to OpenCode provider-specific paths

**File:** `Sources/RepoPrompt/Infrastructure/Process/CLILaunchProfile.swift`

Adds `~/.opencode/bin` to `openCodeProviderSpecificPaths`, matching the standard OpenCode install directory. When `CommandPathResolver`'s interactive shell lookup fails in the app context (launched via `open -n`), this provides a direct fallback. Same pattern as Claude Code's `~/.claude/local`.

### 2. Send `notifications/initialized` after initialize response

**File:** `Sources/RepoPrompt/Infrastructure/AI/ACP/ACPAgentSessionController.swift`

The ACP protocol requires an MCP-style handshake:

```
Client → Server: initialize (id:1)
Server → Client: initialize response (id:1)
Client → Server: notifications/initialized  ← was missing
Client → Server: session/new (id:2)
```

Without this notification, OpenCode silently ignores `session/new` and the 30-second timeout fires.

### 3. Add strict ACP server test

**File:** `Tests/RepoPromptTests/AgentMode/ACPProviderSessionIdentityTests.swift`

A new test using a strict ACP server that only responds to `session/new` after receiving `notifications/initialized`. Verifies the correct method ordering: `initialize` → `notifications/initialized` → `session/new`.

## Testing

- `ACPProviderSessionIdentityTests`: 5/5 pass (including new strict initialized test)
- `OpenCodeACPLaunchResolverTests`: all pass

### DIFF
--- a/Sources/RepoPrompt/Infrastructure/Process/CLINativePathDefaults.swift
+++ b/Sources/RepoPrompt/Infrastructure/Process/CLINativePathDefaults.swift
@@ -38,8 +38,7 @@ enum CLINativePathDefaults {
         "~/bin",
         "~/go/bin",
         "~/.cargo/bin",
-        "~/.bun/bin",
-        "~/.opencode/bin"
+        "~/.bun/bin"
     ]
 
     static let defaultAdditionalPaths: [String] = orderedUnique(
--- a/Sources/RepoPrompt/Infrastructure/Process/CLINativePathDefaults.swift
+++ b/Sources/RepoPrompt/Infrastructure/Process/CLINativePathDefaults.swift
@@ -38,7 +38,8 @@ enum CLINativePathDefaults {
         "~/bin",
         "~/go/bin",
         "~/.cargo/bin",
-        "~/.bun/bin"
+        "~/.bun/bin",
+        "~/.opencode/bin"
     ]
 
     static let defaultAdditionalPaths: [String] = orderedUnique(
--- a/.agents/skills/rpce-test-quality/SKILL.md
+++ b/.agents/skills/rpce-test-quality/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: rpce-test-quality
-description: Select, design, review, consolidate, or delete RepoPrompt CE tests and diagnostics by regression value and maintenance cost. Use for bug fixes or features that may need coverage, test-adequacy reviews, XCTest or provider-package changes, runtime diagnostic harnesses, live or packaged smoke, core migrations, and decisions about whether a test is worth committing.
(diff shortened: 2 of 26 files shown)


Sources/RepoPrompt/Infrastructure/Process/CLINativePathDefaults.swift
```diff
@@ -38,7 +38,8 @@ enum CLINativePathDefaults {
         "~/bin",
         "~/go/bin",
         "~/.cargo/bin",
-        "~/.bun/bin"
+        "~/.bun/bin",
+        "~/.opencode/bin"
```
provencher: this is a cross model hint. there are better places to put this. It also shouldn't be required if your path is setup correclty.

PR #136, opened 2026-06-10
Import Classic RepoPrompt data into CE

yaananth:
## Summary
- add an optional Classic RepoPrompt import flow on the launch screen and Manage Workspaces when Classic app-support data exists
- import Classic workspaces, chat/agent sessions, global settings, workflow/model presets, saved prompts, custom workflows, UserDefaults-backed CLI/provider settings, and agent model selections
- keep default import no-Keychain/no-prompt after local testing showed Classic keychain ACLs can prompt repeatedly

## Validation
- make dev-format
- make dev-test FILTER=ClassicRepoPromptImportServiceTests
- make dev-lint
- ALLOW_ADHOC_SIGNING=1 make dev-build
- local Computer Use E2E: reset CE state, launched debug CE, imported 8 Classic workspaces and 74 settings items, verified no Keychain prompt, verified CLI Providers and Agent Models settings reflected imported config
- .agents/skills/rpce-contribution-check/scripts/preflight.sh commit

## Caveat
- push preflight was run, but the full root test step failed/canceled on unrelated Agent Mode timing failures and then hung in GitWorktreeContextSummaryTests. Focused import tests, build, lint, package, and local E2E all passed.


yaananth:
## Classic import screenshots

These are anonymized screenshots from local CE E2E verification. Workspace names and local paths are replaced with generic examples; provider/model UI is shown with empty secure fields.

**Launch screen import entry**

CleanShot 2026-06-10 at 07 59 30@2x


**Manage Workspaces import entry after startup**

CleanShot 2026-06-10 at 08 00 34@2x

masonjames:
I'd love to see this brought into core. Having the workspaces and associated chat sessions would be super-valuable.

I don't have repopromptce setup to run from the repo so can't test this, but it's something I looked for and thought would be included by default. Hope to see it merged!

baron:
## Maintainer handoff notes for PR 136

I did a deep pass on PR 136 with focused subagents plus an independent Review workflow. This is a promising importer, but I would not merge as-is. Below is the condensed handoff so the next maintainer can save time.

### Current status

- PR is behind/synced locally against current `upstream/main` without textual merge conflicts.
- Hosted checks currently show **Build and Test failing** while Style and Secret Scan pass.
- The local focused validation command `make dev-test FILTER=ClassicRepoPromptImportServiceTests` initially failed at compile time before tests could run.

### Blocking build issue

1. **Compile failure in `APISettingsViewModel`**
   - `Sources/RepoPrompt/Features/Settings/ViewModels/APISettingsViewModel.swift:557`
   - Calls `postCompatibleBackendAvailabilityChanged()`, but that symbol does not exist.
   - Recommended fix: remove the call rather than introducing a new notification. The current code path already has level-triggered availability via `agentAvailability`; `loadCompatibleBackendState(...)` refreshes availability, and `updateAvailableModels()` can remain.

### Must-fix import safety / correctness issues

2. **Dead / unwired Classic keychain reader**
   - `ClassicRepoPromptKeychainValueReader` is defined but never instantiated from production or tests.
   - Default import deliberately passes `sourceSecureValueReader: nil` / `targetSecureStore: nil`, so secure account import is unavailable in the actual UI path.
   - Recommendation: either delete the reader from this PR, or wire it only behind an explicit opt-in flow with clear Keychain prompt expectations and tests against a mocked `SecItemClient`.

3. **Provider/backend “configured” state can be imported without secrets**
   - The importer can bring over configured/connected booleans while the default path does not migrate Keychain secrets.
   - This can make a provider/backend look configured or ready when CE cannot actually authenticate.
   - Recommendation: do not import secret-dependent booleans by default, especially:
     - `ClaudeCodeCompatibleBackendConfigured.*`
     - `ClaudeCodeGLMZAIConfigured`
     - CLI/provider connected flags if they are treated as readiness rather than just local CLI presence
   - Non-secret config blobs can be imported, but availability should still require explicit CE auth/secret presence.

4. **Agent safety/permission settings are silently imported**
   - Current default preference import includes settings such as:
     - `codexAgentTools.bash.enabled`
     - `codexAgentTools.bash.approvalPolicy`
     - `codexAgentTools.bash.sandboxMode`
     - `codexAgentTools.mcpServerToggles`
     - `claudeCodeAllowNativeBashTool`
     - `claudeCodePermissionMode`
     - `claudeCodeMCPStrictModeEnabled`
     - `claudeCodeToolSearchEnabled`
     - `openCodeACPSessionMode`
     - `cursorACPToolPermissionLevel`
   - These change the agent permission/security posture of a CE install and should not be pulled in behind a generic “Import Classic RepoPrompt” button.
   - Recommendation: exclude them from default import, or add a clearly labeled opt-in such as “Import Classic agent permission and sandbox settings.”

5. **SwiftUI source detection probes live Classic data during render**
   - `ManageWorkspacesView` and `WorkspaceLandingView` call `ClassicRepoPromptImportService().defaultClassicSourceExists()` from view/body-derived paths.
   - That method performs filesystem checks, workflow directory enumeration, and Classic defaults access.
   - Recommendation: cache this in `@State` via `.task` / async helper, and refresh only after import or explicit lifecycle events.

6. **Workspace index decoding is all-or-nothing**
   - `loadIndex` decodes `[ClassicWorkspaceIndexEntry]` homogeneously.
   - One malformed entry can fail the entire workspace import section.
   - Recommendation: parse the top-level array, decode entries individually, import valid entries, and record malformed entries as per-entry failures. Fully invalid JSON can still be a section failure.

7. **`customStoragePath` is trusted too broadly**
   - The importer tries `entry.customStoragePath/workspace.json` before source-root fallbacks, then copies that workspace directory into CE.
   - Recommendation: only accept custom paths that resolve under the selected Classic workspace root or validated Classic global custom storage root. Reject symlink escapes.

8. **Sidecar copy should avoid symlink surprises**
   - Recursive sidecar copy can preserve/copy unexpected symlink sidecars.
   - Recommendation: copy known sidecars or enumerate with `.isSymbolicLinkKey` and skip symlinks. Avoid copying arbitrary external-linked content into CE storage.

9. **Source preset future-schema gating is missing**
   - Target preset saves have future-schema protection, but source Classic workflow/model preset JSON is decoded directly into current structs and then merged.
   - Recommendation: validate source `schemaVersion` before merge/write; unsupported future source schemas should fail/skip cleanly without mutating target preset files.

10. **`targetRootIsGlobalCustomStorage` is passed but discarded**
   - The parameter is threaded through public APIs but ignored in `importWorkspacesSynchronously`.
   - Recommendation: either remove it or implement/document the intended semantics. If sanitized workspace names still require an absolute pin for reload correctness, cover that explicitly with tests.

### Test gaps to close

Recommended focused tests before merging:

- Compile/build test for the `APISettingsViewModel` fix.
- Default import skips credential/configured readiness flags when secure stores are unavailable.
- Default import skips agent permission/sandbox settings.
- Lenient workspace index import: one valid entry + one malformed entry imports the valid workspace and reports a failure for the malformed one.
- Unsafe `customStoragePath` outside Classic root is rejected and does not copy external files.
- Symlink sidecars are not copied/followed.
- Future-schema Classic source preset documents do not mutate existing CE preset files.
- Existing CE data preservation tests for settings, saved prompts, context prompts, model presets, workflows, and secure-account failure branches.

### Real Classic data testing guidance

There is real old RepoPrompt Application Support data available on the maintainer machine, but it must not be mutated. The safe procedure is:

1. Copy Classic app support and shared prompt support into a temporary sandbox.
2. Copy Classic preferences into the sandbox as data, not by writing to the live defaults domain.
3. If copied preferences contain `GlobalCustomStorageURL`, do **not** follow the live path. Either copy that storage root into the sandbox and rewrite the copied preference dictionary to point at the sandbox copy, or ignore that preference for the validation run.
4. Call `importClassicRepoPromptData(...)` with explicit sandbox source/target paths.
5. Pass `sourceSecureValueReader: nil` and `targetSecureStore: nil` unless testing a mocked secure flow.
6. Assert all writes land under the sandbox target root only.

Do **not** run the default UI import path against live Classic data for testing.

### Suggested implementation direction

For a mergeable first version, I’d keep the scope conservative:

- Fix the compile blocker.
- Delete or defer the unused keychain reader.
- Import workspaces, settings, presets, prompts, workflows, model/custom-provider non-secret preferences.
- Skip credential readiness flags and agent permission/sandbox settings by default.
- Cache Classic source availability in the UI.
- Add per-entry index resilience and path/symlink guards.
- Add focused tests for the above.

That should preserve the value of the migration while avoiding false provider readiness, silent security posture changes, and any risk to real Classic data.

baron (changes requested):
Three independent blockers remain on `289337811aa1de6c87144c387becd7266295a720`: the head does not compile (`APISettingsViewModel.swift:557` calls undefined `postCompatibleBackendAvailabilityChanged()`); the default import silently transfers agent approval/sandbox/permission settings; and externally controlled workspace/sidecar paths are recursively copied without resolved-root containment or symlink-escape rejection. Fix the compile error, place execution authority behind explicit opt-in, enforce approved-root containment, add regressions, resolve conflicts, and rerun CI.

morluto:
### Audit disposition — **blocker; redesign before merge** (2026-08-14)

This migration spans workspaces, chats/agent sessions, settings, prompts, workflows, provider/CLI configuration, and copied directories, but it does not establish one transactional migration boundary. A failure or concurrent launch can therefore leave a partially imported state whose ownership and retry semantics are unclear.

Please rework this around a versioned staged import manifest, strict source/destination path containment, explicit treatment of secret-bearing configuration, a single commit/recovery protocol, and a process-level exclusion mechanism. The branch is also stale relative to the current workspace/domain authority model. The product need is valid, but I would not merge this implementation as-is.

morluto:

**Disposition: block pending a transactional migration design.** This imports many independent authorities—workspace documents, sessions, settings, presets, prompts, workflows, provider configuration, and filesystem trees—but the current evidence does not establish crash recovery or all-or-nothing/idempotent behavior across those stores. A failure or concurrent launch can leave a mixed partially imported state.

Before merge, add an exclusive migration lock, a versioned manifest/checkpoint, idempotent per-step identities, resume/rollback semantics, strict source/destination path containment, and a dry-run/result report. The no-Keychain default is sensible, but the skipped-secret policy should be explicit in the migration result. The focused tests and one successful E2E do not substitute for interruption and concurrent-launch tests.

PR #1, opened 2026-05-31
Refresh README introduction

najibninaba:
## Summary

Updates the README opening so first-time visitors understand RepoPrompt CE before they reach setup instructions.

Changes include:
- Introduce RepoPrompt CE as the open-source macOS Context IDE for AI coding agents.
- Replace the build-tooling-first intro with an outcome-focused product description.
- Add a concise "What You Can Do" section for context curation, CodeMaps, Git diffs, Context Builder, built-in chat, Agent Mode, and MCP clients.
- Add a "Project Status" section covering the paid-product-to-open-source transition, Apache-2.0 licensing, and release-readiness tracking.
- Add a CONTRIBUTING note that raw generated RP outputs are working artifacts unless distilled into durable docs.

## Review

The README intro copy was reviewed with RP Agent Workflow Review using the `design` agent label.

Result: PASS.

## Validation

- `make guardrails`
- Checked `README.md` and `CONTRIBUTING.md` diff.
- Checked for long/broken lines over 100 chars in both files.

PR #146, opened 2026-06-10
Inherit parent worktrees in Context Builder

provencher:
## Summary
- freeze an authoritative worktree-aware Context Builder invocation context for Agent Mode runs
- thread the inherited projection through provider cwd, nested MCP tools, selection/accounting, exports, and plan/review generation
- initialize and self-heal worktree codemaps while preserving logical display paths and fail-closed behavior
- harden `ask_oracle` packaging against canonical-checkout fallback and add multi-root worktree coverage

## Validation
- contribution commit/push preflights and secret scans passed
- `make dev-lint`
- `make dev-test`
- focused Context Builder, code-structure, Oracle, and worktree projection suites passed
- `git diff --check`

## Live verification
- `make dev-smoke` was attempted non-disruptively, but the existing CE window was not MCP-enabled for the workspace-switch stage
- no visible app relaunch or live smoke launch was performed

### DIFF
--- a/Sources/RepoPrompt/Features/AgentMode/Services/AgentWorkspaceLookupContextResolver.swift
+++ b/Sources/RepoPrompt/Features/AgentMode/Services/AgentWorkspaceLookupContextResolver.swift
@@ -35,6 +35,34 @@ struct AgentWorkspaceLookupContextIdentity: Hashable {
 }
 
 enum AgentWorkspaceLookupContextResolver {
+    static func requiredLookupContext(
+        source: AgentWorkspaceLookupContextSource,
+        store: WorkspaceFileContextStore
+    ) async throws -> WorkspaceLookupContext {
+        guard let sessionID = source.activeAgentSessionID,
+              !source.worktreeBindings.isEmpty
+        else {
+            return .visibleWorkspace
+        }
+
+        try AgentWorktreeRuntimeWorkspaceResolver.validateBindingsAvailable(source.worktreeBindings)
+        guard let projection = await WorkspaceRootBindingProjectionMaterializer(store: store).materialize(
+            sessionID: sessionID,
+            bindings: source.worktreeBindings
+        ),
+            !projection.isEmpty
+        else {
+            throw AgentWorkspaceLookupContextResolutionError.unavailableProjection
+        }
+
+        switch await store.rootScopeAvailability(projection.lookupRootScope) {
+        case .available:
+            return WorkspaceLookupContext(rootScope: projection.lookupRootScope, bindingProjection: projection)
+        case .sessionWorktreeUnavailable:
+            throw AgentWorkspaceLookupContextResolutionError.unavailableProjection
+        }
+    }
+
     static func lookupContext(
         source: AgentWorkspaceLookupContextSource,
         store: WorkspaceFileContextStore
@@ -50,5 +78,13 @@ enum AgentWorkspaceLookupContextResolver {
             return WorkspaceLookupContext.visibleWorkspace
         }
         return WorkspaceLookupContext(rootScope: projection.lookupRootScope, bindingProjection: projection)
+    }
+}
+
+enum AgentWorkspaceLookupContextResolutionError: LocalizedError {
+    case unavailableProjection
+
+    var errorDescription: String? {
+        "The Agent session worktree projection is unavailable. The operation stopped rather than falling back to the canonical checkout."
     }
 }
--- a/Sources/RepoPrompt/Features/AgentMode/Services/AgentWorkspaceLookupContextResolver.swift
+++ b/Sources/RepoPrompt/Features/AgentMode/Services/AgentWorkspaceLookupContextResolver.swift
@@ -1,31 +1,78 @@
 import Foundation
 
+enum AgentSessionWorktreeBindingState: Equatable {
+    case notApplicable
+    case hydrated([AgentSessionWorktreeBinding])
+    case unhydrated
+    case unavailable
+
+    var bindings: [AgentSessionWorktreeBinding]? {
+        guard case let .hydrated(bindings) = self else { return nil }
+        return bindings
+    }
+}
+
 struct AgentWorkspaceLookupContextSource: Equatable {
     let activeAgentSessionID: UUID?
-    let worktreeBindings: [AgentSessionWorktreeBinding]
+    let worktreeBindingState: AgentSessionWorktreeBindingState
+
+    init(
+        activeAgentSessionID: UUID?,
+        worktreeBindingState: AgentSessionWorktreeBindingState
+    ) {
+        self.activeAgentSessionID = activeAgentSessionID
+        self.worktreeBindingState = worktreeBindingState
+    }
+
+    init(
+        activeAgentSessionID: UUID?,
+        worktreeBindings: [AgentSessionWorktreeBinding]
+    ) {
+        self.init(
+            activeAgentSessionID: activeAgentSessionID,
+            worktreeBindingState: activeAgentSessionID == nil ? .notApplicable : .hydrated(worktreeBindings)
+        )
+    }
+
+    var worktreeBindings: [AgentSessionWorktreeBinding] {
+        worktreeBindingState.bindings ?? []
+    }
 
     var identity: AgentWorkspaceLookupContextIdentity {
         AgentWorkspaceLookupContextIdentity(
             activeAgentSessionID: activeAgentSessionID,
-            worktreeBindingFingerprint: Self.worktreeBindingFingerprint(worktreeBindings)
+            worktreeBindingFingerprint: Self.worktreeBindingFingerprint(worktreeBindingState)
         )
     }
 
     static func worktreeBindingFingerprint(_ bindings: [AgentSessionWorktreeBinding]) -> String {
-        bindings
-            .map { binding in
-                [
-                    binding.repositoryID,
-                    binding.repoKey,
-                    StandardizedPath.absolute((binding.logicalRootPath as NSString).expandingTildeInPath),
-                    binding.worktreeID,
-                    StandardizedPath.absolute((binding.worktreeRootPath as NSString).expandingTildeInPath),
-                    binding.branch ?? "",
-                    binding.head ?? ""
-                ].joined(separator: "\u{1F}")
-            }
-            .sorted()
-            .joined(separator: "\u{1E}")
+        worktreeBindingFingerprint(.hydrated(bindings))
+    }
+
+    static func worktreeBindingFingerprint(_ state: AgentSessionWorktreeBindingState) -> String {
+        switch state {
+        case .notApplicable:
+            "not-applicable"
+        case .unhydrated:
+            "unhydrated"
+        case .unavailable:
+            "unavailable"
+        case let .hydrated(bindings):
+            bindings
+                .map { binding in
+                    [
+                        binding.repositoryID,
+                        binding.repoKey,
+                        StandardizedPath.absolute((binding.logicalRootPath as NSString).expandingTildeInPath),
+                        binding.worktreeID,
+                        StandardizedPath.absolute((binding.worktreeRootPath as NSString).expandingTildeInPath),
+                        binding.branch ?? "",
+                        binding.head ?? ""
+                    ].joined(separator: "\u{1F}")
+                }
+                .sorted()
+                .joined(separator: "\u{1E}")
+        }
     }
 }
 
@@ -39,16 +86,34 @@ enum AgentWorkspaceLookupContextResolver {
         source: AgentWorkspaceLookupContextSource,
         store: WorkspaceFileContextStore
     ) async throws -> WorkspaceLookupContext {
-        guard let sessionID = source.activeAgentSessionID,
-              !source.worktreeBindings.isEmpty
-        else {
+        guard let sessionID = source.activeAgentSessionID else {
+            return .visibleWorkspace
+        }
+        guard case let .hydrated(bindings) = source.worktreeBindingState else {
+            throw AgentWorkspaceLookupContextResolutionError.unknownBindingState
+        }
+        guard !bindings.isEmpty else {
             return .visibleWorkspace
         }
 
-        try AgentWorktreeRuntimeWorkspaceResolver.validateBindingsAvailable(source.worktreeBindings)
+        let visibleRootPaths = await Set(store.rootRefs(scope: .visibleWorkspace).map(\.standardizedFullPath))
+        let logicalRootPaths = Set(bindings.compactMap {
+            AgentWorktreeRuntimeWorkspaceResolver.standardizedWorkspacePath($0.logicalRootPath)
+        })
+        guard logicalRootPaths.count == bindings.count,
+              logicalRootPaths.isSubset(of: visibleRootPaths)
+        else {
+            throw AgentWorkspaceLookupContextResolutionError.unavailableProjection
+        }
+
+        do {
+            try AgentWorktreeRuntimeWorkspaceResolver.validateBindingsAvailable(bindings)
+        } catch {
+            throw AgentWorkspaceLookupContextResolutionError.unavailableProjection
+        }
         guard let projection = await WorkspaceRootBindingProjectionMaterializer(store: store).materialize(
             sessionID: sessionID,
-            bindings: source.worktreeBindings
+            bindings: bindings
         ),
             !projection.isEmpty
         else {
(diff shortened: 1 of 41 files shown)


baron:
# Code review — PR #146 (Inherit parent worktrees in Context Builder)

> 🤖 This review was prepared **exclusively by Anthropic's Fable 5 via Claude Code**. Process: 7 independent finder passes over the full diff (line-by-line scan, removed-behavior audit, cross-file caller tracing, reuse/simplification/efficiency/altitude), 36 raw candidates deduped to 13, each verified by an independent adversarial verifier against the actual code on both sides of the diff. 5 correctness bugs confirmed, 3 plausible, 5 candidates refuted (listed at the end so they don't get re-chased).

## Confirmed correctness bugs

### 1. Agent-Mode runs silently drop selected Context Builder prompts (wrong ID space) — `ContextBuilderAgentViewModel.swift:2831`

The new frozen-context branch in `captureRunStartState` sets:

```swift
session.runStartContextBuilderPromptIDs = Set(workspaceContext.frozenTabContext.selectedMetaPromptIDs)
```

`TabContextSnapshot.selectedMetaPromptIDs` are **StoredPrompt** (compose-tab prompt library) IDs, while `runStartContextBuilderPromptIDs` is consumed by `ContextBuilderPromptStorage.promptText(for:)` (`ContextBuilderPromptsOverlay.swift:154`), which filters by membership in its own store (`ContextBuilderPrompts.json` + the built-in "Interview" prompt). Foreign UUIDs match nothing → `promptText` returns nil → no meta-prompt block is emitted.

On `main`, the same call site captured `selectedContextBuilderPromptIDs` (synced from `tabState.contextBuilder.selectedContextBuilderPromptIDs`), so agent-invoked runs *did* inject the selected prompt. This is a deterministic, silent regression on exactly the code path the PR exists for (`purpose == .agentModeRun`); UI and non-agent MCP runs are unaffected. The frozen branch also always sets `runStartPromptText`/`runStartSelection`, so the foreign ID set is always the one used.

**Fix:** carry the Context Builder prompt IDs (`tabState.contextBuilder.selectedContextBuilderPromptIDs`) in the frozen snapshot, or keep the old capture for this field.

### 2. The "frozen" workspace scope is not actually frozen — `ContextBuilderWorkspaceContext.swift:49` (and the worktree branch at ~:83)

The no-bindings branch encodes the frozen workspace as an **exclusion set**:

```swift
.sessionBoundWorkspace(logicalRootPaths: loadedPrimaryRootPaths.subtracting(requestedRootPaths), physicalRootPaths: [])
```

`rootsForPathLookup` (`WorkspaceFileContextStore.swift:4392`) evaluates this against the store's **live** roots at lookup time: include every `primaryWorkspace` root *not* in the exclusion set. In the common case (only the workspace's own roots loaded) the exclusion set is **empty**, so the "frozen" scope degenerates to "whatever primary roots are loaded right now."

Reachable mid-run mutations that leak through the frozen context:

- `addFolder`/`removeFolder` on the active workspace go through **no** session cancellation; an added root immediately appears in the run's file tree / selection accounting / follow-up packaging.
- `removeFolder` of the last root and `handleAllFoldersUnloaded` call `switchWorkspace` directly (`WorkspaceManagerViewModel.swift:6270`, `:6203`), bypassing `requestWorkspaceSwitch`'s run-cancellation — the in-flight run then resolves against the fallback workspace's roots.
- Even the gated switch path has a race: `cancelAllActiveRuns` signals cancellation without awaiting run-task termination.

The worktree branch has the same property for later-loaded foreign primary roots (its `logicalRootPaths` excludes only the bound roots). And availability checking is blind to all of this: `rootScopeAvailability` only diffs `physicalRootPaths` against loaded session-worktree roots, so with `physicalRootPaths: []` it is unconditionally `.available`, and `validateAvailability()` only stats binding dirs + `providerWorkspacePath`.

**Fix:** encode the frozen scope as a positive include-list (e.g. capture root IDs/paths at freeze time and resolve against exactly those), rather than exploiting the inverted `sessionBoundWorkspace` semantics.

### 3. Self-healing codemap repair stalls hot MCP calls ~5s during initial indexing, and retries cancel in-flight scans — `MCPServerViewModel.swift:3543`, `CodeScanActor.swift:749/883`

`buildCodeStructureDTO` unconditionally calls `repairMissingCodemapSnapshots` (default 5s timeout, no caller overrides) for all `get_code_structure` and `workspace_context(include=code)` paths. During initial indexing of a large workspace:

- `.selfHealing` requests get **no priority**: `CodeScanActor.requestScans` does `removeQueuedRequest` + `queue.append(...)` into a strict-FIFO queue (`queue.removeFirst()` at `:1058`), so the repaired files land *behind* thousands of cold initial-load requests, the 25 ms poll loop (`WorkspaceFileContextStore.swift:2938`) runs out the full 5 s, and the tool replies 5 s late with "repair pending".
- Worse, `removeQueuedRequest` (`CodeScanActor.swift:749-758`) **cancels an active scan task** for that fileID — so an agent retrying on the "retry with narrower paths" hint can repeatedly kill the file's scan just as it finally reaches the front, pushing completion further away each time.
- `.selfHealing` bypasses the mod-date skip (`:883-889`), so every retry re-pays the full requeue cost. There is no already-queued check, no backoff, and no skip while the root's initial scan is in flight.

**Fix:** skip submission for fileIDs already queued/active (never cancel an active scan from the repair path), and return "pending" immediately when the root's initial scan is in flight instead of burning the timeout.

### 4. Never-submitted files reported as "repair pending" forever — `WorkspaceFileContextStore.swift:2954`

`pendingFileIDs` is computed from **all** missing-snapshot files, but `codemapScanRequests` (`:3110-3135`) silently drops files whose content load throws or returns nil:

```swift
guard let content = loaded.content else { continue }   // binary-detected or >10MB
...
} catch { continue }                                    // deleted/unreadable
```

A supported-extension file over the 10 MB content limit is a **deterministic, persistent** instance (not a race). Such files land in `pendingFileIDs` on every call, render as *"Codemap repair pending: N (bounded wait expired; retry with narrower `paths` if needed)"* — no wait actually ran for them and no retry can ever succeed — and they never reach `unmappedPaths`, where `main` put them. (The zero-submissions case does return early without burning the timeout — that part is correct.)

**Fix:** compute pending as `submittedFileIDs` without snapshots; classify never-submitted files as unmapped.

### 5. Same-destination transitions bypass the new fail-closed binding validation — `AgentModeViewModel.swift:5135`

`transitionWorktreeBindings` early-returns through `replaceWorktreeBindings` (which does no path/availability validation — it just assigns `session.worktreeBindings`) whenever the **primary** execution destination is unchanged. The new `materializeWorktreeBindingsForTransition` check at `:5168` is only reached on destination-changing transitions, and `executionDestinationPath` (`:5190`) considers only the primary binding.

Reachable: `MCPWorktreeToolProvider.applyBinding` builds `existing.filter{...} + newBinding` and can target a **secondary** root (`repo_key`/`repo_root`), and `GitRepoTargetResolver.resolveWorktree` matches prunable worktree entries whose directories no longer exist. Result: the transition succeeds unvalidated, the session holds an unavailable binding, and every subsequent oracle/context_builder/search/code-structure call fails closed mid-run — where a destination-changing transition would have rejected the same bindings upfront with `ExecutionLocationTransitionError.unavailable`.

**Fix:** run `materializeWorktreeBindingsForTransition` before the same-destination early return (or inside `replaceWorktreeBindings` when bindings actually change).

## Plausible issues (verified as realistic, not 100% confirmed)

### 6. ask_oracle can still silently fall back to the canonical checkout via transiently-empty bindings — `MCPOracleToolService.swift:504`

The fail-closed path requires `agentSessionID != nil && !worktreeBindings.isEmpty`. But a genuinely worktree-bound session can transiently present empty bindings: `session(for:tabID, createIfNeeded: false)` lazily creates an **unhydrated** session (`AgentModeViewModel.swift:2859`) and `seedUnhydratedSession` seeds agent/model/effort but **not** `worktreeBindings`, even though `sessionIndex` carries `worktreeBindingSummaries`; child sessions also have a window before `copyParentWorktreeBindingsIfNeeded` runs. An `ask_oracle` arriving in either window packages canonical-checkout content with no error — the exact silent fallback this PR's `requiredLookupContext` was added to stop. Suggest seeding bindings in `seedUnhydratedSession` (the summaries are available) or failing closed when the session is unhydrated.

### 7. evaluateVirtualPromptEntries scope updated in one caller, missed in two — `MCPServerViewModel+WorkspaceContext.swift:54` and `:300`

`buildTabSelectionReply` now evaluates prompt entries under `lookupContext.rootScope`, but `buildTabWorkspaceContext` and `buildExportSelectedFileInfos` still call `evaluateVirtualPromptEntries` with the default `.allLoaded` while their sibling `collect()` calls use the scoped value. Token accounting can diverge from `manage_selection` for the identical selection in resolution edge cases (ambiguous relative paths, git-data-root entries). Trivial to align by passing `lookupContext.rootScope` at both sites.

### 8. Routing errors now give guidance the new gate rejects — `MCPServerViewModel.swift:2727`

The hard rejection of explicit `context_id` for `purpose == .agentModeRun` (`:2678`) looks intentional and is correctly messaged at the gate. But when *implicit* run-scoped resolution fails, the raw routing errors now propagate, and those messages still advise "pass a matching explicit tab context hint" / "bind_context/context_id" (`MCPServerViewModel+TabContext.swift:1680`, `:1694`) — advice the gate then bounces, creating a contradictory error/retry loop. The old tailored "Retry after routing settles" message was deleted, and the new rejection has no test.

## Cleanup (confirmed, non-blocking)

- **Triplicated fail-closed policy.** The materialize → `isEmpty` → availability sequence exists in three PR-added places with bespoke error types: `AgentWorkspaceLookupContextResolver.requiredLookupContext` (`:38-62`), `AgentModeViewModel.materializeWorktreeBindingsForTransition` (`:5195-5222`, which also *omits* `validateBindingsAvailable` and contains an unreachable `.available: []` switch arm at `:5212`), and `ContextBuilderWorkspaceContext.resolve` (`:35-103`, which substitutes hand-rolled subset checks for `rootScopeAvailability`). The copies already disagree structurally; consolidating on `requiredLookupContext` with mapped errors would prevent divergence.
- **Hot-path inefficiencies.** `rootScopeAvailability` (`WorkspaceFileContextStore.swift:1051-1061`) stats **every** loaded session-worktree root across all sessions on each session-bound check (search, code structure, oracle resolution) — the requested set is only used afterward for subtraction; stat only the requested roots. And `buildCodeStructureDTO` (`MCPServerViewModel.swift:3535/3547`) materializes the full O(all-codemaps) snapshot dictionary twice per call in the common nothing-to-repair case (the repair early-return rebuilds it at `:2920-2925` and the first copy is discarded); guard the repair call on `!repairFiles.isEmpty`.
- The 25 ms polling loop in `repairMissingCodemapSnapshots` duplicates the store's existing `codemapUpdates()` event stream; an event-driven wait would remove up to 200 actor wake-ups per repair and the 25 ms tail latency.

## Checked and refuted (no action needed)

For completeness, these candidates were investigated and ruled out:

- **Display-path/scope narrowing in `buildCodeStructureDTO`** — upstream resolution uses the same `lookupContext.rootScope` at every production call site, so out-of-scope records can't arrive; the silent cross-scope `continue` is the test-asserted worktree-isolation guarantee (`testSwitchingCodeStructureScopeFromWorktreeAToBDoesNotReuseA`).
- **Context export packaging canonical content on broken projections** — impossible: the materializer fabricates fail-closed roots (`WorkspaceRootBindingProjection.swift:303-317`), so lookups miss rather than read base-checkout content. (Exports do fail *silently-empty* rather than loudly like oracle/context_builder — a softer UX inconsistency worth a thought, but not the claimed bug.)
- **Git-diff filesystem-fallback regression for no-bindings agent runs** — `allowsSelectedGitDiffFilesystemFallback` is dead on the headless path: `preAssemblePromptContext` hardcodes `.expandFolders`, whose branch never read the flag on `main` either.
- **`/private` path-standardization mismatch** (`URL.standardizedFileURL` vs `NSString.standardizingPath`) — empirically tested on macOS: both APIs standardize identically for all realistic absolute paths, including the existence-conditional `/private` strip.
- **Frozen-tab-context clobber between concurrent same-kind runs** — pending contexts are runID-keyed (`PendingRunScopedContextStore.enqueueReplacing`), consumption is runID-matched and fail-closed, and the bootstrap window is serialized by `HeadlessAgentConnectionGate`.

---

**Recommendation:** fix #1 (deterministic silent regression on the PR's main path) and #5 before merge; #2 deserves a design decision (positive include-list scope) since it undercuts the PR's central "freeze an authoritative invocation context" guarantee; #3/#4 materially degrade agent UX on large workspaces and have small, contained fixes.

baron:
Code review findings from a fresh worktree review of PR #146:

### Findings

1. **P1: `export_response` handoffs are not worktree-safe.**
`makeOracleExportDestination` still roots generated exports at `workspace.repoPaths.first` (`Sources/RepoPrompt/Infrastructure/MCP/ViewModels/MCPServerViewModel.swift:4055`) and `GeneratedOracleExportFileWriter` writes/validates through `.visibleWorkspace` (`Sources/RepoPrompt/Infrastructure/MCP/OracleExportFileWriter.swift:11`). In a worktree-bound child agent, `read_file` translates a canonical absolute path under the logical root to the physical worktree root (`Sources/RepoPrompt/Infrastructure/WorkspaceContext/WorkspaceRootBindingProjection.swift:70`, `Sources/RepoPrompt/Infrastructure/MCP/WindowTools/MCPFileToolProvider.swift:330`). So `context_builder`, `ask_oracle`, and `oracle_send` can write `/repo/prompt-exports/...` but return an instruction a child resolves as `/repo-wt/prompt-exports/...`, which fails or reads a coincidental stale file. Affected capture sites: `Sources/RepoPrompt/Infrastructure/MCP/WindowTools/MCPContextBuilderToolProvider.swift:205`, `Sources/RepoPrompt/Infrastructure/MCP/MCPOracleToolService.swift:136`, and `Sources/RepoPrompt/Infrastructure/MCP/MCPOracleToolService.swift:274`.

2. **P1: session-worktree codemap initialization is racy and currently fails tests.**
`initializeCodemapsForSessionWorktreeRoots` marks roots initialized before confirming that scan requests were actually submitted (`Sources/RepoPrompt/Infrastructure/WorkspaceContext/WorkspaceFileContextStore.swift:2880`, union at `:2892`), and projection materialization calls it immediately after loading session roots (`Sources/RepoPrompt/Infrastructure/WorkspaceContext/WorkspaceRootBindingProjection.swift:318`). If root catalog ingress has not populated files yet, the initializer can scan zero files and suppress later retries. This reproduced in `WorkspaceRootBindingProjectionTests.testMaterializerInitializesSessionWorktreeCodemapsIdempotently` and `AgentRunWorktreeStartTests.testBindingTransitionMaterializesAndInitializesSessionWorktreeCodemap`.

3. **P2: `manage_selection` content blocks can expose physical worktree paths.**
`buildSelectionPreviewReply` uses a projected formatter but the direct `makeSelectionReply` call does not pass `pathProjection` (`Sources/RepoPrompt/Infrastructure/MCP/ViewModels/MCPServerViewModel+SelectionReply.swift:310`). The callee supports it and block generation uses it (`Sources/RepoPrompt/Infrastructure/MCP/ViewModels/MCPServerViewModel+SelectionCore.swift:594`, `:598`).

4. **P2: `get_code_structure(paths:)` can silently omit fresh supported files beyond the repair window.**
Explicit path calls pass `includeUnmappedPaths: false` (`Sources/RepoPrompt/Infrastructure/MCP/WindowTools/MCPFileToolProvider.swift:170`). `buildCodeStructureDTO` repairs only `prefix(repairLimit)` (`Sources/RepoPrompt/Infrastructure/MCP/ViewModels/MCPServerViewModel.swift:3537`, `:3542`), then files without snapshots are only reported when `includeUnmappedPaths` is true (`:3566`); omission counts are derived from renderable codemaps only (`:3594`).

5. **P3: virtual `workspace_context` token stats can still account against canonical/all-loaded paths.**
The virtual token breakdown is called without a worktree lookup context (`Sources/RepoPrompt/Infrastructure/MCP/ViewModels/MCPServerViewModel+WorkspaceContext.swift:201`), and the helper builds file tree/git diff token components from `context.selection` with `.allLoaded` (`Sources/RepoPrompt/Infrastructure/MCP/ViewModels/MCPServerViewModel+SelectionReply.swift:89`, `:103`, `:124`, `:150`, `:153`).

### Validation

Passed:
- `git diff --check $(git merge-base HEAD origin/main)...HEAD`
- `make dev-test FILTER=ContextBuilderWorkspaceContextTests` (3 tests)
- `make dev-test FILTER=ContextBuilderWorktreeInheritanceTests` (3 tests)
- `make dev-test FILTER=MCPAskOracleWorktreeTests` (4 tests)
- `make dev-test FILTER=MCPCodeStructureWorktreeTests` (6 tests)
- `make dev-test FILTER=ToolOutputFormatterWorktreeTests` (13 tests)

Failed/reproduced:
- `make dev-test FILTER=WorkspaceRootBindingProjectionTests`: failed in `testMaterializerInitializesSessionWorktreeCodemapsIdempotently` waiting for the worktree codemap snapshot.
- `make dev-test FILTER=WorkspaceRootBindingProjectionTests.testMaterializerInitializesSessionWorktreeCodemapsIdempotently`: failed the same way.
- `make dev-test FILTER=AgentRunWorktreeStartTests`: failed in `testBindingTransitionMaterializesAndInitializesSessionWorktreeCodemap` (`repair.pendingFileIDs` nonempty / missing snapshot).

I did not run live UI/app smoke because that would require explicit approval to launch/relaunch/stop the visible CE app.

PR #148, opened 2026-06-10
Prefer OpenCode provider path before native defaults

fitchmultz:
## Summary
- searches OpenCode provider-specific paths before generic native defaults when resolving ACP launches
- keeps the shared OpenCode launch profile in the same provider-first order
- adds a regression assertion that `~/.opencode/bin` is first for OpenCode supplemental paths

## Why
When a machine has a real `opencode` in a native/default location, resolver tests and user installs can resolve that binary before the provider-specific `~/.opencode/bin` path. OpenCode should prefer its provider-owned install location before generic fallbacks.

## Validation
- `make dev-test FILTER=OpenCodeACPLaunchResolverTests`
- `make dev-lint`

## Review
- Subagent reviewer pass: no material findings after moving the ordering into `CLILaunchProfile` and adding the regression assertion.

### DIFF
--- a/Sources/RepoPrompt/Infrastructure/AI/Providers/OpenCode/OpenCodeACPLaunchResolver.swift
+++ b/Sources/RepoPrompt/Infrastructure/AI/Providers/OpenCode/OpenCodeACPLaunchResolver.swift
@@ -123,7 +123,7 @@ final class OpenCodeACPLaunchResolver: @unchecked Sendable {
             return try resolveExplicitLaunch(for: config, environment: environment)
         }
 
-        let effectiveHints = CLIPathHints.nativeDefaultsSupplemented(with: config.additionalPathHints)
+        let effectiveHints = Self.effectiveSearchPaths(providerSpecificPaths: config.additionalPathHints)
         let resolved = CommandPathResolver.resolve(
             configuredCommand,
             environment: environment,
@@ -145,7 +145,7 @@ final class OpenCodeACPLaunchResolver: @unchecked Sendable {
         guard configuredCommand.contains("/") else {
             throw OpenCodeACPLaunchResolutionError.environmentDiscoveryRequired(configuredCommand)
         }
-        let effectiveHints = CLIPathHints.nativeDefaultsSupplemented(with: config.additionalPathHints)
+        let effectiveHints = Self.effectiveSearchPaths(providerSpecificPaths: config.additionalPathHints)
         return try validatedLaunch(
             entryPath: CommandPathResolver.expandPath(configuredCommand, environment: environment),
             configuredCommand: configuredCommand,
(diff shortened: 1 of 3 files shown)


moonray:
This is still an issue on 1.0.13 for me as well, despite opencode being in my PATH in .zshrc

baron:
Non-blocking review note from a local validation pass:

The implementation looks correct and well-scoped. One nuance worth documenting: this changes the supplemental fallback hint order for OpenCode, but it does not override the user's shell/PATH resolution. `CommandPathResolver.resolve` still prefers explicit paths and interactive-shell lookup before falling back to the merged search paths. That seems like the right behavior; the PR description could just clarify that `~/.opencode/bin` is preferred within the fallback hint tier, not ahead of a shell-resolved `opencode`.

Suggested test strengthening: the current `.first == "~/.opencode/bin"` assertion locks in static profile order, but a runtime regression test gives stronger coverage. In my local worktree I added a decoy `opencode` under fake `~/.local/bin` alongside the provider-owned fake `~/.opencode/bin`, then asserted the provider-owned executable wins. That would fail under the old native-first ordering and passes with this PR.

Tiny optional nits, not blockers:
- `effectiveSearchPaths(providerSpecificPaths:)` is a little narrow as a label because `config.additionalPathHints` may include arbitrary/user hints, but promoting them ahead of native defaults seems reasonable.
- The private resolver wrapper only forwards to `CLILaunchProfiles.providerSpecificPathsSupplementedWithNativeDefaults`; calling the helper directly, or exposing this through `CLIPathHints` for symmetry with Cursor/Claude, would be marginally cleaner.

Local validation with the decoy-binary test in place:
- `make dev-test FILTER=OpenCodeACPLaunchResolverTests` passed, 10/10.
- `make dev-lint` passed.

I did also notice the remote Build and Test check was cancelled during the full app-test step after the app/MCP builds succeeded, apparently in unrelated AgentMode tests rather than OpenCode-specific coverage.

baron:
Pushed two updates to this branch (maintainer edit):

- **cff43bc** — strengthens the regression test: plants a decoy `opencode` in a native default path (`~/.local/bin` under the fake home) so `testDefaultProfileResolvesOpenCodeFromProviderSpecificHomeBin` now fails if native defaults ever regain priority over `~/.opencode/bin` at resolution time, rather than only asserting the static hint order.
- **21f621b / fb9419f** — merged latest `main` (Fable 5 catalogs, #155 MCP hardening) so CI runs against the current merge result; the branch was behind and the previous Build and Test run had been cancelled.

## Validation

- `make dev-test FILTER=OpenCodeACPLaunchResolverTests` — 10/10 passing, including the new decoy test
- `make dev-lint` — clean (0 violations, 1107 files)
- `rpce-contribution-check` push preflight — passed (full coordinated suite + secret scans)

### Live validation (latest OpenCode)

- Upgraded OpenCode CLI 1.4.3 → **1.17.3** at `~/.opencode/bin` (sole install on the machine)
- Rebuilt the CE debug app from this branch and drove it with `rpce-cli-debug` against a fresh `sst/opencode` checkout as the workspace
- Two OpenCode Zen free-model sessions completed end to end: `opencode/big-pickle` (smoke reply) and `opencode/deepseek-v4-flash-free` (tool-use session reading the repo README)
- Confirmed the spawned process was `/Users/baron/.opencode/bin/opencode acp` — provider-path-first resolution working live

One scope note for reviewers: this ordering governs the supplemental-hint fallback tier. `CommandPathResolver` still prefers an explicit configured path and the user's interactive-shell `PATH` lookup first, so a user whose shell resolves a different `opencode` keeps that behavior — the provider path wins only within the hint tier, which is what the resolver consults when shell/PATH discovery doesn't settle it.

PR #149, opened 2026-06-10
Synchronize CLI process output capture

fitchmultz:
## Summary
- protects `CLIProcessRunner.run` stdout/stderr buffers while drain queues append concurrently
- snapshots captured output after drain completion before logging and returning
- adds a regression test for fast process stdout/stderr capture

## Why
Full-suite pi validation exposed an order/load-sensitive version-probe failure where a short-lived fake `pi --version` process exited successfully but stdout was sometimes observed as empty. The process runner was mutating captured stdout/stderr `Data` from drain queues without synchronization. This is a generic process-runner correctness fix and is kept separate from the pi provider PR.

## Stack note
This branch is currently stacked on #148 (`Prefer OpenCode provider path before native defaults`) because local full-suite validation needs that independent OpenCode test-isolation fix. After #148 merges, this PR can be rebased onto `main` and should show only the CLIProcessRunner changes.

## Validation
- `make dev-test FILTER=CLIProcessRunnerLifecycleTests\|PiIntegrationConfigurationTests`
- `make dev-lint`
- subagent reviewer: no material findings

### DIFF
--- a/Sources/RepoPrompt/Infrastructure/Process/CLIProcessRunner.swift
+++ b/Sources/RepoPrompt/Infrastructure/Process/CLIProcessRunner.swift
@@ -239,8 +239,7 @@ final class CLIProcessRunner {
             } operation: {
                 // 1) Immediately start draining child output (prevents cross‑pipe back‑pressure)
                 let group = DispatchGroup()
-                var stdoutData = Data()
-                var stderrData = Data()
+                let outputBuffer = LockedProcessOutputBuffer()
 
                 group.enter()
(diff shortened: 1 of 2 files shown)


fitchmultz:
Closing for upstream PR hygiene. The core change is still a valid small process-runner correctness fix, and focused validation passed after rebasing locally onto current main, but the required push preflight could not complete on the standalone branch because current upstream tests can still stall under conductor. The patch remains carried and validated on origin/feature/pi-integration. This can be reopened or recreated after the upstream test/conductor stall is fixed at the root.

PR #155, opened 2026-06-10
Harden MCP concurrency and workspace switching

provencher:
## Summary

- isolate `file_search` in a dedicated serialized per-connection lane so long searches no longer block unrelated same-connection tools
- harden MCP admission, eviction, replacement handoff, execution ownership, cancellation, diagnostics, and watchdog lifecycle behavior
- move broad-search admission ahead of freshness work and coalesce freshness successors to one pending maximum target
- bound workspace-root unload teardown, fence late ingress/watcher completions, and recover/report blocked or stale workspace switches
- apply an argument-aware 120-second watchdog only to switch-producing `manage_workspaces` operations

## TDD coverage

The change was developed red-first with deterministic regressions for:

- same-connection search head-of-line blocking and search serialization
- admission/eviction/replacement identity and lifecycle races
- cross-connection active-tool ownership and cancellation races
- freshness admission ordering and successor-flight coalescing
- wedged publisher ingress and late completion after root recreation
- workspace switch cancellation, recovery, stale blockage, busy settling, and presentation
- cooperative and uncooperative workspace-switch MCP watchdog behavior

## Validation

- full coordinated test suite: 1,373 passed, 1 skipped, 0 failures
- `make dev-lint`: passed, 0 violations
- `make dev-swift-build PRODUCT=repoprompt-mcp`: passed
- `make dev-smoke`: passed
- contribution commit and push preflights: passed
- staged/outgoing secret scans: clean

### DIFF
--- a/Sources/RepoPrompt/App/WindowState.swift
+++ b/Sources/RepoPrompt/App/WindowState.swift
@@ -856,7 +856,7 @@ class WindowState: ObservableObject {
 
         if workspaceManager.activeWorkspaceID != route.workspaceID {
             let switchResult = await workspaceManager.requestWorkspaceSwitch(to: targetWorkspace, saveState: true)
-            if !switchResult.didSwitch, workspaceManager.activeWorkspaceID != route.workspaceID {
+            if !switchResult.didSwitch {
                 return .workspaceSwitchBlocked(switchResult.message)
             }
         }
--- a/Sources/RepoPrompt/App/WindowStateComposition.swift
+++ b/Sources/RepoPrompt/App/WindowStateComposition.swift
@@ -29,10 +29,12 @@ enum WindowStateCompositionFactory {
         windowID: Int,
         deferredInitialAgentSystemWorkspaceRefresh: Bool,
         sharedMCPService: MCPService,
-        contextBuilderProviderFactory: ContextBuilderAgentViewModel.ProviderFactory? = nil
+        contextBuilderProviderFactory: ContextBuilderAgentViewModel.ProviderFactory? = nil,
+        workspaceFileContextStore injectedWorkspaceFileContextStore: WorkspaceFileContextStore? = nil,
+        workspaceSwitchTimingPolicy: WorkspaceSwitchTimingPolicy = .production
     ) -> WindowStateComposition {
         // 1) Workspace file context store + visible file-tree UI adapter
-        let workspaceFileContextStore = WorkspaceFileContextStore()
+        let workspaceFileContextStore = injectedWorkspaceFileContextStore ?? WorkspaceFileContextStore()
         let workspaceSearchService = WorkspaceSearchService()
         let workspaceFilesViewModel = WorkspaceFilesViewModel(workspaceFileContextStore: workspaceFileContextStore)
 
@@ -59,7 +61,8 @@ enum WindowStateCompositionFactory {
         let workspaceManager = WorkspaceManagerViewModel(
             fileManager: workspaceFilesViewModel,
             promptViewModel: promptManager,
-            workspaceSearchService: workspaceSearchService
+            workspaceSearchService: workspaceSearchService,
+            switchTimingPolicy: workspaceSwitchTimingPolicy
         )
         let selectionCoordinator = WorkspaceSelectionCoordinator(
             workspaceManager: workspaceManager,
--- a/Sources/RepoPrompt/Features/Diagnostics/MCP/MCPConnectionManager+DebugDiagnosticsReadSearchLatency.swift
+++ b/Sources/RepoPrompt/Features/Diagnostics/MCP/MCPConnectionManager+DebugDiagnosticsReadSearchLatency.swift
@@ -96,6 +96,33 @@ import MCP
                 return debugDiagnosticsError(op: op, code: "invalid_params", message: "`retry_after_ms` must be an integer between 0 and 60000.")
             }
 
+            let productionConfiguration = StoreBackedWorkspaceSearchLane.Configuration.production
+            let maxActiveLeases: Int
+            switch debugBoundedInt(
+                arguments,
+                "max_active_leases",
+                defaultValue: productionConfiguration.maxActiveLeases,
+                range: 1 ... 64
+            ) {
+            case let .value(parsed), let .defaulted(parsed):
+                maxActiveLeases = parsed
+            case .invalid:
+                return debugDiagnosticsError(op: op, code: "invalid_params", message: "`max_active_leases` must be an integer between 1 and 64.")
+            }
+
+            let maxQueuedWaiters: Int
+            switch debugBoundedInt(
+                arguments,
+                "max_queued_waiters",
+                defaultValue: productionConfiguration.maxQueuedWaiters,
+                range: 1 ... 64
+            ) {
+            case let .value(parsed), let .defaulted(parsed):
+                maxQueuedWaiters = parsed
+            case .invalid:
+                return debugDiagnosticsError(op: op, code: "invalid_params", message: "`max_queued_waiters` must be an integer between 1 and 64.")
+            }
+
             let targets = await debugSearchLaneTargets(windowID: requestedWindowID)
             guard !targets.isEmpty else {
                 let message = requestedWindowID.map { "No RepoPrompt window matched window_id \($0)." }
@@ -115,6 +142,8 @@ import MCP
             }
 
             let configuration = StoreBackedWorkspaceSearchLane.Configuration(
+                maxActiveLeases: maxActiveLeases,
+                maxQueuedWaiters: maxQueuedWaiters,
                 maxQueueWait: .milliseconds(maxQueueWaitMilliseconds),
                 retryAfterMilliseconds: retryAfterMilliseconds
             )
@@ -499,8 +528,8 @@ import MCP
                     [
                         "window_id": entry.windowID,
                         "configuration": [
-                            "active_capacity": 1,
-                            "max_queued": 1,
+                            "active_capacity": entry.snapshot.configuration.maxActiveLeases,
+                            "max_queued": entry.snapshot.configuration.maxQueuedWaiters,
                             "max_queue_wait_ms": entry.snapshot.configuration.maxQueueWaitMilliseconds,
                             "retry_after_ms": entry.snapshot.configuration.retryAfterMilliseconds
                         ],
--- a/Sources/RepoPrompt/Features/Diagnostics/MCP/MCPConnectionManager+DebugDiagnosticsReadSearchLatency.swift
+++ b/Sources/RepoPrompt/Features/Diagnostics/MCP/MCPConnectionManager+DebugDiagnosticsReadSearchLatency.swift
@@ -314,9 +314,30 @@ import MCP
             ]
         }
 
-        private func readSearchLimiterPayload(_ snapshot: AsyncLimiter.DebugSnapshot) -> [String: Any] {
+        private func readSearchLimiterPayload(
+            _ snapshot: MCPConnectionCallLimiterDebugSnapshot
+        ) -> [String: Any] {
             [
                 "found": true,
+                "lane_count": snapshot.laneCount,
+                "limit": snapshot.limit,
+                "permits": snapshot.permits,
+                "active_permit_count": snapshot.activePermitCount,
+                "waiter_count": snapshot.waiterCount,
+                "in_flight_count": snapshot.inFlight,
+                "oldest_waiter_age_ms": Self.debugOptionalValue(snapshot.oldestWaiterAgeMilliseconds),
+                "cancelled_waiter_count": snapshot.cancelledWaiterCount,
+                "is_closed": snapshot.isClosed,
+                "is_idle": snapshot.isIdle,
+                "lanes": [
+                    MCPConnectionCallLane.ordinary.rawValue: readSearchLimiterLanePayload(snapshot.ordinary),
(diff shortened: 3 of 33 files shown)


baron:
# Review: Harden MCP concurrency and workspace switching

## Overview

This PR (+12,494 / −963 across 29 files) does what it says: `file_search` gets its own serialized per-connection lane, the MCP admission/eviction/replacement machinery is identity-fenced against ABA races, workspace switching moves from boolean flags to a single-owner operation token with an explicit commit boundary and recovery, freshness work is coalesced to one active + one pending flight per root, and root unload teardown is time-bounded with layered fencing against late completions.

**The core concurrency engineering is genuinely good.** The limiter bundles keep check-and-mutate in single actor turns, lane routing happens after alias canonicalization so a search can't sneak onto the ordinary lane, the eviction protocol's tentative-close → revalidate → commit/restore handshake has no hangable exit path, and the freshness state machine has no constructible lost-target or dual-flight interleaving. The test suite is largely deterministic (injected clocks, DEBUG-gated continuation gates at exact suspension points) and the TDD claims look credible.

That said, three verified behavioral defects stand out — two of them user-facing regressions on common paths.

## Top findings

### 1. The 120s watchdog force-disconnects healthy connections waiting on the workspace-approval dialog — *verified*

Switch-producing `manage_workspaces` ops are now `.bounded(120s + 5s grace)` (`MCPTimeoutPolicy.swift:9-15`), but `create` (watchdog on by default since `switch_to_created` defaults true) and `delete` with `close_window` block on an interactive user approval first. `WorkspaceApprovalManager.requestApproval` (`Sources/RepoPrompt/Infrastructure/MCP/WorkspaceApproval/WorkspaceApprovalManager.swift:59-87`) parks on a plain `withCheckedContinuation` with no timeout and no cancellation handling, so it cannot observe the watchdog's cancel. If the user doesn't answer within 120s: the deadline fires, the 5s grace expires (the operation can't unwind), `cleanupUnresponsive` escalates to `abortConnectionForExecutionWatchdog` (`MCPConnectionManager.swift:10498-10539`) and the client's connection is severed — while app and connection are perfectly healthy, merely waiting for a human. Worse, the abandoned handler keeps running: when the user eventually clicks Approve, the workspace is created/switched *after* the tool call failed and the connection died. On `main` these ops were unbounded, so this is a regression introduced by the watchdog itself.

**Suggested fix:** make `requestApproval` cancellation-aware (resolve `.denied` via `withTaskCancellationHandler`), or exclude the approval phase from the bounded region the way elicitation waits are treated.

### 2. "Workspace Switch Blocked" modal on benign no-op switches, including at app launch — *verified*

`requestWorkspaceSwitch` now routes every `.blocked` result through `userVisibleWorkspaceSwitchResult`, which unconditionally publishes a notice (`WorkspaceManagerViewModel.swift:2181-2185, 2213-2224`) that renders as a real `.alert("Workspace Switch Blocked", ...)` (`WorkspaceSwitchingModels.swift:240-263`). Three routine callers hit the same-ID branch:

- **Launch restore**: `WindowState.restoreWorkspace` (`WindowState.swift:779`) calls `requestWorkspaceSwitch(to:, reason: "restore")` with no same-ID guard, and it only runs after `isInitialized` (`WindowState.swift:760`) — i.e., after init has already activated the Default workspace. Anyone who quits while on the Default workspace gets a *"Blocked: Already on workspace \"Default\""* modal **on every launch**.
- **⇧⌘S** (save-and-exit-to-fallback) while already on the system workspace.
- **MCP `workspace switch`** to the current workspace — including the documented smoke command — now pops a modal in the app.

**Suggested fix:** treat same-ID as silent success (or silent no-op), or publish the notice only for actionable blockages (concurrent switch, stale, chat-busy).

### 3. Malformed `switch_to_created` performs a switch but escapes the watchdog — *verified against the pinned SDK*

The contract classifier (`MCPToolExecutionContract.swift:123-128`) treats a present-but-non-bool `switch_to_created` (e.g. the string `"true"`, plausible from an LLM client) as *not* switch-producing because `Value.boolValue` is strict. The runtime handler (`WindowRoutingService.swift:2385`) does `args["switch_to_created"]?.boolValue ?? true` — so the same call **performs the full workspace switch with no watchdog**, on exactly the path this PR set out to bound. The new "create malformed flag" test in `MCPToolExecutionContractTests.swift` pins the divergence as expected behavior rather than catching it.

**Suggested fix:** share one predicate between contract and handler (handler rejects non-bool, or contract mirrors `?? true`). Note: fixing #1 first matters, otherwise aligning this *adds* the dialog-kill behavior of #1 to more calls.

### 4. Test seams compiled into release builds

`workspaceSwitchPhaseDidChangeHandlerForTesting` and `workspaceSwitchRecoveryWillBeginHandlerForTesting` (`WorkspaceManagerViewModel.swift:463-464`, awaited in the production recovery path at `:2355-2365`) are not `#if DEBUG`-gated — unlike every other seam in this PR and on main (the MCP, store, search-lane, and server-VM hooks are all gated). Nil by default so low functional risk, but they ship in release and break an otherwise uniform repo pattern. Same, lesser: `debugFinishApplyingHandler` in `WorkspaceFileSystemIngressCoordinator.swift:135`.

## Minor

**MCP / watchdog**
- Single-claim-per-session-token admission can lock a fast-reconnecting client out for up to the 60s reservation TTL when a prior claim is stranded, and reuses the `capacityExceeded` error code for "already pending" (`MCPConnectionManager.swift:1605-1620, 3850-3858`).
- Eviction paths recompute `creditedBootstrapPredecessorConnectionIDs()` and `effectiveActiveConnectionIDs` per candidate inside loops — quadratic actor work on the admission hot path under pressure (`:11848-11865, :12066-12100`). Bounded by the 128-connection cap, so performance-only.

**Workspace switching**
- Stale detection is report-only and counts user think-time: a confirmation dialog left open >30s is fault-logged as "stale", while a genuinely hung non-unload phase blocks all future switches with no force-clear path (`WorkspaceManagerViewModel.swift:2086-2109`).
- A blocked notice owned by switch A is only cleared if A *commits*; if A cancels or recovers, the alert outlives its blocker. The recovery-failure path also publishes the same message twice (`:1998-2007, :2209`).
- Double-fault recovery (switch fails post-disk-load, then every recovery target is blocked by chat-busy/refreshing) can leave `activeWorkspaceID` pointing at a partially hydrated workspace, surfaced only via the notice (`:2433-2438, :2615-2628`).
- `cancelCurrentWorkspaceSwitchAndReturnToSystem` silently drops the user's return-to-system intent once the op is committed/recovering instead of honoring it after completion (`:2814-2823`).

**Workspace context / unload**
- `terminateClosedPublisherIngress` can misreport a graceful drain as `.forced` (fault-level log) in a photo-finish, and *always* does so when the unload caller is cancelled, since `try? Task.sleep` returns immediately (`WorkspaceFileSystemIngressCoordinator.swift:419-435`). Diagnostics-only; check emptiness in `forceTermination` and/or add a `.cancelled` outcome.
- `finishPublisherIngress(rootIDs:)` is now dead production code with weaker (leak-prone) semantics than its replacement; only a test calls it (`:359-373`).
- Pre-existing but adjacent to this PR's theme: `unloadingRootPaths.insert` happens after an `await`, leaving a window where a same-path `loadRoot` is admitted mid-unload and the unload then deletes the new root's path mapping (`WorkspaceFileContextStore.swift:2660 vs 2690`). Cheap to close by hoisting the insertion above the first suspension.
- Behavior shift worth confirming as intended: freshness waiting now burns the queued broad-search's fixed 1500ms admission deadline, so under sustained ingress the queued search deterministically fails `waitExpired` (retryable) where it previously slow-succeeded (`StoreBackedWorkspaceSearch.swift:54-66`).

**Tests** (quality overall is high — injected clocks, exact-duration assertions, real post-conditions that fail if the fixes revert)
- `AsyncLimiterTests.swift` grew from 294 → 4,217 lines / 67 tests, most testing `ServerNetworkManager`, not `AsyncLimiter` — misleadingly named and serialized as one giant case; a few siblings mutate `ServerNetworkManager.shared`.
- Real flake risks: yield-count-bounded polling over real socket I/O followed by a hard assert (`MCPToolExecutionWatchdogIntegrationTests.swift:783-789`), and 1-second real-time bounds on real-socket operations (`PersistentMCPDistinctConnectionConcurrencyTests.swift`).
- `LimiterSnapshotSignal.waitUntil` and the `WorkspaceSwitchRecoveryGate` waits have no timeout — a regression hangs the suite instead of failing with diagnostics.

**Nits**: `@unchecked Sendable` where plain `Sendable` compiles (`WorkspaceSwitchingModels.swift:107`, `WorkspaceRootUnloadTermination.swift:4`); `.superseded` unload outcomes logged at fault level alongside `.forced`; ~390-line `performWorkspaceSwitch` with repeated boundary boilerplate; unreachable supersession-defense branches that the single-op invariant makes dead (`:2137-2139, :2190-2196`); uncancelled 5s grace-timer task in `MCPConnectionStopRace` (`MCPConnectionManager.swift:4228-4275`).

## Verified sound (explicitly checked, no issue)

Search-lane isolation and serialization both hold; no double-resume or leaked continuations in the stop race, start gate, or limiter waiters; the watchdog binds to the same captured arguments the handler receives and cancels its timers on every exit; no stuck-busy path in the switch state machine (begin/complete always pair, no throwing awaits between); recovery can't clobber a healthy switch; freshness coalescing has no lost-wakeup or dual-flight interleaving; late watcher completions are triple-fenced; diagnostics files are fully `#if DEBUG`-gated with no sensitive payloads; source placement follows the CLAUDE.md layout rules throughout.

## Summary

Findings 1–3 are the ones I'd fix before merge (the watchdog/approval interaction and the launch-time modal are user-facing regressions; the `switch_to_created` divergence is the one gap in the bound this PR exists to enforce), plus DEBUG-gating the two test seams. Everything else is non-blocking follow-up. The architecture itself is a clear improvement and the fixes should be small and local.

*Review produced with Claude Code (multi-agent pass over the four subsystems; top findings independently re-verified in source).*

baron:
Pushed 10483f5 addressing the blocking findings from the review above:

**1. Watchdog vs. approval dialog** — `WorkspaceApprovalManager.requestApproval` is now cancellation-aware: the wait is wrapped in `withTaskCancellationHandler`, so a watchdog cancel resolves the approval as `.denied` and the operation unwinds within the 5s cleanup grace instead of escalating to `cleanupUnresponsive` → force-disconnect. A new `cancelPending(requestID:)` denies and clears the specific request (active or queued) and advances the queue; the pre-registration race is closed by the `Task.isCancelled` check inside the continuation body (both paths serialize on the MainActor, so no double-resume window). This also eliminates the late side effect where a workspace was created/deleted after the tool call had already failed.

**2. Launch-time "Workspace Switch Blocked" modal** — the same-ID branch of `requestWorkspaceSwitch` now returns `.blocked(...)` silently (matching pre-PR behavior) instead of publishing the alert notice. Launch restore onto the already-active Default workspace, ⇧⌘S on the fallback workspace, and MCP `workspace switch` to the current workspace are quiet no-ops again; the notice remains for actionable blockages (concurrent switch, refresh-in-progress).

**3. `switch_to_created` contract/handler divergence** — the contract classifier now mirrors the handler's `?? true`: a present-but-malformed flag is classified as switch-producing and stays under the 120s deadline (safe now that finding 1 makes the bounded region approval-tolerant). The `MCPToolDurationInventory` condition string and the contract tests were updated; "create malformed flag" moved from the unbounded to the bounded expectations.

**4. Ungated test seams** — `workspaceSwitchPhaseDidChangeHandlerForTesting` and `workspaceSwitchRecoveryWillBeginHandlerForTesting` (storage, setters, and all call sites) are now `#if DEBUG`-gated like every other seam. (The ingress coordinator's `debugFinishApplyingHandler` flagged in the review turned out to already be fully gated — that part was a false positive.)

**New/updated tests**
- `WorkspaceApprovalCancellationTests` (new): cancelling the awaited active approval resolves denied and clears the overlay; cancelling a queued approval leaves the active request undisturbed; an already-cancelled caller resolves denied without presenting the overlay.
- `WorkspaceSwitchPresentationTests`: new `testActiveWorkspaceRequestIsBlockedWithoutPublishingNotice`; the two tests that previously *relied* on same-ID blocks publishing notices now build their notices from the refresh-blocked branch.
- `WorkspaceSwitchRecoveryTests.testSaveAndExitReturnsExactRequestResult` additionally asserts no notice is published.

**Validation**
- `make dev-format` / `make dev-lint`: clean, 0 violations.
- Focused suites (approval, presentation, recovery, contract, inventory): 31 tests, 0 failures.
- Full coordinated suite: 1,377 tests. Two runs each showed 1–2 failures in `WorkspaceFileContextStoreTests` / `WorkspaceSwitchRecoveryTests` — but a control run of the **unmodified PR head** on the same machine also failed (`testConcurrentStaleRequestReportsActiveSwitchAndForcedUnloadEventuallyClearsGuard`), with a *different* test failing each run. All three implicated tests pass repeatedly in focused runs (3× each). These are the load-sensitive yield-bounded-polling patterns called out in the review's test-quality section, pre-existing on the branch and not introduced by this commit — worth a follow-up but orthogonal to these fixes.
- Commit and push contribution preflights: passed.

*Fixes produced with Claude Code.*

baron (dismissed):
Approving after a full multi-agent review (see review comment above) and direct fixes for all four pre-merge findings in 10483f5: cancellation-aware approval waits (watchdog no longer force-disconnects healthy connections parked on the dialog), silent same-ID switch no-ops (no more launch-time blocked modal), contract/handler alignment for switch_to_created, and DEBUG-gated test seams. Lint clean, focused suites green, full-suite flakes confirmed pre-existing via a control run on the unmodified head.

Issue #172, opened 2026-06-12
MCP tool throughput remediation plan


provencher:
# Tracking status

This issue is the durable roadmap for the MCP tool-throughput remediation effort.

- **PR #173 delivery boundary:** WI-1 through WI-13
- **Still open:** WI-14 shared physical-root service across windows
- **Conditional:** WI-15 completion-publication decoupling, only if correlated measurements still justify it
- **Relationship:** PR #173 references this issue but does not close it because WI-14/WI-15 remain unresolved.

## Canonical plan and checked-in evidence

- **Canonical plan:** this issue body; the plan is intentionally not committed as a repository file.
- [WI-3 pre-optimization baseline](https://github.com/repoprompt/repoprompt-ce/blob/feature/mcp-tool-throughput-remediation/docs/investigations/mcp-tool-throughput-wi3-baseline-2026-06-11.md)
- [Implementation PR #173](https://github.com/repoprompt/repoprompt-ce/pull/173)

The baseline records the original admission topology, Git command counts (status 3; summary 6; quick 7; standard/deep 15 with one untracked file), global invalidation behavior, read-file amplification, and the deferred live-latency measurements. The plan below captures the evidence-backed background, shared invariants, sequencing gates, done-when criteria, validation, and risk inventory for all 15 work items.

### Phase 0 — correctness and measurement

- Closed watcher crawl/start and failed-recovery correctness gaps.
- Added request-correlated permit, MainActor, observer, publication, transport, and work-count diagnostics.
- Checked in the WI-3 baseline used to size later changes.

### Phase 1 — remove proven amplification

- Selective catalog/scope eviction rather than global clearing.
- Request-scoped Git context and single artifact-input construction.
- Indexed transcript correlation and incremental live-ID maintenance.
- Interactive decoded-content reuse and root-narrowed freshness.

### Phase 2 — reduce contention, then open scheduling

- Moved provider projection work off MainActor.
- Consolidated Git commands and added bounded global/per-repository concurrency.
- Replaced the ordinary limit-1 bottleneck with exhaustive static classification and resource-keyed capacities while preserving exclusive mutation/lifecycle ownership.

### Phase 3 delivered in PR #173

- Immutable per-root catalog shards with atomic publication, ARC-retained old generations, deterministic scope composition, DEBUG shadow parity, and an eight-generation retention backstop.
[...]

Issue #161, opened 2026-06-11
Flaky on CI: WorkspaceFileContextStoreTests ScopedAppliedIngress cancellation tests


baron:
## Summary

Two tests in the `ScopedAppliedIngress` cancellation family of `WorkspaceFileContextStoreTests` fail intermittently on CI runners, breaking unrelated PRs and main itself. Both pass consistently locally (Apple Silicon), so the race needs slower CI runner timing to surface.

## Failure samples (both on 2026-06-11)

**1. main @ 34dc4af** (merge of #148, which only touches OpenCode launch-resolver files — unrelated):
- Run: https://github.com/repoprompt/repoprompt-ce/actions/runs/27323411332
- `testScopedAppliedIngressAllCancelledCallersStillReachIdleCleanup` — `WorkspaceFileContextStoreTests.swift:1559`: `XCTAssertEqual failed: ("1") is not equal to ("0")`

**2. PR #116 @ 3ad28f0** (GLM availability notification in the settings view model — also nowhere near this store):
- Run: https://github.com/repoprompt/repoprompt-ce/actions/runs/27323616815
- `testScopedAppliedIngressCancelledPendingCallerDoesNotCancelLiveCoalescedJoiner` — `WorkspaceFileContextStoreTests.swift:1458`: `XCTAssertEqual failed: ("2") is not equal to ("1")`

Two different sibling tests, each off by one in a count assertion, on two unrelated changes — a timing-dependent race in the tests (or the coalescing/cancellation paths they exercise), not a regression from either change.

## Evidence it does not reproduce locally

On the exact PR #116 head (3ad28f0): full `swift test` suite passed twice, and `WorkspaceFileContextStoreTests` (138 tests) was run 3 additional consecutive times — all green.

## Pointers

The `testScopedAppliedIngress*` tests live at `Tests/RepoPromptTests/WorkspaceContext/WorkspaceFileContextStoreTests.swift:1390` and `:1467` (current main). The suite was last touched by the MCP-concurrency hardening work:
- 0943899 "Harden MCP concurrency and workspace switching"
- 1b389a4 "Harden file freshness and MCP cancellation"

The assertions appear to count cancelled/coalesced callers reaching cleanup; on slow runners an extra (or missed) caller slips through the cancellation window.

## Impact

Until stabilized, these will keep red-flagging unrelated PRs (and main) at random, costing a ~16-minute CI rerun each time.

baron:
[...]

PR #176, opened 2026-06-12
Integrate MCP throughput remediation into staging

baron:
## Summary

Integrates the MCP tool-throughput remediation work into `staging`, including the integration and live-evaluation fixes required to preserve correctness under real Agent Mode concurrency.

This is the implementation changeset tracked by #172. The full phased plan, evidence, sequencing gates, and deferred follow-ups live in that issue; the local plan document is intentionally **not** committed to the repository.

### Throughput and watcher work

- adds correlated MCP request timing/work-count diagnostics and the checked-in baseline
- removes proven read/search/Git amplification and narrows cache invalidation
- moves provider projection off `MainActor`
- consolidates/bounds Git command execution and introduces classified admission
- adds immutable per-root workspace catalog shards, canonical delta application, and per-root path indexes
- fixes watcher recovery correctness before incremental indexing
- filters safely ignored FSEvents before the bounded ingress queue, preventing build churn under `.build`, `node_modules`, and equivalent ignored trees from forcing overflow/root rescans
- removes the CE-only five-minute known-file safety scan behavior; reliable watcher events and targeted ambiguity scans remain the correctness mechanism

### Agent Mode and tool-contract fixes

- honors caller-supplied long `agent_run` timeouts instead of clipping them at 300 seconds
- removes deadlines for Context Builder and Oracle operations
- removes the explore-agent Bash restriction and deletes the source-shape-only shell policy test
- fixes Git worktree scope, artifact auto-selection, Context Builder export routing, and active-turn transcript correlation
- preserves explicit and automatic file selection across worktree projection, disconnects, pending-policy handoff, child lifecycle, and peer-window lookup
- makes ranged `read_file` select the exact slice and full-file reads select the full file
- replaces role-name routing special cases with exact run/tab/PID/lifecycle ownership
- uses exact workspace UUID + tab UUID identity, revision ordering, host-generation fencing, and close/replacement guards for cross-window canonical selection

## Live failures found and fixed

The `rpce-cli-debug → agent_run` smoke found production-composition failures that the earlier unit tests did not reproduce:

1. A completed child could report a correct selection internally while an independent connection resolved an empty peer-window model.
2. Explicit `manage_selection set` could return `-32603` with `expected selected=1` but `canonical selected=0`.
3. Full-file `read_file` selection could be lost across pending-policy connection replacement.

The final explicit-set failure was correctly routed to the expected child PID/window/workspace/tab. Its root cause was a queued stale active-UI snapshot overwriting the exact MCP canonical write before revision-aware presentation mirroring began. The final fix installs the existing revision fence before active and deferred MCP presentation; strict canonical verification remains unchanged.

Regression coverage now includes the real socket/pending-policy path, independent post-completion lookup, exact peer identity, two-source ordering, manager replacement/reopen, closing peers, worktree-only paths, and stale UI interleavings.

## Validation

Final local validation on head `4b4b19f8275eadd7272ba2723f1a1efb831c3ca4`:

- contribution push preflight passed
- full coordinated root test suite passed
- `RepoPrompt` build passed
- `repoprompt-mcp` build passed
- SwiftFormat: `0/1151` files require formatting
- SwiftLint strict: `0` violations
- repository guardrails passed
- staged and outgoing-range secret scans passed
- focused final selection suites passed:
  - `WorkspaceSelectionCoordinatorTests`: 21 tests
  - `TabContextRoutingTests`: 31 tests
  - persistent Agent Mode MCP selection suite: 12 tests
- final live CE debug smoke passed through a fresh Agent Mode child:
  - explicit `manage_selection set/get`: 1 full file, 3,094 tokens
  - ranged `read_file`: exact slice lines 2–3, 11 tokens
  - full `read_file`: full file selected, 3,094 tokens
  - independent persistent CLI connection explicitly rebound to the child context and observed the same final full-file selection
  - `agent_run wait` accepted caller timeout `600`

During earlier integration attempts, the full suite parked around an XCTest/AppKit boundary and the live selection smoke exposed the failures above. Those observations were not hidden or treated as success. On the final head, the complete local push-preflight root suite finished successfully and the live smoke passed. The exact-head hosted CI run completed Secret Scan and Style successfully and built both app products, but its `Run app tests` step remained parked for more than fifteen minutes; this matches the documented hosted/full-suite park rather than a reported test assertion failure.

## Tracking and follow-ups

- Canonical implementation plan and remaining follow-ups: #172
- This PR intentionally does not commit `docs/plans/mcp-tool-throughput-remediation-2026-06-11.md`.
- Follow-up candidates in #172 include evaluating bounded search freshness waits and batching disk metadata validation off the contended filesystem actor, based on captured diagnostics rather than assumption.

## Merge scope

This PR targets `staging`. After exact-head validation, `staging` will be merged to `main` as a separate guarded merge.


baron:
*Comment from Claude (Fable 5), the AI model that performed the deep integration review and co-drove the staging validation with @baron. Honest thoughts, advice, and a short retrospective for the merge decision.*

## Assessment

This changeset is a marked, verifiable improvement over the staging baseline, and I would land it. The hot paths heavy agent workloads actually hit — read caching (baseline had 0% cache hits by design), per-root search shards, consolidated/bounded git execution, cross-session admission, incremental transcript correlation — are exactly what this PR fixes, and the discipline in the core primitives held up under adversarial review: admission lease pairing is airtight on all exit paths, the read cache is fingerprint-validated (no stale-read window beyond inherent TOCTOU), shard publication is atomic and generation-gated, and the porcelain-v2 parser checked out against real `git status -z` output.

Four review findings were fixed in-branch and are worth a reviewer's attention as the riskiest-then-best parts of the diff: the `control` admission lane (the original classification made `agent_run`/`ask_user` hold the per-window exclusive slot across unbounded waits — a cross-session starvation regression), the parallel-scan snapshot/restore (delta loss on mid-batch abort was a permanent-staleness bug), recovery resync correctness (unreadable-dir handling + preserving explicitly managed ignored files), and mid-run steering tool-card correlation.

## Known trade-offs the merge should accept knowingly

- **Persistently unreadable directories now park freshness barriers** instead of silently deleting catalog subtrees (the old behavior). Correctness over availability — deliberate. The follow-up design we converged on: a stalled-recovery state owned by the watcher, checked at the single barrier choke point, failing fast with a typed error — *not* per-call-site timeouts, which would pin per-window admission slots and feed agent retry amplification. Recommend filing that as its own reviewed item.
- **Deliberate throughput ceilings** pending live data: small-read cap 2/window, `bind_context` serialized behind `manage_workspaces` on the appWide slot, per-event shard rebuild on multi-record batches, and the uncached fallback regime at the 8-generation shard cap. These are measurement targets for dogfooding, not blockers — tune with eval numbers, not guesses.
- **Git residuals:** the per-repo tool lease is held across the post-publish ingress wait, and multi-repo requests can be starved by sustained single-repo traffic. The hard wedge is gone (120s subprocess kill-switch); these are load-dependent latency follow-ups.
- The full-suite XCTWaiter parking remains unexplained (focused reruns pass); worth keeping #161-adjacent until root-caused.

## Retrospective

The recurring lesson from this integration: every consequential issue lived in a tiny primitive at a funnel point — one classification table entry, one swallowed catch, one index reset on user-append, one per-window constant. Static review at the funnel found mechanisms in hours that live monitoring would have surfaced only as symptoms. For future trains (the core-isolation work especially), I'd repeat this shape: adversarial subsystem review *before* the live eval, fix-then-regression-test in small commits, and defer anything that changes shared semantics to its own reviewed item — the restraint mattered as much as the fixes.

Net: recommend merge, cut the release behind real dogfooding of the eval metrics, and carry the two follow-up issues above.

Issue #164, opened 2026-06-11
MCP Server status popup: Trusted Clients "Remove" button is a silent no-op for built-in default clients


BetaFold3:
**Summary**
In the MCP server status popup, the Trusted Clients section shows an enabled "Remove" button on every row, but clicking it has no effect for any of the built-in default clients (claude-code, codex-mcp-client, gemini-cli-mcp-client, opencode, cursor, cursor-mcp-client, claude-ai). Since those defaults are exactly the clients that typically populate the list, Remove appears completely broken: no list change, no persistence change, no feedback.

The root cause is a UI-vs-policy mismatch, not a broken refresh pipeline: defaults are deliberately un-removable in ServerController.setAlwaysAllowed, which early-returns before any mutation or dashboard notification — but the UI renders an enabled Remove button for them with no indication they are built-in.

Note: this concerns the popup section backed by the MCP connection allow-list (MCPStatusView). Settings → Permissions → Trusted Clients is a separate per-operation approval subsystem (WorkspaceApprovalManager) and is not affected.

**Evidence**
All references at 34dc4af.

The popup renders every allow-list entry with an always-enabled Remove button, with no distinction between built-in and user-approved clients:

https://github.com/repoprompt/repoprompt-ce/blob/34dc4af2ce50122a8bf70af1e1d2d605daa80159/Sources/RepoPrompt/Infrastructure/UI/Components/MCPStatusView.swift#L397-L441

```
let allowedClients = server.dashboard?.alwaysAllowedClients ?? []
// ...
Button(action: { server.setAlwaysAllowed(clientID: client, allowed: false) }) {
    Text("Remove")
```

The remove path early-returns for default clients before any set mutation, UserDefaults write, or notifyDashboardUpdate():

https://github.com/repoprompt/repoprompt-ce/blob/34dc4af2ce50122a8bf70af1e1d2d605daa80159/Sources/RepoPrompt/Infrastructure/MCP/ServerController.swift#L309-L326

```
func setAlwaysAllowed(clientID: String, allowed: Bool) async {
    if !allowed, isDefaultAlwaysAllowed(clientID) {
        return
    }
    if allowed {
        addAlwaysAllowed(clientID: clientID)
    } else {
        alwaysAllowedClients = alwaysAllowedClients.filter {
            !MCPClientIdentity.matches($0, clientID) || isDefaultAlwaysAllowed($0)
        }
        UserDefaults.standard.set(
            Array(alwaysAllowedClients),
[...]

PR #178, opened 2026-06-12
Show a Built-in badge instead of a no-op Remove button for default trusted clients

BetaFold3:
**Fixes** [#164](https://github.com/BetaFold3/repoprompt-ce/issues/164)

In the MCP server status popup, every Trusted Clients row showed an enabled "Remove" button, but ServerController.setAlwaysAllowed deliberately early-returns for the built-in default clients — so for the clients that typically populate the list, Remove was a silent no-op: no list change, no persistence, no feedback.

Defaults are meant to be permanently trusted, so this makes the UI reflect that policy instead of pretending removal is possible:

`MCPStatusView`: built-in rows now show a "Built-in" capsule with a hover tooltip ("Built-in clients are always trusted and can't be removed") instead of the Remove button. Custom clients keep the working Remove button.
`ServerController`: exposed the existing default-client classification as an internal static isBuiltInAlwaysAllowedClient(_:) (exact or same-family match); the private instance check now delegates to it. No behavior change.
`MCPServerViewModel`: added isBuiltInTrustedClient(_:) so the view keeps talking only to the view model.
`ServerControllerAdmissionTests`: new test covering the classification, including family variants ("Claude Code v2.1", "cursor-agent") and negatives ("my-custom-client", "RepoPrompt CLI").

**Validation**

- `make guardrails` — pass
- `make dev-lint` — 0 violations
- `make dev-test FILTER=ServerControllerAdmissionTests` — 5/5 pass
- `make dev-swift-build PRODUCT=RepoPrompt` — pass
- Verified live via make dev-run: launched debug app, MCP server responding

rpce_mcp_status_trusted_clients_builtin


BetaFold3:
@provencher @baron friendly ping 👋

BetaFold3:
^ the conflicts appeared lately are easy to address once this PR gets a go-ahead from maintainers..

provencher (approved):
Reviewed after conflict resolution, accessibility repair, cleanup pass, and exact-head validation.

PR #180, opened 2026-06-12
Show workspace names in window and native tab titles

nickdavis:
## Problem

Window titles and native tab names (including Merge All Windows mode) always show the app name ("RepoPrompt CE") instead of the active workspace name. A workspace named `ta-sa` displays as "RepoPrompt CE"; tabs with Agent sessions sometimes show a truncated session name with no workspace.

## Root cause

`WindowState` resolves a workspace-aware window title and applies it to the `NSWindow` directly, but the `WindowGroup` scene declares no title of its own, so SwiftUI keeps re-applying the default app-name title over the AppKit-set value whenever it refreshes the window chrome. The hook intended to counter this (`reassertWindowTitle`) has no callers, and `WindowAccessor` intentionally installs no observers — so once SwiftUI stomps the title, nothing restores it until the next workspace switch.

## Fix

- `WindowState` now publishes the resolved title (`displayedWindowTitle`), and `WindowContentView` binds it at the scene root with `.navigationTitle`, so SwiftUI owns the title and re-applies the correct value instead of the app-name default. The existing AppKit title application is kept for immediate updates.
- The title also refreshes when the window's active compose tab changes, so the Agent session portion of the composed title does not go stale.
- No-workspace fallbacks use the bundle display name instead of the hardcoded "Repo Prompt" (which predates the CE rename).

## Validation

- `make guardrails`, `make dev-lint` (0 violations), `git diff --check`
- New focused test: `make dev-test FILTER=WindowStateDisplayedTitleTests`
- Related suites: `DockMenuTests`, `WindowCloseCoordinatorDecisionTests`
- Full `make dev-test`: 0 failures
- `make dev-swift-build PRODUCT=RepoPrompt` (push preflight)
- Contribution preflight passed in `commit` and `push` modes
- Verified live in the debug app: window titles and merged native tabs show e.g. "SearchStax Viability — kh", "T1 — parentagent", and follow `workspace switch` via the MCP debug CLI


antonioc-cl:
This looks like the right root-fix direction. We independently investigated the same regression and reached the same broad conclusion: dynamic titles already exist, but title ownership/refresh was incomplete.

One possible gap to verify before merge: session rename / MCP `set_status`.

From our investigation, both sidebar rename and MCP `set_status` route through:

`AgentModeViewModel.renameSession -> PromptViewModel.renameComposeTab`

The local fix we prototyped added a `.composeTabNameChanged` notification posted from `renameComposeTab(...)`, then had `WindowState` refresh the title for the matching `windowID`.

This PR appears to cover SwiftUI title ownership and active compose tab changes, which is likely the more fundamental fix. Could you confirm it also updates immediately when:
1. active Agent session is renamed from the sidebar
2. MCP `set_status` renames the active session

If not, I think the minimal follow-up would be:
- post a compose-tab-name-changed notification from `PromptViewModel.renameComposeTab`
- have `WindowState` refresh/reassert title for matching `windowID`
- add regression coverage for formatter/rename behavior

Otherwise, this PR likely supersedes our separate `fix/window-title-refresh` branch.

nickdavis:
Thanks for flagging this. I verified the rename gap and pushed a follow-up in `a498099`.

The fix now posts `.composeTabNameChanged` from `PromptViewModel.renameComposeTab`, and `WindowState` refreshes/reasserts the title only when the notification matches the same `windowID` and the currently active `tabID`. Since both sidebar rename and MCP `set_status` route through `AgentModeViewModel.renameSession -> PromptViewModel.renameComposeTab`, this covers both paths you called out.

Confirmed:
- Active Agent session rename from the sidebar updates the window/native tab title immediately.
- MCP `set_status` rename of the active session updates the title immediately.
- Extra guard checked: renaming a non-active/background session does not change the active window title until that session is selected.

Validation:
- `make guardrails`
- `make dev-lint`
- full `make dev-test`
- live smoke after relaunch

baron (approved):
Reviewed the PR after merging current main and adding focused regression coverage for workspace/session window titles. Local push preflight passed cleanly after clearing stale RPCE test runners.

baron:
@nickdavis @antonioc-cl thank you so much for working on this fix. I took the liberty of getting this merged (pulling changes from main and making sure the titles were still fixed) since main was accumulating a lot of commits. It'll be in the next release (probably by the time you read this).

nickdavis:
@baron pleasure! Pumped to get my first commit in there :)

Issue #156, opened 2026-06-10
ACP sub-agent launch fails for Homebrew-installed CLIs on Apple Silicon (untrustedWritableDirectory on /opt/homebrew/Cellar)


itsniper:
## Summary

After updating to v1.0.13, ACP-routed sub-agents (e.g. `agent_run` with `model_id: "explore"` → OpenCode, and presumably Cursor-agent too) fail immediately with:

```
OpenCode CLI was not found as a valid executable regular file for `opencode`. Install OpenCode or configure its absolute path.
```

This happens even though `opencode` is correctly installed and runnable from a shell:

```
$ which opencode
/opt/homebrew/bin/opencode
$ opencode --version
1.16.2
```

## Root cause

This is a regression from PR #97 ("Fix ACP mode negotiation and OpenCode routing", commit `ae8d305` "Harden ACP startup and routing races"), **not** PR #128 (which is an unrelated, correct fix for the `notifications/initialized` ACP handshake + `~/.opencode/bin` path hint, also in v1.0.13).

PR #97 added `ExecutableFileIdentity.captureForTrustedPathLaunch(atPath:)` / `validateTrustedOwnershipAndPermissions(atCanonicalPath:)` in `Sources/RepoPrompt/Infrastructure/Process/ExecutableFileIdentity.swift`. After `CommandPathResolver` resolves `opencode` to its canonical path, this new check walks **every directory** in that canonical path and throws `untrustedWritableDirectory` if any directory is group- or world-writable (unless it's a root-owned sticky directory like `/tmp`):

```swift
let isGroupOrWorldWritable = directoryInfo.st_mode & mode_t(S_IWGRP | S_IWOTH) != 0
let isRootOwnedStickyDirectory = directoryInfo.st_uid == 0
    && directoryInfo.st_mode & mode_t(S_ISVTX) != 0
guard !isGroupOrWorldWritable || isRootOwnedStickyDirectory else {
    throw ExecutableFileIdentityError.untrustedWritableDirectory(directory.path, directoryInfo.st_mode)
}
```

On Apple Silicon, the **default Homebrew install layout** symlinks bins into the Cellar:

```
/opt/homebrew/bin/opencode -> /opt/homebrew/Cellar/opencode/1.16.2/bin/opencode
```

[...]

PR #208, opened 2026-06-14
Improve Cursor Agent CLI launch resolution diagnostics

dlcen:
## Summary

• preserve Cursor CLI-only launch policy: still requires `cursor-agent` and still rejects `cursor`/.app descendants
• canonicalize executable paths with POSIX path APIs instead of Foundation URL path normalization
• try discovered `cursor-agent` candidates instead of failing on the first invalid one
• surface candidate-specific validation failures instead of collapsing all failures to "not found"
• replace remaining Foundation URL path usage in `CursorACPLaunchResolver` with `NSString` methods

Refs #156.

## Non-goals

• does not restore Classic `cursor agent acp` fallback
• does not relax Homebrew trust policy
• does not address Cursor MCP injection tracked separately

## Validation

• `make guardrails` ✅
• `make dev-test FILTER=CursorACPLaunchResolverTests` (18 non-MCP tests pass) ✅
• `make dev-lint` ❌ — fails due to Python 3.7 typing.Protocol incompatibility in Scripts/conductor.py (pre-existing environment issue,not caused by this change)
• `git diff --check` ✅

## Notes

• 5 MCP-related tests hang (pre-existing issue, not caused by this change)
• Ad-hoc signed debug build tested locally: Cursor CLI launches successfully

### DIFF
--- a/Sources/RepoPrompt/Infrastructure/Process/ExecutableFileIdentity.swift
+++ b/Sources/RepoPrompt/Infrastructure/Process/ExecutableFileIdentity.swift
@@ -16,8 +16,9 @@ struct ExecutableFileIdentity: Equatable {
             throw ExecutableFileIdentityError.pathMustBeAbsolute(rawPath)
         }
 
-        let canonicalPath = FileSystemService.realpathString(rawPath)
-            ?? (rawPath as NSString).standardizingPath
+        guard let canonicalPath = FileSystemService.realpathString(rawPath) else {
+            throw ExecutableFileIdentityError.unavailable((rawPath as NSString).standardizingPath)
+        }
         var info = stat()
         guard stat(canonicalPath, &info) == 0 else {
             throw ExecutableFileIdentityError.unavailable(canonicalPath)
--- a/Sources/RepoPrompt/Infrastructure/Process/ExecutableFileIdentity.swift
+++ b/Sources/RepoPrompt/Infrastructure/Process/ExecutableFileIdentity.swift
@@ -16,10 +16,7 @@ struct ExecutableFileIdentity: Equatable {
             throw ExecutableFileIdentityError.pathMustBeAbsolute(rawPath)
         }
 
-        let canonicalPath = URL(fileURLWithPath: rawPath)
-            .resolvingSymlinksInPath()
-            .standardizedFileURL
-            .path
+        let canonicalPath = canonicalizePath(rawPath)
         var info = stat()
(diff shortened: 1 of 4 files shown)


Sources/RepoPrompt/Infrastructure/Process/ExecutableFileIdentity.swift
```diff
@@ -43,6 +40,19 @@ struct ExecutableFileIdentity: Equatable {
         )
     }
 
+    private static func canonicalizePath(_ rawPath: String) -> String {
```
provencher: i really question the need for this. perhaps your path is weird?

dlcen: Good point. I narrowed this down a bit: I removed the bespoke `canonicalizePath` helper and reused the existing `FileSystemService.realpathString(_:)` helper instead, with the same `NSString.standardizingPath` fallback.

I did keep realpath-based canonicalization because fully reverting this locally caused the Cursor ACP launch resolver tests to fail on SwiftPM temporary paths under `/var/folders/...`; the resolver ended up reporting the executable as unavailable after Foundation URL path normalization.

After the update, I rebuilt the ad-hoc debug app locally and verified that Cursor Agent launch still works in practice.

dlcen: I did some more local testing and wanted to share a bit more precise context in case it helps.

For Cursor Agent, the path seems fairly normal on my machine: `cursor-agent` is available from the shell, resolves under my home directory to `~/.local/bin/cursor-agent`, is not inside a .app bundle, and keeps the expected `cursor-agent` basename.

With the current Homebrew/release CE app, Cursor still fails with:
> Cursor Agent CLI was not found as a valid executable regular file for cursor-agent.

I built this PR branch locally and tested the debug app with the same Cursor Agent install/path, and it connects successfully.

So at least locally:

• release CE app + same Cursor Agent path: fails
• this PR build + same Cursor Agent path: connects

I also saw a similar “not found as a valid executable” symptom with OpenCode, where the release app fails but this PR build connects.

Of course, happy to adjust the scope or approach if you’d prefer a different shape for the fix.

provencher (approved):
Reviewed after rebase onto current main. Addressed fail-open canonicalization and cross-provider path expectation regressions; pair cleanup complete. Local full suite, strict lint, app build, and fresh exact-head CI all pass.

PR #221, opened 2026-06-15
Fix Agent Files preview loading state

### LINKED ISSUE #217: Agent Files preview can get stuck on Loading preview
The **Agent Files** preview popover can remain stuck on `Loading preview…` instead of rendering the selected file’s contents.

This appears to happen when previewing a file from the Agent Files list, especially for sliced selections. The preview popover opens and shows the correct file path, but the preview body never finishes loading.

## Screenshot

Image

## Expected behavior

When clicking the preview button for a file in **Agent Files**:

- The preview popover should open.
- The selected file’s contents should load and display.
- If the file cannot be previewed, the UI should show a clear fallback message such as “No preview content available.”
- The preview should not remain indefinitely in a loading state.

## Actual behavior

The preview popover opens, but the content area stays on:

```text
Loading preview…
```

The file path/title appears to be known, so the UI is able to identify the selected file, but the preview content never appears.

## Why this matters

The Agent Files list is used to review what context is being sent to an agent. If previews can get stuck, users cannot reliably inspect selected files or slices before continuing.

This makes it harder to verify agent context, especially when working with sliced selections where the user wants to confirm exactly what was included.

## Reproduction notes

Observed while previewing a file from the **Agent Files** panel in Agent Mode.

Example from the screenshot:

- Agent Files list contains multiple selected/sliced files.
- Clicking the preview button opens the preview popover.
- The popover title/path points to a Swift source file.
- The body remains stuck on `Loading preview…`.

## Current investigation status

We have confirmed this is likely a real lifecycle/state issue in the preview experience:

- The preview enters a loading state.
- In some lifecycle paths, loading can remain active instead of resolving to either file content or a fallback message.
- Missing/unreadable files should normally result in a fallback message, not an indefinite loading state.
- This means the issue is not just that a specific file cannot be read; the preview UI can fail to leave the loading state.

## Acceptance criteria

- Previewing a file from Agent Files reliably resolves to either:
  - rendered file contents, or
  - a clear fallback/error state.
- The preview popover should not remain indefinitely on `Loading preview…`.
- This should hold for both full-file selections and sliced selections.
- Add regression coverage for the stuck-loading case.

erauner12:
Closes #217.

- Fix Agent Files preview loading lifecycle by moving preview state into a small `AgentSelectedFilePreviewLoadCoordinator`.
- Clear preview loading state on row disappearance and explicit popover dismissal so late cancelled loads cannot leave stale loading UI behind.
- Add a coordinator `contentRevision` and pass it as `externalUpdateTick` to the preview `TextKitView`, forcing the AppKit text view to re-sync when loading transitions to loaded content.
- Add focused regression tests for normal completion, row-disappearance cancellation, and popover-dismissal cancellation.

## What was wrong
The original preview row kept `showPreview`, `previewText`, `isLoadingPreview`, and the load task as row-local SwiftUI state. Row disappearance cancelled the load task but did not clear the loading state/task reference. If the cancelled task resumed later, it skipped its cleanup path, so the preview could remain stuck on `Loading preview…`.

Visual retesting then exposed a second layer: even after the coordinator published loaded text, the popover rendered dynamic content through `TextKitView(text: .constant(displayText), ...)` without an update tick. If AppKit made the underlying selectable `NSTextView` first responder, `TextKitView.updateNSView` intentionally skipped external writes while `externalUpdateTick` was unchanged. That meant SwiftUI state could be correct while the visible text stayed frozen on the initial `Loading preview…` string.

## Why this fixes it
The coordinator makes the loading lifecycle explicit and testable:
- normal completion publishes `previewText`, clears `isLoadingPreview`, and drops the task reference;
- dismissal and row disappearance both cancel and clear loading state immediately;
- late cancelled completions cannot reinstall stale UI state.

The `contentRevision` counter advances whenever the preview display text can change. Passing that revision into `TextKitView` as `externalUpdateTick` uses the existing bridge contract to force programmatic text updates even when the underlying `NSTextView` is first responder. This preserves the existing read-only text view behavior while ensuring the popover transitions from `Loading preview…` to the loaded file content.

A short call-site comment documents that the `externalUpdateTick` wiring is load-bearing so future cleanup does not accidentally remove it.

## Visual confirmation

before:

image

after:
image

## Validation
Passed:
- `make dev-format`
- `make dev-test FILTER=AgentSelectedFilePreviewLifecycleTests`
- `make dev-format-check`
- `make dev-lint`
- `ALLOW_ADHOC_SIGNING=1 make dev-run`
- `make dev-smoke`
- commit preflight: `.agents/skills/rpce-contribution-check/scripts/preflight.sh commit`

Visual validation:
- Retested the Agent Files preview popover after rebuilding/relaunching the debug app.
- Confirmed the preview now displays `AgentRuntimeExportCard.swift` content instead of staying on `Loading preview…`.

Additional notes:
- Full `make dev-test` / push preflight root tests repeatedly failed in unrelated `AgentModeWorkspaceSwitchCleanupTests.testDetachedMCPTeardownDoesNotDeactivateNewLiveControlContextWithSameSessionID` with a timeout; that suite passed when run alone via `make dev-test FILTER=AgentModeWorkspaceSwitchCleanupTests`.
- Earlier non-disruptive `make dev-smoke` could not connect through `/usr/local/bin/rpce-cli-debug` before the debug app was relaunched. After `ALLOW_ADHOC_SIGNING=1 make dev-run`, `make dev-smoke` passed.


baron (approved):
Approved — the fix looks sound to me.

I reviewed the preview lifecycle change and the new coordinator/test coverage. The cancellation paths for dismissal/disappear look correct, the loading → loaded transition is covered, and I don’t see a blocking correctness or security issue.

Validation performed:

- `make dev-test FILTER=AgentSelectedFilePreviewLifecycleTests`
- Passed: 3 tests, 0 failures

Non-blocking suggestions for follow-up:

1. `AgentRuntimeExportCard.swift`: `contentRevision` is bumped from `@Published didSet` observers on the coordinator. Because the row owns the coordinator as a `@StateObject`, each bump invalidates the whole row body, not just the preview content. This seems acceptable for the active row, but it is a bit heavier than the more localized `@State` tick pattern used elsewhere.
2. Opening and completing a preview can bump the revision more than once per visible transition (`previewText` and `isLoadingPreview` both mutate). A small helper that updates preview state and bumps once when the rendered display text changes would reduce SwiftUI/AppKit update churn.
3. The `Task` body appears to inherit the coordinator’s `@MainActor` context already, so the inner `await MainActor.run { ... }` looks redundant and could be simplified.
4. Consider adding a test for “open preview A, open preview B before A resolves, then A completes late” to lock down stale completion cancellation behavior.
5. Consider adding a nil-result test after a prior successful preview load to ensure unreadable/empty preview results don’t leave stale text behind.

None of those are blockers for this PR.

PR #225, opened 2026-06-15
Fix agent preview retention architecture

baron:
## Summary

Follow-up to #221 after lead review flagged that per-row preview coordinators could retain preview state for many selected files.

This PR reworks the Agent selected-files preview path so it no longer scales retained preview state with row count:

- moves preview loading state from per-row `@StateObject`s to a single popover/card-level coordinator
- clears active preview text/task/row identity on dismissal and reconciles active state when rows change
- bounds preview reads to a prefix while keeping copy/export reads full-fidelity
- adds prefix loading through the file-system/content store path instead of materializing whole large files for preview
- clears TextKit preview backing strings during teardown
- adds lifecycle/regression tests for stale completions, row identity changes, dismissal, and truncation behavior

## Context

#221 fixed stale preview loading state, but after merge we identified a broader architecture concern: each row could own a coordinator that retained preview content, and preview loading could materialize full large files even though the UI only needs bounded preview text. This PR is intended as the proper follow-up architecture fix.

## Validation

- `make dev-test FILTER=AgentSelectedFilePreviewLifecycleTests` — passed, 6 tests
- `make dev-test FILTER=AgentContextExportResolverTests/testPreviewContentIsPrefixBoundedWhileCopyRemainsFullContent` — passed
- `make dev-lint` — passed, 0 violations
- `.agents/skills/rpce-contribution-check/scripts/preflight.sh commit` — passed before commit
  - outgoing range: `upstream/main..HEAD`
  - full coordinated `swift test` — passed
  - `swift build --product RepoPrompt` — passed

## Informal performance notes

I will add a separate PR comment with the informal local measurements from the synthetic preview probe and live RSS monitoring so the numbers are easy to find/discuss.

### DIFF
--- a/Sources/RepoPrompt/Features/AgentMode/Services/AgentContextExportResolver.swift
+++ b/Sources/RepoPrompt/Features/AgentMode/Services/AgentContextExportResolver.swift
@@ -128,6 +128,22 @@ extension AgentContextExportRow {
     }
 }
 
+enum AgentContextPreviewContentPolicy {
+    static let maximumBytes = 256_000
+    static let maximumCharacters = 200_000
+
+    static func boundedPreviewText(_ text: String, wasTruncated: Bool = false) -> String {
+        let exceedsCharacterLimit = text.count > maximumCharacters
+        guard wasTruncated || exceedsCharacterLimit else { return text }
+        let preview = exceedsCharacterLimit ? String(text.prefix(maximumCharacters)) : text
+        return """
+        \(preview)
+
+        … Preview truncated to avoid retaining large file content. Copy the file content for the full text.
+        """
+    }
+}
+
 struct AgentContextClipboardRequest {
     let cfg: PromptContextResolved
     let source: AgentContextExportSource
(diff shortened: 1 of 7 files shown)


baron:
Informal local performance/memory notes from the follow-up validation pass:

### Synthetic preview probe

Temporary XCTest probe used a real resolver/store path with an 8 MB synthetic text file, comparing the old-equivalent full-content path against the new bounded preview path.

Observed median results:

- preview path: **3.02 ms**
- full-content path: **49.37 ms**
- speedup: **16.33x**
- preview median bytes: **200,103 bytes**
- full median bytes: **8,000,000 bytes**
- byte reduction: **97.50%**

### Live debug-app RSS sampling

After launching the debug build from this branch, we selected/dragged around large files and sampled the app process RSS every ~5 seconds for ~3 minutes.

- PID sampled: `76349`
- samples: **36**
- start RSS: **316.1 MB**
- end RSS: **319.6 MB**
- net change: **+3.5 MB**
- peak RSS: **345.9 MB**
- low after cleanup: **294.9 MB**
- median RSS: **319.9 MB**
- linear trend: approximately **-0.43 MB/min**

Interpretation: the interaction produced transient allocation bumps, but did not show the retained upward staircase we were concerned about from per-row preview state / full-file preview materialization.

Caveat: this is deliberately informal local evidence, not a benchmark harness. The committed regression coverage is focused on bounding preview content, stale completion handling, row identity reconciliation, and dismissal cleanup.

PR #227, opened 2026-06-15
Dismiss hover tooltips when MCP toolbar buttons open a surface

0xNino:
## Problem

Clicking the MCP **connections counter** (the green link/count pill) or the **MCP Server** toggle in the toolbar could leave a hover tooltip floating over the freshly opened status window / popover.

Why it sticks:
- A hover tooltip's only auto-dismiss paths are hover-exit (`onHover(false)`) and the click/scroll/key event monitors — but those monitors are installed *only after* the tooltip becomes visible (after the 300 ms hover delay). A click that lands during that delay can't dismiss it.
- When the button opens a new window/popover, that surface appears under the cursor and steals focus, so the button's `onHover(false)` never fires.
- The tooltip overlay renders at `.floating` window level, so it sits *on top of* the newly opened surface — very visibly stuck until the next unrelated click/scroll/keypress.

A second, related failure: a stale/empty anchor geometry report (e.g. `.zero` during a layout/window transition) fed `owner.convertToScreen(.zero)`, snapping the bubble to the owner window's bottom-left corner.

## Fix

- Add `HoverTooltipCoordinator.dismissAll()`, which posts a notification observed by both `HoverTooltipModifier` and `TooltipOverlayController`. The modifier-side cleanup cancels a *pending-but-unshown* tooltip as well as hiding a visible one. The two MCP toolbar buttons call it before opening their surface.
- Guard `show()` / `reposition()` against empty or non-finite anchor rects — hide instead of mispositioning to the window origin.
- Refactor observer install/teardown so the new dismiss observer is cleaned up alongside the owner-will-close observer.

## Tests

Adds `TooltipOverlayControllerTests` covering dismiss-all and the invalid-anchor `show`/`reposition` paths.

## Validation

- `make dev-lint` — 0 violations (format-check + SwiftLint strict)
- `make dev-test` — full suite green (run via push preflight); new suite passes, stable across repeated runs
- `make dev-swift-build PRODUCT=RepoPrompt` — clean
- Contribution preflight passed in `commit` and `push` modes (staged-index + outgoing-range secret scans)

UI-only change; does not touch MCP CLI/server, Agent Mode, or packaging.

## Screenshot

This shows the issue: the MCP status surface opens while the hover tooltip remains floating over any app.

Screenshot 2026-06-15 at 15 09 53

Screenshot 2026-06-15 at 15 09 09

### DIFF
--- a/Sources/RepoPrompt/Infrastructure/UI/Components/MCPConnectionsCounter.swift
+++ b/Sources/RepoPrompt/Infrastructure/UI/Components/MCPConnectionsCounter.swift
@@ -29,6 +29,7 @@ struct MCPConnectionsCounter: View {
 
     var body: some View {
         Button {
+            HoverTooltipCoordinator.dismissAll()
             NotificationCenter.default.post(
                 name: .showMCPStatusWindow,
                 object: nil,
--- a/Sources/RepoPrompt/Infrastructure/UI/Components/MCPServerToggleView.swift
+++ b/Sources/RepoPrompt/Infrastructure/UI/Components/MCPServerToggleView.swift
@@ -95,7 +95,10 @@ struct MCPServerToggleView: View {
     }
 
     var body: some View {
-        Button(action: { showPopover.toggle() }) {
+        Button(action: {
+            HoverTooltipCoordinator.dismissAll()
+            showPopover.toggle()
+        }) {
             HStack(spacing: 6) {
                 Image(systemName: "server.rack")
                     .imageScale(.medium)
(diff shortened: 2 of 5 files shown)


rodgomesc:
@0xNino Why did you closed the PR? The issue is still happening. I almost went crazy over the last hour, debugging my Mac processes  to figure out what the heck this pop-up was. My immediate thought was that someone was watching my machine through a backdoor 🫠

image

baron:
@rodgomesc are you on v1.0.29 (the current)? If you can confirm we'd be happy to re-open this PR or fix it separately. Sorry it's happening to you.

0xNino:
I closed this PR because I was not satisfied with the breadth of the original approach and wanted to revisit the fix against current `main`.

The bug is now tracked in #563 and reproduced on RepoPrompt CE v1.0.29 (30). A narrower replacement is open as #564.

The replacement changes only the shared `TooltipOverlayController`: standalone tooltips now dismiss when their owner closes, moves, or resigns key-window status. It does not add the global dismiss coordinator, MCP-specific call sites, or invalid-anchor changes proposed here. The patched debug build has been manually validated against the reported MCP popover flow.

Please continue review and discussion in #563 / #564.

PR #167, opened 2026-06-11
Add local Developer ID production install path

fitchmultz:
## Summary
- add a local-only Developer ID production packaging/install mode for contributors with their own Apple Developer ID Application identity
- require personal display/bundle identity and derive an isolated URL scheme so local installs do not collide with public/debug app identity
- add local Developer ID runtime signing validation and a Keychain namespace derived from team ID + bundle ID

## Validation
- Subagent reviewer loop completed with no blockers or simplification findings worth fixing before commit
- `bash -n Scripts/package_app.sh Scripts/install_local_production.sh Scripts/validate_staged_release.sh`
- `python3 Scripts/test_local_production_installer.py`
- `python3 Scripts/test_release_tooling.py`
- `.agents/skills/rpce-contribution-check/scripts/preflight.sh commit`
- `.agents/skills/rpce-contribution-check/scripts/preflight.sh push`

PR #229, opened 2026-06-15
Bound conductor XCTest stalls

fitchmultz:
## Summary
- cap conductor `test` / `provider-test` jobs at 15 minutes by default
- enable the XCTest progress-stall watchdog by default for test operations
- arm the watchdog after SwiftPM reports `Build complete!` so pre-first-test XCTest stalls are diagnosed and killed instead of running until the generic timeout

## Validation
- `make conductor-selftest`
- `./conductor test --filter WorkspaceFileContextStoreTests` now fails fast after the watchdog threshold and captures stall diagnostics instead of hanging indefinitely; observed diagnostic: `currentTest: null`, xctest sampled inside `XCTWaiter` before first progress marker.

## Notes
This keeps the conductor/test-runner fix separate from PR #167.

### DIFF
--- a/Scripts/conductor.py
+++ b/Scripts/conductor.py
@@ -68,8 +68,10 @@ APP_STOP_CONFIRM_TIMEOUT_SECONDS = 8.0
 APP_STOP_DELAYED_LAUNCH_CONFIRM_TIMEOUT_SECONDS = 25.0
 
 SHORT_TIMEOUT_SECONDS = 5 * 60
+TEST_TIMEOUT_SECONDS = 15 * 60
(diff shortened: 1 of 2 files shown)


fitchmultz:
Closing for upstream PR hygiene. The conductor stall guard is useful and remains in the validated pi integration branch, but as a standalone upstream PR it currently fails the required push gate by exposing an existing XCTest stall: StoreBackedWorkspaceSearchTests/testStoreBackedSearchAbsoluteFolderAndFileScopesReportScopedCounts stalled after testSearchScopeParserKeepsRequiredResolutionOrder. This should come back paired with the root-cause test fix rather than remain as a non-green PR.

PR #240, opened 2026-06-17
Add Z.ai GLM 5.2 model support

baron:
## Summary
- add GLM 5.2 / 1M defaults for CC Zai and migrate legacy persisted GLM slot defaults
- expose distinct CC Zai picker labels and legacy direct GLM choices across Agent Mode, settings, and MCP/catalog surfaces
- route direct legacy GLM selections through transient slot overrides while preserving configured slot mappings
- address review follow-ups by validating encoded GLM effort selections in native runtime resolution and migrating untouched legacy GLM slots per slot while preserving custom overrides
- add regression coverage for GLM migration, effort gating, direct legacy routing, picker labels, ordering, and native encoded-model validation

## Validation
- `git diff --check`
- `make dev-format`
- `make dev-lint`
- `make dev-test FILTER=ClaudeCompatiblePluginBridgeTests`
- `make dev-test FILTER=ClaudeNativeApprovalAndResumeTests`
- `make dev-test`
- commit preflight: `.agents/skills/rpce-contribution-check/scripts/preflight.sh commit`
- push preflight: `.agents/skills/rpce-contribution-check/scripts/preflight.sh push`
- GitHub CI on `564201757ec4d951aa5801a77dbde3ef1910ebd7`: Style, Secret Scan, and Build and Test all passed
- debug app packaged and launched from this branch via `./conductor app relaunch`

## Notes
- The launched debug bundle is `/Users/baron/Library/Application Support/RepoPrompt CE/DebugApps/RepoPrompt.app` with bundle id `com.pvncher.repoprompt.ce.debug`.
- Non-disruptive `make dev-smoke` was attempted after launch, but `rpce-cli-debug` still could not connect to the debug app MCP endpoint (`RepoPrompt app is not running or MCP is disabled`, exit 73). No additional visible relaunch was performed for smoke.
- A delegated review pass found two follow-up issues; both are addressed in commit `5642017`.


Sources/RepoPrompt/Infrastructure/AI/Providers/ClaudeCode/SDK/ClaudeNativeProcessSessionController.swift
```diff
@@ -584,7 +584,7 @@ final actor ClaudeNativeProcessSessionController {
     ) async throws -> (launchEnvironment: ClaudeCodeLaunchEnvironment, request: [String: Any]?) {
         let modelSpecifier = model.map { ClaudeModelSpecifier(raw: $0) }
         let requestedModel = modelSpecifier != nil
-            ? modelSpecifier?.runtimeModelParam
+            ? model
```
erauner12, line 612: @baron I think there’s one edge case with direct GLM selections here: `applyModelAndEffort` resolves the compatible backend launch environment, but after the process is already initialized it only sends `apply_flag_settings`. For direct GLM choices, the actual backend model comes from the `ANTHROPIC_DEFAULT_*` env overrides set at process launch, so switching live between a slot/default and `glm-5-turbo` / `glm-5.1` / `glm-4.7` could keep the old env mapping while the UI thinks the model changed.

Could we either relaunch when the resolved backend env changes, or reject/defer live model switches that require env mutation? A regression test for direct GLM ↔ slot switching would be good too.

erauner12 (changes requested):
The Claude slot labels showing as Sonnet/Opus/Haiku are confusing for Z.ai, but I agree this PR is mostly preserving Claude Code’s slot semantics.

Should we open a follow up issue to improve the UI copy/help text around "slot routed to GLM model" rather than block this PR on a larger label refactor?

cc @provencher 

image
image

PR #252, opened 2026-06-18
Fix Claude Safe Managed launch settings and MCP CLI wait bounds

morluto:
## Summary

This PR fixes several safety, lifecycle, timeout, and CLI discovery issues found during an Agent Mode/MCP review.

The main theme is keeping UI promises, runtime behavior, and asynchronous cleanup aligned. In a few places the code showed a safe or bounded state to the user, but the underlying process or background task could still behave differently.

### 1. Claude Safe Managed UI/runtime mismatch

Safe Managed showed Claude as running with Bash disabled and MCP strict mode enabled, but the actual Claude launch path only passed the permission mode. That meant Claude could fall back to user/default tool settings even when the UI/snapshot said Safe Managed was active.

Fix:
- Extend `AgentProviderRuntimePermissionBinding` with Claude Bash and MCP strict launch flags.
- Derive Claude UI tool settings and runtime launch settings from the same helper.
- Pass those values into `ClaudeCodeAgentConfig.agentMode`.
- Track the launch flags on the active Claude controller so changed settings recycle the controller on the next idle run.
- Add UI copy making active Claude tool-setting changes explicitly next-run-only.

### 2. MCP service lifecycle wait could hang behind watcher tasks

`MCPService.run()` said it waited for the first lifecycle task to finish, but the implementation drained the whole task group. A clean transport exit could therefore wait behind kill-signal or parent-process watcher tasks.

Fix:
- Wait for the first completed lifecycle task.
- Cancel the remaining watcher tasks immediately after that first completion.

### 3. CLI timeout response could wait on cancel notification delivery

When a tool call timed out, the CLI waited for cancel-notification delivery before returning the timeout. If that write stalled, the user-visible timeout could overrun.

Fix:
- Timeout outcomes return immediately after settlement.
- Caller-cancelled outcomes still wait for cancellation delivery, preserving the explicit caller-cancellation behavior.

### 4. Bootstrap handshake write had no deadline

The initial bootstrap request write used a blocking `Darwin.write` loop. If the peer accepted the connection but stopped reading, the bounded read timeout did not protect the write.

Fix:
- Use the existing non-blocking FD writer for the handshake request.
- Add `MCPBootstrapTiming.initialRequestWriteTimeout`.
- Map writer failures back into the existing interactive-session error types.

### 5. External `agent_run` timeouts accepted unsafe values

`agent_run` accepted any finite non-negative timeout, then later converted seconds into nanoseconds. Oversized values could overflow or trap during duration conversion.

Fix:
- Cap external MCP `agent_run` timeouts at `86400` seconds.
- Reject larger int/double/string timeout values during parsing.
- Use a shared clamped nanosecond conversion for `Task.sleep`.

### 6. CLI tool discovery repeated unnecessary `tools/list` round trips

Several discovery commands forced a fresh tool list even when the catalog had already been fetched and the server had not marked it dirty.

Fix:
- Add `cachedToolsOrRefresh()`.
- Use it for repeated `tools`, schema, describe, and snapshot paths.
- Split `toolsDirty` from `toolsChangeNoticePending` so acknowledging a user notice does not incorrectly mark a stale catalog as fresh.

### 7. Existing upstream test asserted asynchronous cleanup synchronously

`WorkspaceFileContextStoreTests.testSessionWorktreeOwnershipReleaseDuringRootLoadUnloadsLateRoot` failed on clean `origin/main`. The product code schedules late root cleanup through a background load-flight cleanup task; the test asserted the root was gone immediately after the stale preparation completed.

Fix:
- Keep the same contract, but wait with the existing bounded async-condition helper before asserting the late root has unloaded.

## Review Guide

Suggested review order:

1. `AgentProviderPreferenceSnapshotStore.swift`, `AgentProviderBindingModels.swift`, and `ClaudeAgentModeCoordinator.swift`
   - Confirm Safe Managed now derives UI and launch flags from one runtime binding.
   - Confirm active Claude controllers recycle only when idle.

2. `InteractiveMCPClientSession.swift`
   - Review timeout settlement behavior.
   - Review non-blocking bootstrap write mapping.
   - Review tool-catalog cache state: `toolsDirty` vs `toolsChangeNoticePending`.

3. `MCPService.run()` in `Sources/RepoPromptMCP/main.swift`
   - Confirm task-group behavior now matches the existing comment.

4. `AgentMCPToolHelpers.swift` and `AgentRunSessionStore.swift`
   - Confirm timeout bounds are applied before duration conversion.

5. Tests
   - Safe Managed runtime binding regression test.
   - CLI tool-catalog cache regression test.
   - Oversized `agent_run` timeout regression test.
   - Workspace late-cleanup test stabilization.

## Validation

Passed:

- Commit preflight: whitespace, staged secret scan, guardrails
- Push preflight: outgoing secret scan, `make dev-lint`, full `make dev-test`
- Focused tests during implementation:
  - `AgentPermissionSecureStoreTests`
  - `AgentLifecycleExecutionContractTests`
  - `InteractiveMCPClientSessionCancellationTests`
  - `AgentModeRunServiceLifecycleTests`
  - `ClaudeCompatiblePluginBridgeTests`
  - `InteractiveMCPClientSessionConnectTimeoutTests`

## Branch

`codex/agent-mcp-safety-hardening`


provencher:
fyi just a note. there's a test contribution policy that requires updating a ledger. Ensure your model updates it in the pr

morluto:
@provencher thanks for letting me know, updated it in the PR

Sources/RepoPrompt/Features/AgentMode/Views/AgentInputBar.swift
```diff
@@ -1321,6 +1325,19 @@ struct AgentComposerView: View, Equatable {
     private var claudeToolsPopoverContent: some View {
         if let claudeTools = props.providerControls?.claudeTools {
             Form {
+                if isSelectedClaudeRunActive {
```
provencher: see if you can avoid branching in a swiftui view like this viewbuilder is ususally a way around it

morluto:
refactored it to use viewbuilder instead of branching

provencher (dismissed):
Found two issues that should be addressed before merging:

1. **High — `ClaudeAgentModeCoordinator.swift:219-227, 550-633`**  
   When the effective permission mode tightens during an active turn, `ensureClaudeNativeSession` returns without recycling the mismatched controller. The subsequent message path then interrupts and sends through that retained controller, so the new turn can still use the process launched with the older, less-restrictive Bash/MCP permissions. Please block the send until the old turn is interrupted and the controller is recreated with the new effective mode.

2. **Medium — `InteractiveMCPClientSession.swift:403-407, 424-425`**  
   `refreshTools()` suspends in `client.listTools()` and then unconditionally clears `toolsDirty`. A tool-list notification or overlapping refresh during that await can therefore be overwritten by an older response, losing the invalidation and leaving the cached catalog stale. Please track an invalidation/request generation and only publish/clear dirty state when the response is still current.

morluto:
addressed the issues + other fixes

morluto:
@provencher the requested changes should be addressed in the current head:

- `98fd354` blocks sends through a stale Claude native controller when effective launch permissions change mid-turn; it interrupts, recycles, and recreates the controller before sending.
- `b003d40` adds generation/epoch checks around MCP `refreshTools()` so an older `listTools()` response cannot clear a newer invalidation or publish stale catalog state.
- `33008ae` updates the test contract ledger for the added coverage.

CI is green on the current head (`f2930da`). Could you re-review when you have a chance?

provencher (approved):
Reviewed after safety repairs, independent cleanup, architecture containment pass, full local validation, and exact-head hosted CI.

Issue #267, opened 2026-06-20
Update menu shows stale disabled “Check for Updates…” item and duplicate Help entry


BetaFold3:
## Summary

RepoPrompt CE shows two “Check for Updates…” menu items:

- One in the application menu after “About RepoPrompt CE”
- One in the Help menu

The application-menu item can appear disabled/stale, while the Help-menu item remains clickable.

## Impact

This makes the update UI confusing. macOS/Sparkle convention is to expose a single “Check for Updates…” command in the application menu, not a duplicate in Help.

## Reproduction

1. Build and run RepoPrompt CE.
2. Open the application menu.
3. Observe “Check for Updates…” after “About RepoPrompt CE” is disabled.
4. Open the Help menu.
5. Observe another “Check for Updates…” item that is clickable.

## Expected behavior

- Only one update command appears, in the application menu after About.
- The command updates when Sparkle state changes.
- If an update is available, the menu should reflect that state instead of staying on a stale disabled “Check for Updates…” item.

## Actual behavior

- `UpdateMenu` adds the app-menu item via `CommandGroup(after: .appInfo)`.
- `HelpMenu` adds a second independent “Check for Updates…” button.
- Runtime diagnostics showed Sparkle was ready and an update was available:
  - `can_check_for_updates=true`
  - `sparkle_can_check_for_updates=true`
  - `sparkle_configuration_valid=true`
  - `updater_started=true`
  - `update_available=true`
  - `update_version=1.0.20`

Despite that, the app menu still showed a disabled “Check for Updates…” item.

## Likely cause

There appear to be two related issues:

1. Duplicate menu command:
   - `Sources/RepoPrompt/App/RepoPromptApp.swift` installs both `UpdateMenu()` and `HelpMenu()`.
   - `Sources/RepoPrompt/App/HelpMenu.swift` declares a duplicate `Button("Check for Updates…")`.

2. Stale app-menu state:
   - `Sources/RepoPrompt/App/Sparkle/UpdateMenu.swift` reads `SparkleUpdaterManager.shared` through an unobserved computed property.
   - It does not use `@ObservedObject`, `@StateObject`, `@EnvironmentObject`, or injected observed state.
   - As a result, the command may render before Sparkle publishes readiness/update availability and then fail to refresh.

## Suggested fix

- Keep the native update command in `UpdateMenu` after `.appInfo`.
- Make `UpdateMenu` observe `SparkleUpdaterManager` so menu state refreshes.
[...]

PR #268, opened 2026-06-20
Fix duplicate Check for Updates menu item

BetaFold3:
## Summary

Fixes the duplicated “Check for Updates…” menu command and aligns RepoPrompt CE with standard macOS menu conventions.

On macOS, application update commands conventionally live in the app menu near “About <App>”, not in Help. RepoPrompt CE already had an app-menu update command after “About RepoPrompt CE”, but it also added a second “Check for Updates…” item under Help.

This PR keeps the native app-menu location as the single update entry point and removes the Help-menu duplicate.

It also fixes stale update-menu state by passing the app-owned `SparkleUpdaterManager` into `UpdateMenu` as an observed object instead of reading the singleton through an unobserved computed property. This lets the app-menu command refresh when Sparkle publishes readiness or update-availability changes.

## Changes

- Keep the update command in the native macOS app menu after “About RepoPrompt CE”
- Remove the duplicate “Check for Updates…” command from Help
- Inject `appDelegate.sparkleManager` into `UpdateMenu`
- Observe `SparkleUpdaterManager` from `UpdateMenu` so published Sparkle state updates the command UI

## Validation

- `make dev-format`
- `git diff --check`
- `make guardrails`
- `make dev-lint`

### DIFF
--- a/Sources/RepoPrompt/App/HelpMenu.swift
+++ b/Sources/RepoPrompt/App/HelpMenu.swift
@@ -1,19 +1,8 @@
-import Sparkle
 import SwiftUI
 
 struct HelpMenu: View {
-    private var sparkleManager: SparkleUpdaterManager {
-        SparkleUpdaterManager.shared
-    }
-
     var body: some View {
         Group {
-            Button("Check for Updates…") {
-                sparkleManager.checkForUpdates()
-            }
-
-            Divider()
-
             Link(
                 "What's New",
                 destination: URL(string: "https://repoprompt.com/docs#s=changelog")!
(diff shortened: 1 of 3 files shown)


BetaFold3:
Before: 

rpce_app_menu_before
rpce_help_menu_before

BetaFold3:
After: 
rcpe_app_menu_after
rcpe_help_menu_after

provencher (approved):

BetaFold3:
Fixed https://github.com/repoprompt/repoprompt-ce/issues/267

PR #274, opened 2026-06-21
keep @file tag paths pathful and non-absolute

### LINKED ISSUE #266: @-mention file tags use basename only, causing same-basename ambiguity
## Impact / scope
In Agent Mode, picking a file via the `@` file-tag picker commits a **basename-only** reference into the composer text and the attachment chip (e.g. `@SKILL.md`), even though the selected file's full relative path is preserved internally. When two selected/contextualized files share a basename, the model receives ambiguous raw text (`@SKILL.md`) alongside correctly path-qualified appended `File: <relative/path>/SKILL.md` blocks in the same message, and can anchor on the wrong file.

The picked file *identity* is not lost — this is a display/serialization bug in the composer file-tag flow, not in path resolution or provider packaging.

## Expected vs actual
- **Expected:** The committed inline `@` reference and the attachment chip use the file's logical/workspace-relative path (e.g. `@docs/area-a/SKILL.md`), matching the path-qualified `File:` block sent to the provider.
- **Actual:** Both the inline `@` token and the chip collapse to the basename (`@SKILL.md`) for non-duplicate indexed candidates and for the bare-`@` selected-file fallback. The raw ambiguous text and the correct file block then coexist in one provider message.

Composer after picking a file via @: both the attachment chip and the inline token show basename-only 'SKILL.md' / '@SKILL.md', dropping the path

## Reproduction
1. In a workspace containing two files with the same basename in different folders (e.g. `area-a/SKILL.md` and `area-b/SKILL.md`).
2. In the Agent Mode composer, type `@` and pick one of them (e.g. `area-a/SKILL.md`).
3. Observe the inserted reference and the attachment chip show `@SKILL.md` / `SKILL.md` (basename only).
4. Send a message referring to "this file". 
5. Result: the model has `@SKILL.md` in the raw text plus more than one `SKILL.md` in context, and may reason about / answer for the wrong file.

Frequency: deterministic for same-basename selections.

## Root cause (confirmed via code reading)
Basename-only serialization in the composer file-tag flow:
- `AgentFileTagSuggestionService.commitDisplayText(fileName:tokenRelativePath:isDuplicateName:)` returns the basename for non-duplicate indexed candidates; only duplicate basenames return the relative path — `Sources/RepoPrompt/Infrastructure/UI/Mentions/AgentFileTagSuggestionService.swift` (~L367-378).
- The bare-`@` selected-file fallback hardcodes `commitDisplayText: file.name` even though `relativePath` holds the full logical path — same file (~L305-335).
- `FileTagMentionHelper.committedReplacementText(...)` prefers `commitDisplayText`, so the composer text becomes `@SKILL.md` — `Sources/RepoPrompt/Infrastructure/UI/TextField/TextFieldMentionHelpers.swift` (~L340-348).
- `AgentFileMentionText.attachmentDisplayName(for:)` also prefers `commitDisplayText`, and `addPendingTaggedFile` defensively falls back to `lastPathComponent`; `AgentAttachmentsStrip` renders that as the chip — `Sources/RepoPrompt/Features/AgentMode/Views/AgentInputBar.swift` (~L185-202), `Sources/RepoPrompt/Features/AgentMode/ViewModels/AgentModeViewModel+DraftsAttachments.swift` (~L199-207), `Sources/RepoPrompt/Features/AgentMode/Views/AgentAttachmentsStrip.swift` (~L213-237).
- Provider packaging is correct: `AgentTaggedFileAttachment.relativePath` is preserved and `File:` blocks use relative/logical paths. The defect is that the final provider payload is a single text string where ambiguous raw text and the correct file block coexist.

The correct source value already exists as `tokenRelativePath` (logical path for worktree-bound sessions; `ClientPathFormatter.displayPath` for multi-root) rather than `file.name`.

## Suggested fix direction (non-binding)
- Prefer `tokenRelativePath` (logical/workspace-relative) for committed Agent file-tag text and chip labels instead of basename — primary fix.
- Change the selected-file fallback from `commitDisplayText: file.name` to the relative path.
- Make `attachmentDisplayName(for:)` and the `addPendingTaggedFile` default consistent with the inserted `@path`.
- Provider-side `@SKILL.md → relative/path` rewriting is reasonable defense-in-depth for manually typed/legacy basename mentions, but should not be the primary fix.

## Caveat for "no absolute paths" guarantees
`ClientPathFormatter.displayPath` intentionally falls back to an **absolute** path when multiple visible roots share the same root name — `Sources/RepoPrompt/Infrastructure/WorkspaceContext/PathResolution/WorkspacePathPolicy.swift` (~L160-166). A UI-safe root alias may be needed before promising universal no-absolute-path behavior.

## Acceptance criteria
- Selecting a file via `@` inserts a path-qualified reference (logical/workspace-relative) for both unique and duplicate basenames.
- The attachment chip label matches the inserted `@path` reference (not basename-only).
- For same-basename selections, the provider message contains a path-qualified `@.../SKILL.md` that matches the appended `File: .../SKILL.md` block.
- No physical worktree path leakage for worktree-bound sessions; logical paths preserved.
- Regression coverage added: unique indexed file tags, selected-file fallback, duplicate-basename behavior, chip display text, worktree logical paths, and a duplicate-root-name case documenting the absolute-path fallback.

## Environment
- RepoPrompt CE, Agent Mode composer. Confirmed by source reading on the current working tree (2026-06-20). Not provider-specific.

aAtila:
Agent Mode `@file` mentions now keep resolvable, non-absolute logical paths, so composer text, chips, and provider context stay aligned.

## Changes

- Picked file mentions commit their logical relative path instead of collapsing to an ambiguous basename like `@index.ts`.
- Pending tagged-file rendering stays pathful with middle truncation, matching the committed text, visible chips, and provider context.
- Added non-absolute root aliases for same-named workspace roots, and resolve generated file-tag aliases before shadowing root aliases, so pathful tags no longer fall back to absolute paths.
- Regression coverage for pathful commit text, chip labels, removal, duplicate roots, and alias shadowing.

## Why

`@file` mentions were shrinking down to just the filename (like `@index.ts`), so when you had multiple files sharing a name you couldn't tell which one was actually attached. It confused the model too, in some cases it would read multiple `index.ts` in the workspace instead of just the selected one. Separately, when two workspace roots shared a name, a pathful tag could resolve to a machine-local absolute path and leak local filesystem layout into provider context. Anchoring on resolvable, non-absolute logical paths fixes all of these in one place.

Fixes [#266](https://github.com/repoprompt/repoprompt-ce/issues/266)


baron (approved):
Approved. I reviewed the PR delta against the merge-base with `origin/main`, focusing on the `@file` tag path handling changes and performance risk.

Assessment:
- No blocking correctness issues found.
- The pathful, non-absolute token behavior is covered for same-named visible roots, selected-file fallback, generated-alias resolution, attachment display/removal, and provider XML display.
- No new filesystem scans or obvious unbounded runtime work were introduced.
- The only performance note I found is non-blocking: duplicate-root alias formatting recomputes suffix/path-component matching per suggestion candidate. That work is bounded by existing suggestion candidate limits and tagged-file caps, so I do not think it should block this PR. It could be optimized later by precomputing root aliases per visible-root set.

Validation run:
- `make dev-test FILTER=AgentFileTagWorktreeResolutionTests`
- Passed: 14 tests, 0 failures.

PR #275, opened 2026-06-21
Add scoped Agent Models settings

crstdr:
## Summary
- add global/workspace-scoped Agent Models settings with workspace override selection
- resolve Agent Models choices through the active workspace scope for UI, Context Builder, Oracle, and MCP-launched agents
- polish the workspace settings UI and inline the copy-settings action into the scope card
- document settings persistence and stabilize the Oracle pill routing fixture

## Validation
- .agents/skills/rpce-contribution-check/scripts/preflight.sh push
  - guardrails
  - outgoing secret scan
  - make dev-lint
  - make dev-test
- focused checks during validation:
  - make dev-test FILTER=AgentOraclePillRoutingTests
  - make dev-test FILTER=ApplyEditsCoreRecoveryTests
  - make dev-test FILTER=PersistentAgentModeMCPReadFileConnectionTests/testWorktreeReadCoverageCertificateMintsOnlyAfterVerifiedPersistenceAndResponseStaysAsync


baron (commented):
Thanks for the PR — I did a deeper local review plus QA pass against a prepared worktree.

Overall, the scoped Agent Models settings work is promising and most exercised flows behaved correctly, but we found one state/UI issue that I think should be addressed before merge. I’m leaving this as review feedback rather than force-pushing our local repair onto the PR branch.

## Must fix: explicit role override matching recommendation is persisted but not clearable

In `Settings -> Agent Models -> Sub-Agent Role Defaults`, if a workspace override is set to the same model as the current recommendation, the override is persisted as an explicit workspace override, but the UI currently treats it as indistinguishable from recommendation-tracking state.

QA repro:

1. Switch to `Use workspace overrides`.
2. In `Sub-Agent Role Defaults`, select a role/model that already matches the recommendation, e.g. `Design -> Cursor CLI Composer 2`.
3. The selection is persisted in `globalSettings.json` under the workspace Agent Models profile, e.g. `profile.mcpAgentRoleOverrides.design = "cursor:composer-2"`.
4. But the UI still looks fully “recommended” and does not expose a per-row clear/reset affordance unless some other role also visibly differs from recommendation.

Why this matters: this creates a hidden pinned state. The workspace looks like it is tracking recommendations, but it is actually locked to a specific model and may not follow future recommendation changes.

## Suggested implementation direction

I tested a local proof-of-fix by separating two concepts in the role-default resolution model:

- `hasStoredOverride`: any explicit stored override exists, even if it equals the current recommendation.
- `hasCustomOverride`: the effective selection differs from the current recommendation.

Then the UI can use:

- `hasStoredOverride` to decide whether reset/clear affordances should appear.
- `hasCustomOverride` to decide whether the role counts as “differs from recommended.”

In the local proof-of-fix, same-as-recommended explicit pins show:

- `Pinned to recommended`
- `Clear Pin`
- section-level `Reset All to Recommended`

while visibly different overrides still show:

- `Recommended: ...`
- `Apply`
- `Role defaults: 1 of 4 differ`

## QA synthesis

Initial QA found:

- Basic launch/workspace load passed.
- Agent Models settings scope behavior passed.
- Built-in Chat / Oracle sync behavior passed.
- The same-as-recommended role override clearability issue above failed.
- Multi-window recommendation refresh and legacy migration sanity were blocked/not fully exercised.

After applying the local proof-of-fix and rebuilding the debug app, QA retested and confirmed:

- Basic launch/workspace load passed.
- CLI smoke passed: `windows`, `tree --type roots`, `context --tree`, and `agent_manage list_agents roles_only=true`.
- Same-as-recommended `Design -> Cursor CLI Composer 2` now persisted explicitly and showed `Pinned to recommended`, `Clear Pin`, and `Reset All to Recommended`.
- `Clear Pin` removed the stored override while leaving the visible recommended model selected.
- Different override path also behaved correctly: `Explore -> Codex CLI GPT-5.5 High` showed `Recommended: Codex CLI GPT-5.5 Low`, `Apply`, and `Role defaults: 1 of 4 differ`; clicking `Apply` cleared the override.
- Scope behavior and Copy to Workspace / Copy to Global still behaved correctly.
- Oracle/chat sync retested cleanly, including the picker returning with the synced Oracle value rather than stale previous chat value.
- No failures reproduced in the retest.

One note: the formatted `agent_manage list_agents roles_only=true` output in QA did not print the new/custom override flags directly, but persistence was verified in `globalSettings.json`.

## Suggested tests

I’d recommend adding regression coverage for:

1. A same-as-recommended explicit role selection remains persisted.
2. Resolution reports “stored override exists” separately from “differs from recommendation.”
3. The settings view model exposes reset affordance state for same-as-recommended pins.
4. Clearing that pin removes the workspace override while preserving recommendation-tracking behavior.

I have a local prepared worktree with a passing proof-of-fix if useful, but I’d prefer to let you incorporate/adapt the fix rather than overwrite the PR branch.

baron (approved):
Final maintainer review complete after rebase onto current main. Prior stored-pin must-fix is resolved; exact-head CI is green. Remaining wizard-scope and compact-popover pin visibility items will be tracked separately.

PR #289, opened 2026-06-23
Fix symlinked global skill root readability

### LINKED ISSUE #288: Skill reference files under symlinked global skill roots should remain RepoPrompt-readable
**Problem**

When the global skills root is a symlink, for example:

`<home>/.agents/skills -> <external-skills-checkout>`

a triggered skill may provide reference paths like:

`<home>/.agents/skills/<skill-name>/references/verifier.md`

RepoPrompt discovers the skill through the lexical ~/.agents/skills path, but native read resolution can canonicalize the file to the symlink target:

`<external-skills-checkout>/<skill-name>/references/verifier.md`

and then reject it because the always-readable support allowlist still contains only the lexical ~/.agents/skills root.

As a result, the agent may fall back to shell commands such as:

`cat references/verifier.md`

instead of using RepoPrompt’s native read path. This wastes an agent turn, adds avoidable workaround chatter, and pollutes the limited agent context window with shell transcripts that should not be necessary for a user-triggered skill’s own reference files.

**Expected Behavior**

If a file is under an allowed global skill root, it should remain RepoPrompt-readable even when that global root is a symlink.

This should apply to both:

	•	the main agent session, and
	•	spawned worker sessions.

**Actual Behavior**

Skill reference reads under a symlinked ~/.agents/skills root may fail or be avoided, causing shell fallback.

**Suspected Cause**

Support-root containment appears asymmetric across symlink resolution:

	•	the requested path starts under lexical ~/.agents/skills
	•	RepoPrompt canonicalizes it to the symlink target
	•	the second allowlist check compares against only the lexical support root
	•	the canonical target is rejected

Relevant areas:

	•	Sources/RepoPrompt/Infrastructure/WorkspaceContext/PathResolution/AgentSupportDirectoryCatalog.swift
	•	Sources/RepoPrompt/Infrastructure/WorkspaceContext/WorkspaceReadableFileService.swift
	•	Sources/RepoPrompt/Features/WorkspaceFiles/ViewModels/WorkspaceFilesViewModel.swift

**Acceptance Criteria**

	•	Files under a symlinked ~/.agents/skills root are readable through RepoPrompt native reads.
	•	The nominal lexical path and canonical realpath both resolve correctly.
	•	Main agent sessions and spawned workers do not need shell fallback for these files.
	•	Nested symlink escapes outside the resolved support root remain blocked.
	•	Regression coverage exists for symlinked global skill roots.

BetaFold3:
**Summary**

This fixes native RepoPrompt reads for files under symlinked global skill/support roots, such as:

`~/.agents/skills -> /path/to/external-skills-checkout`

Previously, a skill reference file could pass the initial allowlist check using the lexical ~/.agents/skills/... path, then be canonicalized to the symlink target and rejected by the second allowlist check. That could force agents or spawned workers to fall back to shell reads for files that should be RepoPrompt-readable.

**Changes**

	•	Makes always-readable support-root containment symmetric across symlink resolution.
	•	Allows both the nominal lexical path and canonical realpath under a symlinked support root to resolve correctly.
	•	Preserves nested symlink escape protection so paths escaping outside the allowed support root remain blocked.
	•	Adds focused regression coverage for:
	•	WorkspaceReadableFileService
	•	MCP read_file external fallback behavior

**Validation**

	•	make dev-test FILTER=WorkspaceFileContextStoreTests
	•	make dev-test FILTER=MCPReadFileExactAbsoluteCatalogFastPathTests
	•	make dev-lint
	•	rpce-contribution-check commit preflight

Closes https://github.com/repoprompt/repoprompt-ce/issues/288

### DIFF
--- a/Sources/RepoPrompt/Infrastructure/WorkspaceContext/PathResolution/AgentSupportDirectoryCatalog.swift
+++ b/Sources/RepoPrompt/Infrastructure/WorkspaceContext/PathResolution/AgentSupportDirectoryCatalog.swift
@@ -88,9 +88,13 @@ enum AgentSupportDirectoryCatalog {
     ) -> Bool {
         let normalizedDirectory = normalizedPath(for: directoryPath)
         guard normalizedDirectory.hasPrefix("/") else { return false }
-        for candidate in containmentCandidates(for: absolutePath, fileManager: fileManager) {
-            if candidate == normalizedDirectory || candidate.hasPrefix(normalizedDirectory + "/") {
-                return true
+        let directoryCandidates = containmentCandidates(for: normalizedDirectory, fileManager: fileManager)
+        let pathCandidates = containmentCandidates(for: absolutePath, fileManager: fileManager)
+        for directoryCandidate in directoryCandidates {
+            for pathCandidate in pathCandidates {
+                if pathCandidate == directoryCandidate || pathCandidate.hasPrefix(directoryCandidate + "/") {
+                    return true
+                }
             }
         }
         return false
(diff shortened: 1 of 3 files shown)


BetaFold3:
Before:
rpce_skill_symlinked_before

BetaFold3:
After:

rpce_skill_symlinked_after

provencher (approved):

BetaFold3:
@provencher I can't merge this. Can you do the honor please?

PR #296, opened 2026-06-24
fix(claude-code): emit CLI identity preamble for GLM/z.ai backend

### LINKED ISSUE #295: GLM/z.ai: 529 "overloaded" failures caused by the Agent SDK identity preamble
## Summary

Running Claude Code against the GLM (z.ai) backend from RepoPrompt CE frequently fails with:

```
API Error: 529 [1305][The service may be temporarily overloaded, please try again later][…]
```

This happens mostly during evening peak hours, occasionally during the day. The same models work fine when running `claude` interactively in a terminal.

## Root cause (proven)

This is **not** a timeout or retry-budget problem (both investigated and disproven — see below). z.ai's gateway **selectively rejects requests whose system prompt carries the Claude Agent SDK identity preamble**, returning a misleading `529 overloaded_error` (code `1305`).

The `claude` CLI can emit two different identity preambles (block 1 of the system prompt):

| CLI invocation | Block 1 (identity preamble) | z.ai under load |
| --- | --- | --- |
| interactive, or `-p` without `--verbose` | `You are Claude Code, Anthropic's official CLI for Claude…` | ✅ served |
| `-p --verbose` (required for `--output-format stream-json`) | `You are a Claude agent, built on Anthropic's Claude Agent SDK.` | ❌ rejected (529) |

RepoPrompt CE launches Claude Code with `--verbose --output-format stream-json --input-format stream-json` (required for the interactive streaming protocol), which makes the CLI emit the **Agent SDK** preamble — the one z.ai sheds. Interactive terminal usage emits the **Claude Code** preamble and is never shed. That is the entire interactive-vs-RPCE difference.

## Evidence (A/B capture over a live overload window)

Over a ~67-minute live z.ai overload window (2026-06-24 04:19–05:26 UTC) a probe sent three requests every 150 s and logged the HTTP status. `sdk_full` and `cli_full` were **byte-identical ~127 KB request bodies captured from the real CLI, differing only in block 1** (the identity preamble):

| Variant | Block 1 | 200 | 529 | median latency |
| --- | --- | --- | --- | --- |
| `sdk_full` | `…Claude agent, built on Anthropic's Claude Agent SDK.` | 8 | **16** | 529 → **1.41 s** |
| `cli_full` | `…Claude Code, Anthropic's official CLI…` | **23** | 0 | 200 → 6.67 s |
| `none_min` | trivial baseline | **23** | 0 | 200 → — |

- **16 PROOF cycles** (`sdk=529` and `cli=200` in the same 150 s cycle), **0 CAPACITY cycles** (both `529`). Every time the SDK-preamble request was shed, the byte-identical CLI-preamble request sent seconds later succeeded.
- The 529 returns in **~1.4 s** — a fast, deterministic rejection, not a capacity timeout. `status.hellozai.com` was green throughout and trivial requests kept returning 200.

This is consistent with an anti-impersonation / routing policy that sheds traffic self-identifying as the "Claude Agent SDK" under peak load, mislabeled as `overloaded_error`.

## Why retries / timeout do not fix it

- The 529 is a **deterministic** content/routing rejection (~1.4 s), not a transient capacity blip. Claude's bundled SDK already retries it ~10× with backoff (~3.5 min budget); every retry of the SDK-preamble request failed identically within the window.
- `API_TIMEOUT_MS` governs stream-read patience and has no bearing on a fast 529.
- Exposing a retry setting to pass through to the CLI wouldn't help: Claude Code exposes no env/flag to configure 5xx/529 retry behavior (upstream issues anthropics/claude-code#23115, #35801, #60577 are closed / not-planned; retry logic is hardcoded in the bundled SDK).

## Proposed fix

Pass `--append-system-prompt <non-empty>` for the GLM/z.ai runtime variant so the CLI emits the **Claude Code** identity preamble (the form z.ai never sheds) instead of the Agent SDK one.

- Verified empirically: any **non-empty** value flips block 1 to the `You are Claude Code…` form; an empty string is ignored by the CLI and does nothing, so a real value is required.
- Scoped to the GLM/z.ai variant — no change for real-Anthropic usage. The same mechanism applies to other Claude-compatible backends (Kimi, custom) if they show the same symptom.
- Deliberately **not** using `--fallback-model` (would silently use a model the user did not choose).

Captured request bodies / repro scripts are available on request but not attached here (the captures include auth headers from the test token, which is being rotated).

## Environment

- Claude Code CLI 2.1.187
- Backend: z.ai (`api.z.ai/api/anthropic`), GLM models (`glm-5.1`, `glm-5.2[1m]`, `glm-5-turbo`)
- RepoPrompt CE, Agent Mode, GLM runtime variant

moonray:
Fixes #295.

## Problem

GLM (z.ai) Agent Mode runs frequently fail with `API Error: 529 [1305][The service may be temporarily overloaded…]`, mostly during evening peak hours, while the same models work in an interactive terminal.

## Root cause

This is **not** a timeout or retry-budget issue (both disproven in #295). z.ai's gateway selectively rejects requests whose system prompt carries the Claude Agent SDK identity preamble — `"You are a Claude agent, built on Anthropic's Claude Agent SDK."` — returning a misleading `529 overloaded_error`. RepoPrompt CE launches Claude Code with `--verbose` (required for `--output-format stream-json`), which makes the CLI emit that SDK preamble. Interactive terminal usage emits the never-shed `"You are Claude Code…"` preamble instead.

Proven by A/B capture over a live overload window: byte-identical ~127 KB requests differing only in block 1 were shed with the SDK preamble (16×529, median 1.4 s) and served with the CLI preamble (0×529). Full evidence in #295.

## Fix

Pass `--append-system-prompt` for the `.glm` runtime variant in `ClaudeNativeProcessSessionController.buildArguments`, so the CLI emits the interactive `"You are Claude Code…"` identity preamble (the form z.ai never sheds). Scoped to `.glm` only — real-Anthropic (`.standard`) and the other compatible variants are unchanged.

The value (`"Running within RepoPrompt CE."`) is:
- **non-empty** — an empty/whitespace value is ignored by the CLI and does **not** flip the preamble (verified);
- **free of the trigger words** ("Claude"/"Anthropic"/"agent"/"SDK") so it cannot reintroduce the shedding.

## Testing

- **Unit test** (`ClaudeNativeIdentityPreambleTests`): `.glm` emits the flag with a non-empty, trigger-word-free value; `.standard`/`.kimi`/`.customCompatible` do not. Regression-proven — the `.glm` test fails on the unpatched code (`XCTUnwrap` for the flag) while the negative test passes.
- **Integration check** (capture proxy, the real `claude` CLI + RPCE's exact `buildArguments` output): with the fix, block 1 of the generated system prompt is `"You are Claude Code, Anthropic's official CLI…"`; without it, block 1 is the shed `"You are a Claude agent…"` form. (Run locally; not a committed test since it needs the binary + live credentials.)

### DIFF
--- a/Sources/RepoPrompt/Infrastructure/AI/Providers/ClaudeCode/SDK/ClaudeNativeProcessSessionController.swift
+++ b/Sources/RepoPrompt/Infrastructure/AI/Providers/ClaudeCode/SDK/ClaudeNativeProcessSessionController.swift
@@ -1640,7 +1640,7 @@ final actor ClaudeNativeProcessSessionController {
             Self.buildSetPermissionModeRequest(permissionMode: permissionMode ?? config.permissionMode)
         }
 
-        /// Build CLI arguments for testing (verifies prompt flags are absent).
+        /// Build CLI arguments for testing.
         func test_buildArguments(
             existingSessionID: String?,
             model: String?
@@ -1893,6 +1893,15 @@ final actor ClaudeNativeProcessSessionController {
         return (stream, continuation)
     }
 
+    /// Non-empty system-prompt suffix that makes the CLI emit its interactive `You are Claude Code…`
+    /// identity preamble instead of the Agent SDK `You are a Claude agent…` form. z.ai (GLM)
+    /// selectively rejects the Agent SDK self-identification under peak load with a misleading 529
+    /// `overloaded` that exhausts the CLI's retry budget and fails the run; the interactive form is
+    /// never shed. Empty/whitespace-only values are ignored by the CLI, so this must be a real token,
+    /// and it deliberately avoids the trigger words ("Claude"/"Anthropic"/"agent"/"SDK") so it cannot
+    /// reintroduce the shedding. See https://github.com/repoprompt/repoprompt-ce/issues/295.
+    private static let glmZAIAppendSystemPrompt = "Running within RepoPrompt CE."
+
     private func buildArguments(
         existingSessionID: String?,
         model: String?
@@ -1905,6 +1914,12 @@ final actor ClaudeNativeProcessSessionController {
         ]
 
         args.append(contentsOf: ["--permission-prompt-tool", "stdio"])
+
+        // GLM/z.ai sheds requests carrying the Agent SDK identity preamble under load (see #295).
+        // Any non-empty --append-system-prompt flips the CLI onto the never-shed "Claude Code" preamble.
+        if config.runtimeVariant == .glm {
+            args.append(contentsOf: ["--append-system-prompt", Self.glmZAIAppendSystemPrompt])
+        }
 
         if let existingSessionID, !existingSessionID.isEmpty {
             args.append(contentsOf: ["--resume", existingSessionID])
--- a/Sources/RepoPrompt/Infrastructure/AI/Providers/ClaudeCode/SDK/ClaudeNativeProcessSessionController.swift
+++ b/Sources/RepoPrompt/Infrastructure/AI/Providers/ClaudeCode/SDK/ClaudeNativeProcessSessionController.swift
@@ -1640,7 +1640,7 @@ final actor ClaudeNativeProcessSessionController {
             Self.buildSetPermissionModeRequest(permissionMode: permissionMode ?? config.permissionMode)
         }
 
-        /// Build CLI arguments for testing.
+        /// Build CLI arguments for testing (verifies prompt flags are absent).
         func test_buildArguments(
             existingSessionID: String?,
             model: String?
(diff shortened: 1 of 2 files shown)


Sources/RepoPrompt/Infrastructure/AI/Providers/ClaudeCode/SDK/ClaudeNativeProcessSessionController.swift
```diff
@@ -1640,7 +1640,7 @@ final actor ClaudeNativeProcessSessionController {
             Self.buildSetPermissionModeRequest(permissionMode: permissionMode ?? config.permissionMode)
         }
 
-        /// Build CLI arguments for testing (verifies prompt flags are absent).
+        /// Build CLI arguments for testing.
```
moonray: This change was excessive. Should undo.

provencher (approved):

baron (approved):
Re-reviewed latest head (033ce3e0). Diff remains scoped and checks are green; approving.

PR #305, opened 2026-06-26
Fix Oracle and sub-agent role model selections resetting on restart

### LINKED ISSUE #304: Oracle model is silently cleared by sync/fallback side-effects and persists empty across restart
## Summary

When the **"sync Built-in Chat model with Oracle"** toggle is on (or is inferred-on because the Oracle and Chat models are the same), automatic side-effects can write a blank value into the **global** Oracle model (`scalarPreferences.modelSelection.planningModel` in `GlobalSettingsStore`). Because the Oracle planning model is deliberately never auto-healed, the blank is eagerly persisted and the Oracle shows up **empty after the next launch**. The same class of transient-state → persistent-loss bug also threatens the global MCP sub-agent role defaults.

Reproduced against `main` at `55881b8b5c6531d5a753232cba69a6be5b407220`.

## Impact

Users lose their configured Oracle model (and report sub-agents "resetting") across restart, with no user action that intentionally cleared it. It is intermittent and hard to reason about because it only triggers when an automatic fallback/reconcile fires while the sync toggle is on.

## Observed behavior

`scalarPreferences.modelSelection.planningModel` is GLOBAL (one document, all workspaces), and quit only persists the window session — never model selection — so quitting alone is fine. The corrupting write is an eager `save()` from a setter firing with a blank value while the app runs. Three concrete side-effect paths can do it:

1. **Sync mirror of a blank chat model (primary).** `PromptViewModel.pickDiffCapableFallback()` calls `setPreferredModelRaw("")` when `availableModels` is empty/unhydrated (`PromptViewModel.swift:5181-5185`). With sync on, `GlobalSettingsStore.setPreferredComposeModelRaw` mirrors that empty value into `planningModel` (`GlobalSettingsManager.swift:655-660`) and `save()`s it. `validatePreferredModel()` then deliberately does **not** re-heal the Oracle (`PromptViewModel.swift:5145-5148`), so the blank sticks.

2. **Provider-reset writes a nil replacement.** `APISettingsViewModel.resetPreferredModelIfNeeded` computes `replacement = availableModels.first(where: { !condition($0) })?.rawValue`, which is `nil` when the removed provider was the only one, then (under sync) passes it straight to `setPlanningModelRaw(replacement)` (`APISettingsViewModel.swift:2050-2056`) — directly persisting `nil` into the Oracle, bypassing any compose-mirror guard.

3. **MCP `app_settings` sync-sibling mirror of a null.** `AppSettingsMCPService.postModelRawDidWrite` mirrors a `null`/blank `models.preferred_compose_model` write into `models.planning_model` when sync is on (`AppSettingsMCPService.swift:1309-1318`).

### Expected behavior

- A blank/empty/whitespace chat-model value must **never** be mirrored or written into the Oracle `planningModel` as a sync/fallback/reconcile side-effect.
- Clearing the Oracle stays an **explicit** action (UI picker / `app_settings models.planning_model = null`).
- Real (non-empty) model selections continue to mirror normally under sync.

## Related risk (sub-agents)

`MCPAgentRoleDefaultsService.setSelection` removes a stored sub-agent role override when the selection equals the currently-recommended one (`MCPAgentRoleDefaultsService.swift:120-127`), and "recommended" is derived from *transient* availability. If ever invoked during degraded availability (startup/provider probing), it could permanently drop a user's override — the same transient-state → persistent-loss shape. The non-mutating `resolve()` path already handles degraded availability correctly without clearing.

## Proposed fix

Enforce one invariant — *only an explicit `planning_model` write may clear the Oracle; sync/fallback/reconcile side-effects must never blank it* — by guarding the three side-effect paths above (treat `nil`/empty/whitespace as "not a real model" and skip the Oracle write). Consider making `setSelection` clear only via the explicit reset APIs rather than recommendation-equality during transient availability.

datawisebets:
## Summary

Fixes #304.

Two related "model selection resets after restart" bugs in Agent Models, both an instance of the same class — *a transient/automatic write silently destroys a persisted user model selection*:

1. **Oracle model** — blanked to empty by sync side-effects.
2. **Sub-agent role defaults** (explore/engineer/pair/design) — an explicit pick that matched the current recommendation was never stored, so the role drifted after restart.

Context Builder was investigated as part of this and found **not** to have a reachable durable-reset bug (details below); it is intentionally not changed.

### Oracle model — never let a side-effect blank the Oracle (`planningModel`)
The Oracle planning model is deliberately never auto-healed, so any blank that reaches it persists. Invariant: *only an explicit `planning_model` write may clear the Oracle.*
- `GlobalSettingsStore.setPreferredComposeModelRaw` — don't mirror a blank/whitespace chat model into `planningModel` under sync (the fallback `setPreferredModelRaw("")` in `PromptViewModel.pickDiffCapableFallback()` was the source).
- `APISettingsViewModel.resetPreferredModelIfNeeded` — don't write a `nil` replacement into `planningModel` when no real model is available.
- `AppSettingsMCPService.postModelRawDidWrite` — don't mirror a `null`/blank `app_settings` compose write into `planning_model`.
- All guards treat `nil` / empty / whitespace as "not a real model".

### Sub-agent role defaults — always persist an explicit pick
- `MCPAgentRoleDefaultsService.setSelection` previously **removed** the override when the selection equalled the current recommendation. The recommendation is transient (availability / provider-filter dependent), so nothing was stored and the role resolved to a possibly-different recommendation after restart. Now an explicit selection is always persisted; reverting to recommendation-tracking stays explicit via `clearOverride` / `clearAllOverrides`.

### Tests + ledger
- New/updated regression tests for the Oracle mirror (incl. whitespace, across a simulated relaunch) and for role persistence; a guard test that `clearOverride` still reverts a role. Contract-ledger reconciled (rename + adds).

## Validation

- **Oracle:** `swift test --filter AIModelPreferenceRegressionTests` — RED before the guard (Oracle → `""` in memory and after reload), GREEN after.
- **Roles:** `swift test --filter MCPAgentRoleDefaultsServiceTests` — RED before the guard (explicit pick dropped → `nil`), GREEN after; `clearOverride` still reverts.
- Relevant coordinated suites (Settings / AI / MCP / AgentMode): no regressions.
- SwiftFormat + `swiftlint --strict` clean on all changed files; `rpce-contribution-check commit` preflight + outgoing-range secret scan clean.
- Both fixes were independently adversarially reviewed (RepoPrompt Oracle for the Oracle path; a separate verification pass for the role path) and survived; the whitespace hardening came from that review.

## Context Builder — investigated, not a reset bug

The Context Builder agent/model was traced for the same failure class. The `didSet` "reset to default when invalid" branches do **not** durably reset a user's selection on a normal restart, because: CLI availability flags are seeded from persisted UserDefaults (a previously-connected provider reads available immediately), the menus only offer available agents, the menu handlers force-commit the user's real model, and all automatic/restore writers are `isSyncingSettings`/`isRestoringState`-guarded and gated on validation-complete. A guard there would be net-negative (it would drop a genuine user agent switch made before validation completes). The user-visible "Context Builder values differ" is the known **global-vs-workspace** drift (resolvable via *Use MCP/UI for both*); collapsing it to a single source is a separate planned change.

## Follow-ups (out of scope)
- Role-override writes don't flow through `recordSettingsWriteDiagnostic`, so role resets aren't reproducible from `recentSettingsWriteDiagnostics` — worth adding for parity with the Oracle path.
- Context Builder global/workspace single-source collapse.


datawisebets:
**Review pass (ponytail + high-recall code review)** — addressed in `7660f6a`:

- **In-memory Oracle mirror** (`PromptViewModel.setPreferredModelRaw`): the durable fix protected persistence, but the in-memory `_planningModel` mirror could still briefly blank the Oracle in the unhydrated fallback path until the next sync. Now guarded the same way (blank/whitespace not mirrored).
- **Stale doc comment** (`AppSettingsMCPService.postModelRawDidWrite`): the comment still claimed a null chat-clear mirrors into the Oracle; updated to match the new guard.
- **Dead parameter**: removed the now-unused `availability` arg from `MCPAgentRoleDefaultsService.setSelection` (left behind after dropping dedup-on-recommended) and its call sites.

Verified **benign / by-design** (no change): preserving a stale Oracle when a removed provider has no replacement (strict MCP surfaces "not available"); storing a role override that equals the current recommendation (consumers use `hasCustomOverride = sel != recommended`); the trimmed-empty predicate repeated across 3 guards.

Validation: `MCPAgentRoleDefaultsServiceTests` + `AIModelPreferenceRegressionTests` green (RED→GREEN on both reset bugs), SwiftFormat + SwiftLint `--strict` clean, contribution preflight + secret scan clean.

provencher (approved):

datawisebets:
**Oracle sign-off** (RepoPrompt CE oracle, `gpt-5.5-pro-extended`, full-diff review of all 3 commits):

1. **Correct and complete for both reset bugs? YES** — Oracle blanking is guarded across the GUI/store/MCP paths, and explicit sub-agent role picks now persist.
2. **Remaining reset/blanking path or regression not handled:** none found.
3. **Regression risk to legit flows? NO** — explicit Oracle clear, explicit role reset, and real (non-empty) model sync mirroring remain intact.

PR #319, opened 2026-06-29
fix(agent-mode): stop listener leaks and orphaned Claude processes on close

morluto:
## Summary

Agent Mode windows could leave behind registered callbacks, MCP tool observers, and Claude child processes after close. Reopening windows or switching workspaces could then hit stale hooks or orphaned subprocesses.

This PR adds removable workspace listener tokens, explicit teardown in `prepareForWindowClose`, a sync Claude process fallback in `deinit`, and moves Codex/Claude MCP tool-tracking cleanup off `coordinator.stop()` onto lifetime-only paths.

Beyond the core lifetime hardening, this PR also includes several supporting refactors and CI fixes:
- **Session index store extraction** — `AgentWorkspaceSessionIndexStore` with owner-epoch machinery to guard against stale workspace activations.
- **Workspace switch background cleanup provider** — `AgentModeWorkspaceSwitchCleanupProvider` encapsulates the detached async cleanup that runs after a workspace switch.
- **Listener registry consolidation** — `ListenerRegistry` unifies workspace/before-save listener tokens.
- **Per-tab teardown extraction** — `AgentModeViewModel+TabSession` extension centralizes session-level cleanup.
- **Mass `@Published` removal** — duplicate UI state vars that mirrored already-published sources were removed to eliminate redundant SwiftUI invalidations.
- **CI runner hardening** — `ci_app_test_runner.py` now runs suites via `xcrun xctest` directly to avoid SwiftPM per-invocation resolution wedge on hosted macOS runners.

~20 files, ~1600 insertions / ~600 deletions, across lifetime hardening, supporting refactors, and CI runner changes.

## Problem / impact

**1. Stale workspace listeners (accumulation bug — the primary fix)**

`WorkspaceManagerViewModel` kept workspace-switch and before-save callbacks in append-only arrays with **no removal API at all**. `AgentModeViewModel` registered those listeners on init but `prepareForWindowClose()` never unregistered them (only compose-tab close had a token).

This was impossible to fix without first adding the token-based `removeWorkspaceDidSwitchListener` / `removeBeforeSaveListener` APIs this PR introduces. Each Agent Mode window lifecycle appended more entries that could never be removed. Closures use `[weak self]`, so this is not a strong retain cycle on the view model, but:

- listener arrays grow without bound across repeated open/close
- dead callbacks are still invoked on every workspace switch/save (no-ops once `self` is gone, but wrong lifecycle semantics)
- a live `AgentModeViewModel` from a new window could see cross-talk if old registrations were still present

**2. Orphaned Claude child processes (resource leak)**

`ClaudeNativeProcessSessionController` normally shuts down via async `shutdown()`. If the controller is deallocated without that path (crash path, race, or missed teardown), the actor `deinit` did not terminate the child process — leaving a zombie `claude` subprocess and open pipes.

**3. Incomplete window teardown (listener / coordinator leak)**

`prepareForWindowClose()` stopped Codex/Claude coordinators but did not remove the workspace / before-save / compose-tab listeners it owned. Closing a window left those listener records alive on the long-lived `WorkspaceManagerViewModel` after the UI owner was gone (see #1 for why this was unfixable before this PR).

Note: per-tab MCP tool tracking was **already** drained per-session via `prepareSessionForWindowClose` -> `shutdownCodexSession` -> `stopCodexToolTrackingAndWait(for:)`. The new `stopAllCodexToolTracking()` in `stop()` is a safety net for orphaned tracker entries whose `TabSession` is no longer in `sessions`, not the primary cleanup path.

**4. Session delete race**

On tab/session delete, `cleanupACPStateForDeletedSession` ran before `agentTask?.cancel()`. An in-flight agent task could keep touching ACP state while teardown was already running.

**5. Tool tracking vs UI hide (correctness, not a leak)**

`coordinator.stop()` runs from two very different paths:

- **Teardown** — `prepareForWindowClose()`, `deinit`: sessions are being destroyed, tool tracking must be drained.
- **Visibility toggle** — `setAgentModeActive(false)` when the user switches to the Prompt tab: sessions keep running, the Codex process stays alive, only the UI is hidden.

Clearing tool tracking in `stop()` breaks the visibility-toggle path: it unregisters MCP tool observers for a run that is still in flight, so tool-call events stop arriving and are lost from the transcript. When the user switches back to Agent Mode, nothing re-establishes tracking for the in-flight run (`ensureCodexToolTrackingIfNeeded` only runs at turn setup, not on reactivation). Tool tracking now drains only on window close, workspace switch, and `deinit` fallback — not from `stop()`.

This is not primarily a "high memory" fix — it is lifecycle hygiene: unbounded listener growth, orphaned processes, and observers that outlive their owner.

## Review order

1. **`AgentModeViewModel.swift`** — `unregisterObserverRegistrations()`, `prepareForWindowClose()`, `deinit` fallback, workspace-switch path, `agentTask` cancel before ACP cleanup
2. **`AgentModeViewModel+TabSession.swift`** — `cancelEphemeralRuntimeState()` and per-tab teardown helpers
3. **`WorkspaceSwitchSessionProvider.swift`** — background cleanup provider; coordinators captured before yield so process termination survives provider dealloc; `cancelAllBackgroundCleanup` drains instead of cancelling to avoid hot-spin in `ProcessTermination.terminateAndReap` polling
4. **`AgentWorkspaceSessionIndexStore.swift`** / **`ListenerRegistry.swift`** — owner-epoch session index and unified listener token registry
5. **`CodexAgentModeCoordinator.swift`** / **`ClaudeAgentModeCoordinator.swift`** — tool tracking is **not** cleared in `stop()`; see `stopAll*ToolTrackingAndWait()`
6. **`ClaudeNativeProcessSessionController.swift`** — sync `deinit` cleanup with detached async reap fallback; normal path remains async `shutdown()`
7. **`WorkspaceManagerViewModel.swift`** — token add/remove APIs; compatibility overload still discards tokens
8. **`ci_app_test_runner.py`** / **`test_ci_app_test_runner.py`** — xcrun xctest direct invocation, fallback command prefix, multi-bundle guard
9. **`WorkspaceSwitchCleanupTests.swift`** + **`WorkspaceSwitchRecoveryTests.swift`** + ledger TSV — cleanup drain tests, listener removal test, `verify-ledger` reconciliation

## Behavior changes

| Area | Change |
|------|--------|
| Window / workspace teardown | Unregisters listeners, drains tool tracking, stops coordinator subscriptions/watchdogs |
| `setAgentModeActive(false)` (Prompt tab) | Coordinator `stop()` still runs, but **does not** clear MCP tool tracking |
| Session delete / tab close | `agentTask` cancelled before `cleanupACPStateForDeletedSession` |
| Claude native process | `deinit` releases IO handles, sends `SIGTERM`, and schedules a detached async reap via `ProcessTermination.terminateAndReap` |
| Workspace switch | Now awaits `cancelAgentRun(tabID:)` for every active session inline before clearing sessions (see latency note below) |
| Background cleanup cancellation | `cancelAllBackgroundCleanup` now drains (stops tracking) instead of cancelling tasks, to avoid hot-spin in process termination polling loops |

### Workspace switch latency note

`handleWorkspaceSwitch` now awaits graceful `cancelAgentRun(tabID:)` for every active session before proceeding to clear sessions and schedule background cleanup. This is necessary for correct routing/lifetime cleanup — the terminal commit barrier must finalize each run before its session is discarded. In practice this adds the provider terminal publication latency to the workspace switch path. If a provider terminal publication wedges, the switch blocks on that run. The terminal commit barrier has its own internal timeout, so this is bounded, but it is an unadvertised latency change to a user-facing operation.

## Test plan

- [x] `make dev-test FILTER=WorkspaceSwitchRecoveryTests`
- [x] `make dev-test FILTER=AgentToolTrackingControllerTests`
- [x] `make dev-test FILTER=AgentModeChatSwitchActivationTests`
- [x] `make dev-test FILTER=WorkspaceSwitchCleanupTests`
- [x] `python3 -m unittest test_ci_app_test_runner` (CI runner unit tests)
- [x] `python3 Scripts/test_suite_optimizer.py verify-ledger --ledger Scripts/Fixtures/test-suite-contract-ledger.tsv`


Sources/RepoPrompt/Features/AgentMode/ViewModels/AgentModeViewModel.swift
```diff
@@ -1787,6 +1789,27 @@ final class AgentModeViewModel: ObservableObject {
...
 
     deinit {
+        let promptManager = promptManager
+        let workspaceManager = workspaceManager
+        let tabCloseListenerToken = tabCloseListenerToken
+        let workspaceDidSwitchListenerToken = workspaceDidSwitchListenerToken
+        let beforeSaveListenerToken = beforeSaveListenerToken
+        let codex = codexCoordinator
+        let claude = claudeCoordinator
+        Task { @MainActor in
```
provencher: i dont like having a task like this setup. We should find a cleaner way to handle this.

Sources/RepoPrompt/Features/AgentMode/ViewModels/AgentModeViewModel.swift
```diff
@@ -2739,9 +2756,50 @@ final class AgentModeViewModel: ObservableObject {
...
+            beforeSaveListenerToken: beforeSaveListenerToken
+        )
+        tabCloseListenerToken = nil
+        workspaceDidSwitchListenerToken = nil
+        beforeSaveListenerToken = nil
+        cancellables.removeAll()
+    }
+
+    @MainActor
+    private static func removeRegisteredObserverListeners(
```
provencher: try and reduce the workspace specific code here. Might be good to find a cleaner arch for all this cleanup and binding

Sources/RepoPrompt/Features/AgentMode/ViewModels/AgentModeViewModel.swift
```diff
@@ -9910,6 +9968,8 @@ final class AgentModeViewModel: ObservableObject {
         if workspace != nil {
             publishLoadingTranscriptPresentation(tabID: initialActiveTabID)
         }
+        await codexCoordinator.stopAllCodexToolTrackingAndWait()
```
provencher: Ideally shutdown doesnt block on waiting for all this tool tracking, or at least the work in the stop should be a bare minimum to avoid adding shutdown latency

provencher (dismissed):
Almost there. I just need the code to be a bit cleaner

Sources/RepoPrompt/Features/AgentMode/ViewModels/AgentModeViewModel.swift
morluto: understood, taking a look now

morluto: Update: 
The manual token capture + fire-and-forget Task pattern is gone — Phase 1 introduced a ListenerRegistry that owns token-based registrations. unregisterObserverRegistrations() i s now just listeners.removeAll()

No need for a task anymore

Sources/RepoPrompt/Features/AgentMode/ViewModels/AgentModeViewModel.swift
morluto: Implemented refactoring and architecture cleanups:
   • AgentWorkspaceSessionIndexStore now owns the session-index data + owner-epoch validation (Phase 2)
   • WorkspaceSwitchSessionProvider owns the detached background cleanup (Phase 5)
  • TabSession.cancelEphemeralRuntimeState() consolidates per-tab teardown (Phase 4)
  • Duplicate @Published UI vars collapsed onto existing UI stores (Phase 3)
The VM keeps the switch orchestration (~180 lines) and refresh-token flow because they're too coupled to VM state to extract cleanly, but the workspace-specific data and cleanup are now in dedicated types.

Sources/RepoPrompt/Features/AgentMode/ViewModels/AgentModeViewModel.swift
morluto: Update:
Replaced with detachAllClaudeToolTrackingHandlersForWorkspaceSwitch(): it removes handlers from the coordinator map synchronously and fires Task { await handler.
stopTracking(for:) } per tab without waiting. The workspace-switch foreground path no longer blocks on tool tracking drain. The background cleanup path also now passes detachedRunID to shutdownClaudeSession so it clears only the matching handler instead of skipping cleanup entirely.

morluto:
ready for rereview @provencher

baron (commented):
# PR #319 Review — fix(agent-mode): stop listener leaks and orphaned Claude processes on close

Reviewed at PR head `5effdc5e` against `main`. Scope observed: 20 files, +1610/−608. CI on this head appears green.

## Verdict

**Approve with comments** — the PR appears to fix the advertised problems: removable listener tokens, teardown-scoped tool tracking, and Claude native process cleanup fallback. I did not find a merge-blocking correctness issue. That said, the PR description is stale relative to the actual branch, and there are several medium/low-risk issues worth addressing before or soon after merge.

## Findings

### M1 — PR description no longer describes the branch

The body says “7 files, ~232 lines, 5 commits,” but the branch currently touches 20 files with +1610/−608 and substantially more commits. It includes a session-index-store extraction, background-cleanup provider extraction, mass `@Published` removals, and CI runner changes.

The “Review order” section also references changes that do not appear to exist in this diff. This makes the review surface look much smaller and safer than it actually is.

**Suggested fix:** update the PR body before merge so reviewers/maintainers understand the real risk surface.

### M2 — `cancelAllBackgroundCleanup()` can cancel cleanup that does not actually stop, with possible hot-spin behavior

`AgentModeViewModel.prepareForWindowClose()` now cancels in-flight workspace-switch background cleanup. However, the cleanup task body in `WorkspaceSwitchSessionProvider.swift` does not appear to check `Task.isCancelled` in a way that would stop the cleanup body. Cancellation therefore reads as “stop cleanup,” but the cleanup can continue anyway.

A worse edge: cancellation may cause sleeps inside process termination polling to return immediately, which can turn a bounded wait loop into a busy poll on the main actor while slow child processes are being terminated.

**Suggested fix:** either let cleanup drain instead of cancelling it, or make cancellation explicit and hand residual process termination to a cancellation-shielded detached task.

### M3 — Window-close/workspace-switch now hard-cancels `agentTask` before graceful `cancelAgentRun`

`prepareSessionForWindowClose` and workspace-switch discard paths call `cancelEphemeralRuntimeState()` before `cancelAgentRun(tabID:)`. That cancels and nils task handles before the graceful cancellation path runs.

The terminal commit barrier probably prevents double-commit damage, but this reverses previous ordering on a teardown-race-prone path.

**Suggested fix:** add an explicit test covering this ordering or document why the new ordering is safe.

### M4 — Workspace switch now awaits graceful cancellation of every active run inline

`handleWorkspaceSwitch` now awaits `cancelAgentRun(tabID:)` for active sessions before proceeding. This is probably necessary for correct routing/lifetime cleanup, but it is an unadvertised latency change to a user-facing workspace switch operation.

**Suggested fix:** mention this in the PR body and consider a timeout/guard if provider terminal publication can wedge.

### L1 — Broken `xcrun` fallback in CI runner

If `xcrun --find xctest` fails, `xctest_binary_path()` falls back to returning `"xcrun"`. The eventual command becomes roughly:

```bash
xcrun -XCTest <suite> <bundle>
```

That is not a valid `xcrun xctest` invocation.

**Suggested fix:** represent the fallback as a command prefix such as `["xcrun", "xctest"]`, or otherwise make sure the fallback invocation includes the `xctest` tool name. Add a unit test for the fallback path.

### L2 — Bundle discovery can silently choose the wrong test bundle in the future

`discover_test_bundle` picks the first sorted `*.xctest` under the bin path. Today there may only be one bundle, but if another test target is added later, suites could run against the wrong bundle. Depending on `xctest` behavior for unmatched suites, this risks silent test loss.

**Suggested fix:** fail if multiple candidate bundles exist, or assert each requested suite actually ran at least one test.

### L3 — Dead DEBUG scaffolding in workspace-switch cleanup provider

`test_backgroundCleanupDrainWaiters` and `test_backgroundCleanupDrainTimeoutTasks` are drained/read but never populated. These appear to be leftovers from an older continuation-based drain implementation.

**Suggested fix:** remove them.

### L4 — Claude native `deinit` fallback sends SIGTERM but usually does not reap

The synchronous deinit fallback sends SIGTERM and then does a single non-blocking `waitpid`. That will often return before the child exits, leaving a zombie until process exit. This is acceptable as a last-resort fallback, but a detached async reap would be cleaner.

**Suggested fix:** consider `Task { await ProcessTermination.terminateAndReap(pid:) }` where safe.

### L5 — Residual orphan window remains for ACP/Codex if VM/provider deallocates before cleanup resumes

The background cleanup task still has a `guard let self else { return }` shape after an initial yield. If the provider/view model deallocates before the task resumes, ACP/Codex cleanup may not run. Claude native now has a deinit fallback, but ACP/Codex do not appear to have equivalent protection.

This looks pre-existing rather than a new regression, but it is relevant to the PR’s orphan-process framing.

## Checked and did not find an issue

- `@Published` removals: UI access appears to go through snapshot/pipeline paths; no obvious lost UI update found.
- Session-index store extraction: owner validation, local overlay, sort-date rebuild, and delegate notification behavior appear faithful.
- Tool tracking vs visibility toggle: stopping Agent Mode visibility no longer appears to detach Claude tool handlers prematurely; run-ID guarded detach path looks correct.
- Before-save listener removal timing: final workspace persistence appears to happen before teardown on close/termination paths.
- Restore-support refactors around preferred entries by tab ID appear behavior-preserving.
- Ledger TSV change looks mechanically consistent, though see validation note below.

## Validation notes

Observed/ran:

- PR checks appeared green on this head.
- `python3 Scripts/test_ci_app_test_runner.py` passed locally: 14/14.
- `verify-ledger` was attempted but did not produce a reliable final result in the review window. Treat ledger verification as **not confirmed** from this review alone.

Recommended before merge:

```bash
make dev-test FILTER=WorkspaceSwitchCleanupTests
make dev-test FILTER=WorkspaceSwitchRecoveryTests
make dev-test FILTER=WorkspaceFileContextStoreCodemapSeamTests
python3 Scripts/test_suite_optimizer.py verify-ledger --ledger Scripts/Fixtures/test-suite-contract-ledger.tsv
```

## Bottom line

The fixes look real and generally careful. I would not block merge on the core implementation, but I would strongly prefer the PR body be updated before landing, and the cheap cleanup/fallback issues above are worth addressing either now or immediately after merge.

baron:
@provencher — follow-up check of your earlier comments against current head `5effdc5e`:

**1. Manual token capture + fire-and-forget `Task` in listener teardown — resolved.**
The pattern is gone. `Runtime/WorkspaceLifetime/ListenerRegistry.swift` now owns token registrations (`addToken(_:remover:)` → single `removeAll()`), and `unregisterObserverRegistrations()` in `AgentModeViewModel.swift` is reduced to `listeners.removeAll()` + `cancellables.removeAll()`. `deinit` no longer spawns a task; it only does nonisolated-safe cancellations and documents that `prepareForWindowClose()` is the primary teardown path.

**2. Workspace-specific cleanup/binding code concentrated in the VM — substantially resolved, small residual.**
There's a new `Features/AgentMode/Runtime/WorkspaceLifetime/` seam with three focused types:

- `AgentWorkspaceSessionIndexStore` owns the session index, sort dates, local overlay upserts/removals, and owner/epoch validation; the VM participates through a delegate protocol.
- `AgentModeWorkspaceSwitchCleanupProvider` (`WorkspaceSwitchSessionProvider.swift`) owns detached background cleanup: MCP teardown via delegate, provider dispose, ACP shutdown, and coordinator session shutdown.
- `ListenerRegistry` owns listener token lifecycle.

Per-tab synchronous teardown shared by window close and workspace switch is consolidated in `TabSession.cancelEphemeralRuntimeState()` (`AgentModeViewModel+TabSession.swift`). The remaining `handleWorkspaceSwitch` orchestration and `prepareWorkspaceSwitchSessionDiscard` logic still live on the VM, but they touch a lot of VM-owned state, so extracting them further would mostly move coupling rather than remove it. This looks like an acceptable stopping point for this PR; further slimming of the VM can be a separate effort.

**3. Shutdown blocking on tool-tracking drain — resolved.**
`stopAllCodexToolTrackingAndWait` no longer exists. The foreground switch path now calls synchronous `ClaudeAgentModeCoordinator.detachAllClaudeToolTrackingHandlersForWorkspaceSwitch()`, which clears the handler map and fires per-handler `Task { stopTracking }` without awaiting. `codexCoordinator.stop()` / `claudeCoordinator.stop()` are synchronous task-cancellation only. Heavier teardown moved to the background cleanup provider, which also passes `detachedRunID` into `shutdownCodexSession` / `shutdownClaudeSession` so only the matching handler is cleared.

The remaining foreground awaits are for active-run cancellation through terminal publication and worktree ownership release. Those look like correctness waits rather than tool-tracking drains.

**4. “Code a bit cleaner” — addressed in my view.**
The WorkspaceLifetime extraction gives cleanup/binding logic named owners with clearer seams. Small nit: the DEBUG-only test-drain plumbing in `AgentModeWorkspaceSwitchCleanupProvider` is bulky, but it is test-scoped and appears to exist to stabilize the switch-cleanup tests.

Summary: the three inline comments look resolved in the current code. The remaining VM-resident switch orchestration is a defensible boundary rather than a missed cleanup.

morluto:
updated the description + addressed the concerns @baron @provencher 

this PR is ready for review

provencher (approved):

PR #322, opened 2026-06-29
Fix duplicate validated root scope crash

antonioc-cl:
Summary:
- Avoid `Dictionary(uniqueKeysWithValues:)` traps for validated session root refs with duplicate UUIDs.
- Keep validated scope matching strict by root ID + current standardized path.
- Add regression coverage for duplicate physical root refs during lookup/catalog access.

Validation:
- `make dev-test FILTER=WorkspaceFileContextStoreTests` PASS.
- `make dev-lint` PASS.
- `make dev-swift-build PRODUCT=RepoPrompt` PASS.
- `git diff --check` PASS.
- `make guardrails` PASS.
- Push preflight full `make dev-test` FAILS outside this change: `AgentOraclePillRoutingTests.testExactInMemoryResolutionUsesUUIDOrShortIDInsteadOfLatestSession` returned nil at lines 119 and 126.


dah:
We waited too long to merge this and now it's been superseded by another merge which fixes the same root-ID crash. I'm sorry about that.

PR #347, opened 2026-07-02
Optimize workspace switch load readiness

baron:
## Summary

- Defer outgoing-root checkbox reconciliation during workspace switches and reconcile once after target roots hydrate and selection replay completes.
- Start normal target-root hydration concurrently with restore-state work so the switch overlay can hide as soon as restore is ready and primary roots are visible.
- Keep same-workspace-ID replacement reloads on the serial path to preserve existing cancellation/recovery semantics.
- Skip reusable snapshot admission for primary workspace root loads; session-worktree loads keep the prior serial admission path.
- Add a generation/cancellation guard before overlapped root-load teardown can unload roots.

## Validation

Passed:

- `make dev-lint`
- `make dev-test FILTER=WorkspaceSwitchRecoveryTests`
- `make dev-test FILTER=WorkspaceFileContextStoreTests`
- `make dev-test FILTER=MCPCodeStructureWorktreeTests` after removing background session-worktree admission
- Live debug workspace-switch measurement pass via `__repoprompt_debug_diagnostics`

Caveat:

- `.agents/skills/rpce-contribution-check/scripts/preflight.sh pr-ready` reran lint successfully, then the full root `swift test` lane failed in `MCPCodeStructureWorktreeTests.testSelectedScopeRejectsAtLimitPlusOneWithoutContentOrCodemapWork` only when run in the full-suite order. The exact method and the full `MCPCodeStructureWorktreeTests` suite pass focused after the cleanup. The failure appears order/environment-dependent rather than a deterministic failure of the touched workspace-switch path.


baron:
Follow-up context for the perf work, so this is not hand-wavy.

## What changed and why

This PR is intentionally a small reordering/removal-shaped optimization rather than a broad hardening pass:

1. **Skip outgoing-root checkbox reconciliation during switch restore.**
   `restoreWorkspaceState` no longer pays `refreshRootFolderState()` for roots that are about to be unloaded. The switch path now performs one explicit reconciliation after target-root hydration and selection replay, which is the point where the checkbox state actually matters.

2. **Overlap normal root hydration with restore-state work.**
   For ordinary workspace switches, target `loadWorkspaceFolders` starts while restore applies workspace/tab state. The overlay waits for both restore readiness and primary-root visibility before hiding.

3. **Keep same-ID replacement reloads serial.**
   Same-workspace-ID reload/cancellation recovery is deliberately left on the older serial path because those tests encode subtler recovery semantics. This avoids adding synchronization complexity just for theoretical coverage.

4. **Remove primary-workspace reusable snapshot admission from the load path.**
   The measured snapshot admission cost was far larger than the root walk it was meant to accelerate, so primary workspace root loads no longer run that admission. Session-worktree roots keep the prior serial admission behavior; I initially tried backgrounding that too, but it introduced order-dependent MCP test instability, so I removed that extra machinery.

5. **Guard overlapped destructive unload.**
   The overlapped load path checks cancellation/generation before the initial unload step, so a superseded/cancelled load cannot late-wipe roots.

## Measurements

Baseline live debug switch, before this branch:

- Overlay/root shell visible: **3120.7 ms**
- Total switch: **5081.6 ms**
- Reusable snapshot/background admission was observed around **~31 s**, which is too expensive to justify on the primary load path.

After this branch, live debug switches from `Default -> repoprompt-ce-upstream`:

Sample 1:

- Root load trace start: **281.1 ms**
- Root record created: **443.7 ms**
- First prepared chunk: **476.4 ms**
- Root load walk: **616.0 ms**
- Root load commit: **19.3 ms**
- Root load end: **639.9 ms**
- Root shell visible: **1081.6 ms**
- Overlay hidden: **1082.4 ms**
- Search index barrier: **385.1 ms**
- Total switch: **2315.3 ms**

Sample 2:

- Root load trace start: **487.6 ms**
- Root record created: **748.8 ms**
- First prepared chunk: **801.3 ms**
- Root shell visible: **1606.5 ms**
- Overlay hidden: **1607.3 ms**
- Search index barrier: **448.9 ms**
- Total switch: **3018.0 ms**

Best observed delta:

- Visible overlay duration: **3120.7 ms -> 1082.4 ms** (~65% faster)
- Total switch: **5081.6 ms -> 2315.3 ms** (~54% faster)

The second sample still improved but shows the remaining root problem: root identity/shell publication is still coupled too tightly to fuller catalog hydration and downstream readiness.

## Validation notes

Passed focused/local checks include:

- commit + push contribution preflights
- `make dev-lint`
- `make dev-test FILTER=WorkspaceSwitchRecoveryTests`
- `make dev-test FILTER=WorkspaceFileContextStoreTests`
- `make dev-test FILTER=MCPCodeStructureWorktreeTests` after removing the background session-worktree admission experiment

One caveat: the `pr-ready` full root lane still failed in full-suite order at `MCPCodeStructureWorktreeTests.testSelectedScopeRejectsAtLimitPlusOneWithoutContentOrCodemapWork`, while the exact method and the full MCP code-structure suite pass focused. I’m calling that out because it should not be hidden, but the deterministic coverage for the touched switch path is green.

## Suggested next steps

To get reliably sub-second, and possibly closer to ~500 ms on modern machines, the next PR should probably be architectural rather than more micro-hardening:

1. **Split root identity/shell publication from full catalog hydration.**
   Publish enough root identity for the sidebar shell around root record / first chunk time, then hydrate the deeper catalog behind it. The current post-change samples still show shell visibility closer to root load completion than first-root availability.

2. **Measure and isolate selection replay / token recount costs.**
   If selection replay or recount is still on the visible path, move more of it behind the shell-ready gate where safe.

3. **Profile the per-entry root walk.**
   The root walk is still hundreds of ms. The next target should be ignore/resource/symlink/policy checks per entry, but only with a narrow benchmark so we don’t optimize ghosts.

4. **Revisit reusable snapshot economics separately.**
   If reusable admission comes back, it needs a cheaper evidence path than per-path revalidation. Otherwise it is not worth complicating the workspace-switch critical path.

baron:
Cleanup pass pushed in `1cb2088c` in response to the diagnostics/test minimization feedback.

What changed:

- Moved the new search-index diagnostic field formatter out of `WorkspaceManagerViewModel` into `Sources/RepoPrompt/Features/Diagnostics/App/WorkspaceSwitchSearchIndexDiagnostics.swift`.
- Removed the DEBUG-only `refreshRootFolderStateDidRunHandlerForTesting` hook from `WorkspaceFilesViewModel`.
- Removed the added implementation-detail test that existed mainly to assert the internal refresh phase ordering.

Why:

- The functional switch/load change is small, and the diagnostic payload shaping does not need to live in the core workspace manager.
- The removed test had a weak public oracle without the test hook; its strongest assertion was internal timing/phase behavior, so it was not worth adding core-only test surface plus runtime weight.
- Existing workspace-switch recovery/selection tests still exercise the user-visible selection preservation paths, while the PR now relies on the smaller code path and review for the specific refresh deferral ordering.

Validation after cleanup:

- `make dev-format` — passed, formatted the new diagnostics file
- `make dev-test FILTER=WorkspaceSwitchRecoveryTests` — passed, 21/21
- `make dev-lint` — passed, 0 violations / 0 format issues
- `.agents/skills/rpce-contribution-check/scripts/preflight.sh commit` — passed

PR #349, opened 2026-07-02
Make git fixtures hermetic and hang-proof codemap engine test

### LINKED ISSUE #348: WorkspaceCodemapBindingEngineTests hang forever at 100% CPU when host git config has core.autocrlf=input (non-hermetic fixtures + unbounded test poll)
**Summary**

The full root suite (make dev-test) cannot complete on machines with core.autocrlf=input in the global gitconfig — a common, valid setting. testCancelledManifestAdoptionWaiterReleasesAdmissionWithoutCancellingSharedLoad spins indefinitely at 100% CPU; the rest of the warm-manifest-seed test family fails (boundedly) on the same machines. Reproduced at HEAD and at 5993203 (the commit that added the test), so this is not a regression from the 2026-06-29 engine work.

**Reproduction**

	1.	git config --global core.autocrlf input
	2.	swift test --filter WorkspaceCodemapBindingEngineTests/testCancelledManifestAdoptionWaiterReleasesAdmissionWithoutCancellingSharedLoad → never terminates, 100% CPU
	3.	Sibling warm-seed tests (e.g. testPathInvalidationDuringBlockedWarmAdoptionRemainsRetryable) fail boundedly under the same condition.

**Root cause chain (runtime-verified)**

	1.	Fixtures inherit host git config. Fixture repos don’t pin core.autocrlf; the engine reads it via plain git config --get (GitService.swift:6287). The classifier flags any value ≠ "false" as a checkout transformation (GitBlobIdentityService.swift:443), so the seed build gets no clean git-blob association.
	2.	Warm manifest is silently never persisted. Worktree-classified completions publish with association: nil, so clean manifest persistence is skipped with zero telemetry (WorkspaceCodemapBindingEngine.swift:6030–6041; instrumented run: manifestWrites=0, manifestFailures=0, no manifest shard on disk).
	3.	Warm load misses before the test’s gate hook. loadCurrentManifest returns .miss at the missing-shard guard (CodeMapRootManifestStore.swift:285–290), before afterReadAdmission() (:319) — the test’s loadGate.enter() never fires.
	4.	Test converts the miss into an infinite spin. waitUntilEntered() times out (10s) returning false; XCTAssertTrue(sharedLoadEntered) is non-fatal, so the test continues. The first demand completes as a terminal miss, the second demand is admitted rather than queued, and the unbounded poll while await fixture.engine.accounting().queuedRequestCount != 1 { await Task.yield() } (WorkspaceCodemapBindingEngineTests.swift:2820) busy-spins forever.

**Proposed fixes**

	1.	Make fixture repos hermetic: pin core.autocrlf=false (or run fixture git with GIT_CONFIG_GLOBAL=/dev/null + GIT_CONFIG_NOSYSTEM=1) in the repository fixture helper. Fixes the whole warm-seed family.
	2.	Make the test hang-proof: replace the unbounded Task.yield() poll with the existing bounded waitForEngineCondition helper (:5196), and make the sharedLoadEntered check a fatal guard. Audit the file for other unbounded polls.
	3.	(Product follow-up, maintainer call): warm-start manifest persistence is silently disabled for any real user with core.autocrlf=input — no counter or event marks the skip, and .durabilityFailure is swallowed. Even if the conservative classification is intentional, it should be observable.

Validation for a fix PR: make dev-test FILTER=WorkspaceCodemapBindingEngineTests on a machine (or env) with core.autocrlf=input, plus make guardrails / make dev-lint.

BetaFold3:
Fixes https://github.com/repoprompt/repoprompt-ce/issues/348

## Problem

On machines with `core.autocrlf=input` in the global gitconfig (a common, valid
setting), the full root suite cannot complete:
`WorkspaceCodemapBindingEngineTests.testCancelledManifestAdoptionWaiterReleasesAdmissionWithoutCancellingSharedLoad`
spins at 100% CPU forever, and the rest of the warm-manifest-seed test family
fails. Reproduced at HEAD and at the commit that introduced the test — not a
regression, a latent host-config leak.

## Root cause

Fixture repos never pinned checkout-transform config. The product reads
`git config --get core.autocrlf` at engine runtime with the host environment
(`GitService.gitBlobCheckoutConfiguration`), so the host's global value leaked
into fixture repos. The classifier treats any value ≠ `false` as a checkout
transformation (`GitBlobIdentityService`), which silently skips warm-manifest
persistence — the warm fixture then misses before the test's gate hook, and the
test's unbounded `while … { await Task.yield() }` poll spun forever on an
unsatisfiable condition.

## Changes (test-only, 2 files)

- `ReviewGitRepositoryFixture.makeRepository`: pin repo-local
  `core.autocrlf=false`, `core.eol=native`, `core.attributesfile=/dev/null`.
  Tests that want transform behavior can still override after creation.
- Hang-proof the affected test: bounded wait + fatal guards (which release the
  test gate and cancel the in-flight demand), bounded final await.
- New characterization test codifying the current silent persistence-skip under
  `core.autocrlf=input`, so any future change to that behavior is deliberate.

Deliberately untouched: the file's four other `Task.yield()` polls sample
transient states; bounding them at 10 ms granularity was measured to cause
failures and was reverted.

## Validation (on a machine with host `core.autocrlf=input`)

- Previously-hanging test: passes focused (~2.6 s).
- Previously-failing warm-seed siblings
  (`testPathInvalidationDuringBlockedWarmAdoptionRemainsRetryable`,
  `testDemandWarmLocatorAndCASBypassUnstartedManifestAdoption`): pass.
- Characterization test: passes.
- Fixture blast radius: 9 consumer suites run focused; 7 fully green, and the
  pins *reduce* failures in the other two (`GitBlobIdentityServiceTests`
  18 → 2 vs clean main; `WorkspaceFileContextStoreCodemapSeamTests` 21 → 1;
  residuals reproduce on clean main).
- `make guardrails`, `make dev-format`, `make dev-lint`: clean.

### DIFF
--- a/Tests/RepoPromptTests/Helpers/ReviewGitRepositoryFixture.swift
+++ b/Tests/RepoPromptTests/Helpers/ReviewGitRepositoryFixture.swift
@@ -35,6 +35,11 @@ final class ReviewGitRepositoryFixture {
         _ = try runGit(["config", "user.name", "RepoPrompt Test"], at: root)
         _ = try runGit(["config", "user.email", "repoprompt@example.test"], at: root)
         _ = try runGit(["config", "commit.gpgSign", "false"], at: root)
+        // The product reads checkout config at engine runtime with the host environment, so
+        // fixture repos pin transform-relevant keys locally to stay hermetic against host gitconfig.
+        _ = try runGit(["config", "core.autocrlf", "false"], at: root)
+        _ = try runGit(["config", "core.eol", "native"], at: root)
+        _ = try runGit(["config", "core.attributesfile", "/dev/null"], at: root)
         _ = try runGit(["checkout", "-b", "main"], at: root)
 
         for (path, contents) in files {
--- a/Tests/RepoPromptTests/WorkspaceContext/WorkspaceCodemapBindingEngineTests.swift
+++ b/Tests/RepoPromptTests/WorkspaceContext/WorkspaceCodemapBindingEngineTests.swift
@@ -2283,6 +2283,33 @@ final class WorkspaceCodemapBindingEngineTests: XCTestCase {
         XCTAssertEqual(source, .cleanManifest)
     }
 
+    /// Characterization test codifying the current silent-skip behavior described in
+    /// docs/investigations/codemap-binding-engine-test-hang-2026-07-02.md. The codemaps owner
+    /// may deliberately change this behavior; update this test in the same commit if so.
+    func testAutoCRLFInputClassifiesWorktreeAndSkipsManifestPersistence() async throws {
+        let repository = try makeRepositoryFixture(name: #function)
+        let root = try repository.makeRepository(
+            named: "repository",
+            files: ["Sources/Warm.swift": "struct Warm {}\n"]
+        )
+        _ = try repository.runGit(["config", "core.autocrlf", "input"], at: root)
+        let artifactRoot = try makeSecureDirectory(in: repository.sandbox, named: "artifacts")
+        let runtime = try CodeMapArtifactRuntime(rootURL: artifactRoot)
+        let fixture = try await makeEngineFixture(root: root, runtime: runtime)
+        _ = await fixture.engine.registerRoot(fixture.registration)
+
+        guard case .ready = await fixture.engine.demand(fixture.demand(path: "Sources/Warm.swift")) else {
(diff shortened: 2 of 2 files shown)


BetaFold3:
Appeared that a couple of engineers and their agents thought highly of this PR's idea but worked and (about to) merge their own implementation. This PR does its job and could be closed then :)

Issue #291, opened 2026-06-23
OpenCode ACP runtimes can remain orphaned after Agent Mode tab/window cleanup leading to extreme CPU memory pressure


morluto:
## Summary

I've been using DeepSeek v4-flash in OpenCode for a while now as my explorer/context-builder agent in RepoPrompt CE, and noticed that my system memory was under constant pressure. After digging in, I found that OpenCode-backed Agent Mode sessions can leave `opencode acp` subprocesses running after the associated Agent Mode sessions are no longer visible or cleanup-addressable through MCP/session listing APIs.

This appears separate from the intentional completed-session ACP retention used for follow-up prompts. The bug case is: the session disappears from UI/MCP reachability, but the retained ACP controller/process is not shut down.

## Impact

Each orphaned OpenCode ACP runtime can retain hundreds of MB to 1GB+ RSS. Because the session is no longer listed, users cannot clean it up through normal `agent_manage.cleanup_sessions` or UI session cleanup flows.

## Observed behavior

After running OpenCode-backed Agent Mode child sessions and later closing/cascading/removing the corresponding tabs/sessions:

- `agent_manage.list_sessions` showed no relevant OpenCode sessions.
- The UI showed no remaining subsessions.
- Live process inspection still showed multiple `opencode acp` subprocesses under the RepoPrompt CE app.
- Those processes also had RepoPrompt CE debug CLI child processes, indicating they were still live ACP runtimes rather than harmless stale rows.

## Expected behavior

If an Agent Mode session/tab/window is closed, deleted, cascaded, or otherwise removed from session visibility, any retained integrated ACP controller for that session should be explicitly shut down.

Completed ACP session retention is useful for follow-up prompts, but it should only persist while the owning session remains reachable and cleanup-addressable.

## Suspected root cause

Integrated ACP completion intentionally retains the controller for successful runs:

- `Sources/RepoPrompt/Features/AgentMode/Runtime/Runners/ACPIntegratedAgentModeRunner.swift`
  - completed runs finalize with `shouldShutdownController: outcome.terminalState != .completed`
  - this keeps `session.acpController` alive for follow-up reuse

That behavior is safe only if all later session-removal paths release the retained controller.

The good cleanup path exists:

- `Sources/RepoPrompt/Features/AgentMode/ViewModels/AgentModeViewModel.swift`
  - `cleanupACPStateForDeletedSession(...)` calls `controller.cancelPrompt()` and `controller.shutdown()`
[...]

PR #360, opened 2026-07-03
Reduce Agent Mode memory from retained provider processes

morluto:
## Summary

This PR reduces Agent Mode memory pressure from retained provider subprocesses while preserving short-term session reuse.

Local investigation found two related lifecycle problems:

1. **Retained live provider processes**
   - All supported provider families can retain subprocesses for reuse, but their lifecycle gaps differed.
   - ACP/OpenCode and Claude-compatible runtimes could remain alive indefinitely after completed conversations because they lacked Codex's existing five-minute idle shutdown.
   - Codex Agent Mode already had bounded idle cleanup, but its separate model-polling app-server remained resident between refreshes.
   - Retained provider processes were the main observed source of memory pressure: individual ACP/OpenCode processes consumed hundreds of MB, sometimes around 1 GB each; the Codex polling process added a smaller but unnecessary resident process.

2. **Cancellation-sensitive cleanup**
   - Cancelling idle-cleanup ownership could propagate cancellation into process termination.
   - That could enter the busy-poll behavior tracked in #481.
   - Teardown now runs in an independent, non-cancelled task after runtime ownership is revalidated.

The PR adds bounded cleanup for retained runtimes, hardens terminal and shutdown ordering, and releases the smaller Codex model-polling app-server when it is idle.

## Local memory observations

This work was motivated by investigation of serious memory pressure in the RepoPrompt CE debug app. We inspected process trees, sampled stack traces, and correlated retained child processes with Agent Mode sessions.

Observed retained subprocess RSS in that environment:

- OpenCode/ACP: often hundreds of MB, with samples around **700 MB–1 GB RSS per `opencode acp` process**.
- Claude-compatible runtime: one completed probe retained about **305 MB RSS** until explicit cleanup.
- Codex model-polling app-server: about **60 MB RSS** while resident.

These values vary by provider and workload, but the process attribution showed that unbounded live provider retention could dominate RepoPrompt's memory footprint.

### Provider runtime retention

Successful runs intentionally retain their controller briefly so a follow-up can reuse the provider session. Codex Agent Mode already bounded that reuse window with a five-minute idle shutdown. ACP/OpenCode and Claude-compatible runtimes did not have an equivalent bound, so completed sessions could keep large provider processes alive indefinitely.

Codex was still part of the observed process-retention problem through a separate lifecycle: its dedicated model-polling app-server stayed resident between refreshes even though model snapshots were cached and the client could restart on demand.

### Cleanup cancellation

The first idle-cleanup implementation tied provider shutdown to the cancellable timer task. If ownership changed after teardown began, cancellation could propagate into process termination. The final implementation atomically revalidates and detaches the exact controller, then completes shutdown in an independent task.

### Codex model polling

Codex model polling used a dedicated non-agent app-server and kept it resident between refreshes. The snapshot is already cached and the client can restart on demand, so the service now stops the owned app-server after refresh.

Review also found a shutdown race: waiting for a blocked model request before stopping its app-server could deadlock window teardown. Idle and shutdown cleanup now interrupt an in-flight request from an independent task, serialize that stop with refresh completion, and safely restart polling if a subscriber arrives concurrently.

## Retained-runtime lifecycle

```mermaid
flowchart TD
    A[Agent run completes] --> B{Terminal publication accepted?}
    B -- no --> C[Do not schedule idle cleanup]
    B -- yes --> D[Retain controller briefly for follow-up reuse]
    D --> E{New work or lifecycle transition?}
    E -- yes --> F[Cancel pending idle ownership or perform explicit cleanup]
    E -- no, TTL expires --> G{Still safe to teardown?}
    G -- no --> H[Keep controller]
    G -- yes --> I[Detach exact controller]
    I --> J[Shutdown in independent task]
    J --> K[Preserve authoritative resume metadata]
```

## What changed

- Added DEBUG-only child PID and RSS attribution for Codex, ACP/OpenCode, and Claude-compatible controllers.
- Added one shared idle-shutdown policy and scheduler for ACP and Claude-compatible runtimes.
  - five-minute production TTL
  - test-injectable delay
  - token-owned tasks so stale cleanup cannot remove a newer schedule
- Recheck teardown eligibility at expiry:
  - exact run, provider, and controller authority
  - completed and accepted terminal state
  - no active approvals, questions, tools, steering, binding transition, or child `agent_run.wait`
- Run provider shutdown independently from cancellable idle-task ownership.
- Preserve authoritative provider session identifiers before controller teardown.
- Tear down stale ACP controllers on provider changes, startup failure, tab closure, window closure, and incompatible new runs.
- Fix terminal barrier ordering so cleanup is scheduled only after accepted terminal publication.
- Stop the owned Codex model-polling app-server after refresh while retaining cached models.
- Serialize model refresh, subscriber return, idle stop, and window shutdown without deadlocking a blocked request.
- Reconcile the contract ledger for all current executable lifecycle tests.

## Validation

- `make dev-test FILTER=CodexModelPollingServiceTests`
- `make dev-test FILTER=AgentModeRunServiceLifecycleTests` — 50 tests passed
- `make dev-test FILTER=WindowCloseCoordinatorLifecycleTests` — 5 tests passed
- `make dev-lint`
- `python3 Scripts/test_suite_optimizer.py verify-ledger --ledger Scripts/Fixtures/test-suite-contract-ledger.tsv` — 3,166 exact IDs
- Commit and push contribution preflights passed

Hosted CI is running for the rebased head.

## Review order

1. `AgentModeIdleShutdownPolicy.swift` and the shared scheduler in `AgentModeRunService.swift`
2. ACP and Claude-compatible detach/shutdown paths
3. terminal publication and lifecycle gates
4. `CodexModelPollingService.swift`
5. focused lifecycle tests and contract-ledger entries

## Related issues

Follow-up to #291. The cancellation-safe shutdown work also avoids the busy-poll path tracked in #481.


morluto:
Fresh live evidence on July 11 supports the retained-runtime work in this PR:

- five loaded Codex sessions owned five isolated app-server processes
- those app-servers used approximately 120–173 MB RSS each
- the separate model-polling app-server used about 53 MB RSS

Process-per-session isolation is intentional; the evidence shows why bounded idle teardown and stopping the polling transport between refreshes matter.

A separate lifecycle/API problem is now tracked in #482: agent_manage.stop_session does not detach a completed runtime and does not expose attached-transport state. That issue is complementary to this PR rather than a request to expand its implementation scope.

morluto:
@baron this one is still a pretty big issue, bumping it here

morluto:
### Review update

After rebasing this PR onto current `main`, we performed a focused lifecycle and complexity review.

That review found and fixed two additional issues:

- Idle-cleanup cancellation could propagate into provider shutdown and enter the cancelled-task busy-poll path tracked in #481. Teardown now completes in an independent, non-cancelled task after exact runtime ownership is revalidated.
- Codex model polling could deadlock during window teardown by waiting for a blocked refresh before stopping the app-server needed to interrupt it. Refresh, stop, subscriber return, and shutdown are now serialized without blocking teardown.

The review also consolidated ACP and Claude-compatible idle cleanup behind one scheduler, added child `agent_run.wait` and pending-interaction gates, protected replacement schedules with ownership tokens, and preserved authoritative resume identifiers.

Focused lifecycle, polling, window-close, lint, product-build, and exact ledger validation now pass on the rebased head (`73b47985`).

baron:
@morluto awesome and amazing work! I'm going to just flood your issues with a review/comment wave from Sol/Fable so we can better process your PRs among the big contributors. Use your own discernment whether you take the feedback. Will try to process the easy decisions asap. Then talk over the rest so we're not making a rash decision.

baron (changes requested):
### Intent Analysis
**Problem:** Completed Agent Mode sessions retain heavyweight provider subprocesses indefinitely: Claude-compatible and ACP/OpenCode controllers had no bounded idle teardown (only Codex did), and the Codex model-polling app-server stayed resident after refresh despite cached results.

**Before → After:** Before: successful Claude/ACP runs retained controllers indefinitely; Codex alone had a 5-minute idle policy; Codex model polling stopped only on shutdown or subscriber-idle. After: a shared 5-minute idle policy (`AgentModeIdleShutdownPolicy.swift:1-15`) schedules teardown from accepted terminal `postCommit` hooks (`ClaudeIntegratedAgentModeRunner.swift:345-376`; `ACPIntegratedAgentModeRunner.swift:961-995`); teardown revalidates run/controller/interaction/steering/child-wait ownership (`AgentModeRunService.swift:1298-1434`), detaches controllers while preserving resume identifiers (`ClaudeAgentModeCoordinator.swift:489-534`; `WorkspaceSwitchSessionProvider.swift:228-249`), shuts down in a non-cancelled child task, and stops the Codex polling client after refresh with serialized refresh/stop/shutdown (`CodexModelPollingService.swift:145-185,206-229,251-294`). DEBUG-only process-memory diagnostics added.

**Evidence:** Two independent code-first reviews of all 23 changed-file patches, new tests, timeline, and related issues (#291, #481, #482) converge on the same conclusions with matching line refs; intent derived from the diff and tests, not PR prose. All 8 hosted checks pass at head `73b4798`; mergeable state is MERGEABLE.

**Confidence:** High — both independent reports rate high with line-for-line agreeing citations of the execution paths.

**Red Flags:** One-shot idle scheduler discards cleanup on transient safety failure (blocking finding below). The child-wait regression test manually reschedules cleanup after the wait drains — a call production never makes — and the test ledger repeats that overclaim. Large DEBUG-only diagnostics addition with non-atomic snapshots (noted, not blocking).

### Findings
**F1 — HIGH / BLOCKING: Safety-gated idle cleanup is discarded, not deferred — unbounded retention path survives.**
`scheduleIfNeeded` cancels/returns when a safety predicate is false at scheduling time (`AgentModeRunService.swift:1334-1342`); on TTL expiry a failed recheck returns and `defer` removes the sole scheduler entry with no retry (`AgentModeRunService.swift:1355-1365,1397-1399`); an active child `agent_run.wait` makes teardown unsafe (`AgentModeRunService.swift:1430-1432`). The only production scheduling sites are one-shot terminal `postCommit` hooks (`ClaudeIntegratedAgentModeRunner.swift:373-376`; `ACPIntegratedAgentModeRunner.swift:992-995`), and no production callback re-arms cleanup when the wait drains. So when terminal completion overlaps an active child wait — or a wait is active when the TTL fires — the retained Claude/ACP process persists indefinitely until an unrelated lifecycle action, violating the PR's own bounded-retention invariant. The regression test at `Tests/RepoPromptTests/AgentMode/AgentModeRunServiceLifecycleTests.swift:1046-1087` passes only by manually invoking `scheduleClaudeIdleShutdownIfNeeded` a second time after clearing the wait, masking the gap. Confirmed independently by both reviews with matching refs.

**Required fix:** Re-arm/defer cleanup after transient safety failures (or schedule from the child-wait drain transition); replace the test with a production-shaped one — schedule once, drain the wait asynchronously, assert eventual teardown — and correct the test-ledger entry that overclaims coverage.

No other blocking findings.

### Maintainer-guidance check
- **User impact / invariant:** Completed provider runtimes must be released after bounded idle time without disrupting active interactions and while preserving resume identity; F1 breaks exactly this on a realistic path (concurrent `agent_run.wait` at terminal publication or TTL expiry).
- **Root-cause confidence:** Confirmed — for both the pre-existing unbounded retention and the PR's one-shot scheduling gap.
- **Authority:** `TabSession` controller ownership and provider coordinators own runtime state; `AgentModeRunService` owns the idle policy; the MCP wait-scope registry owns child-wait state. Identity revalidation and cancellation-independent shutdown are sound.
- **State-safety risks:** No demonstrated data or resume-identity loss; the blocker is resource-lifetime liveness only.
- **Scale / observability:** Retained subprocess cost multiplies across sessions; DEBUG diagnostics improve attribution but dropped timers leave lifetimes unbounded.
- **Scope:** Fix the re-arm in this PR; keep #481 (shared cancellation-safe process handling) and #482 (explicit runtime-control API) as follow-ups.
- **Validation boundary:** Production-shaped one-shot-schedule + async-drain test proving automatic teardown, then live CE MCP validation with a concurrent `agent_run.wait` and accelerated-TTL cleanup observed via the process-memory diagnostics.

### Verdict
**REQUEST_CHANGES.** The lifecycle design is directionally strong — identity revalidation, resume-ID preservation, cancellation-safe shutdown, and polling serialization are well-built, and CI is fully green at `73b4798` with a MERGEABLE state. But both independent reviews confirm, with matching evidence, a concrete unbounded-retention path in the feature's core mechanism: safety-gated cleanup is discarded rather than deferred, and the sole regression test papers over it with a test-only second scheduling call. Since bounded teardown is the point of the PR, this single confirmed defect blocks merging. Fix the re-arm/deferral and the masking test; the rest is ready.

morluto:
Thanks @baron — addressed the REQUEST_CHANGES item on `79d53e9374d2f4849560bec8c177de0a8c88e230`:

**Blocking (F1)**
- Idle cleanup is no longer a one-shot that is discarded when teardown is unsafe.
- `scheduleIfNeeded` always arms a timer; if fire-time revalidation fails while the completed run still owns the retained controller (including active child `agent_run.wait`), it **re-arms** instead of dropping the only cleanup opportunity.
- New runs / agent switches still cancel via `cancel(for:)`.

**Test**
- `testClaudeIdleShutdownWaitsForActiveChildAgentRunWait` now schedules **once**, holds an active child wait past the TTL, clears the wait, and asserts autonomous teardown + resume identity preservation (no test-only second schedule).
- Ledger row updated to match the production-shaped contract.

**Non-blocking**
- DEBUG-only process-memory diagnostics / non-atomic snapshots: left as intentional DEBUG instrumentation (your review marked non-blocking). Happy to harden snapshots in a follow-up if useful.

**Sentry:** none on this PR.

**Validation:** local XCTest cannot run on this Linux agent (no Darwin/macOS SDK). Hosted CI will re-run on the new head. Ready for re-review when green.

morluto:
Addressed the non-blocking DEBUG snapshot follow-up in `696d0b60`:

- Agent session fields and controller identities are now captured synchronously without a `MainActor` suspension.
- Controller process probes run afterward against those frozen controller identities, preventing one session payload from mixing state across lifecycle moments.
- Window metadata/counts are captured before process awaits, and the existing diagnostics JSON shape is preserved.
- The consistency marker now documents the precise contract: session-atomic state with best-effort process probes.

Commit/push safety preflights passed. Hosted CI is running on the new head.

baron:
Recommend closing this PR form and extracting any remaining idle-runtime/API work into small current-main PRs. The comparison is now polluted across hundreds of files with extensive conflicts and red app builds; the original orphan-cleanup issue was closed by #292, while remaining concerns are tracked separately.

Issue #168, opened 2026-06-11
Concurrent MCP sessions: run-scoped routing silently rebinds connections bound with bind_context (binds drift to other sessions' tabs)


dbryson:
## Summary

With several independent MCP client sessions connected to one RepoPrompt instance, an explicit `bind_context(op:"bind", context_id:<id>)` reports success, but `bind_context(op:"status")` on the same connection moments later reports a different tab, typically one belonging to another concurrent session's run. Re-binding fixes it momentarily; the next tool call re-asserts the wrong tab. Frequency scales with session count: rare at 1-2 concurrent sessions, constant at 4-5.

After auditing the CE source (commit `f8bbff3`), this is not a per-window shared "active tab" pointer. Bindings are stored per connection (`tabContextByConnectionID`). The drift is caused by two interacting defects in run-scoped routing:

1. Run policy state is keyed by client *family*, so every Claude Code session and every spawned reviewer agent share one identity and one pending-policy queue.
2. Explicit `bind_context` bindings are not senior to run-scoped rebinding. Once a connection picks up any runID association, a per-tool-call fallback force-rebinds it to that run's tab, stomping the explicit binding on every call.

## Environment

- RepoPrompt on macOS, MCP over stdio (CLI shim to bootstrap socket), multiple simultaneous client connections (4-5 Claude Code sessions)
- One shared workspace, one window; each session creates its own compose tab (`manage_workspaces create_tab bind:true focus:false`) and binds it via `bind_context op=bind context_id=<id>`
- Sessions periodically run `agent_run` reviews (this is the necessary ingredient, see repro)
- Note: I observed this on a build close to CE; the analysis below references CE source at `f8bbff3`. CE HEAD's PID gating may partially mitigate the claim path, but the seniority defect (item 2) is present at HEAD.

## Observed signature

- `bind_context op=bind context_id=<own id>` returns success, "Matched by: context_id", "Created tab: no"
- `bind_context op=status` immediately after returns a different tab, the one associated with another session's most recent run
- `op=list` shows the `[bound]` marker on the foreign tab; the intended tab usually still exists
- Re-bind corrects status briefly, then any subsequent call drifts again, sometimes to a third session's tab
- In one occurrence the intended tab was also removed from the tab list entirely (closed by run/session cleanup)

## Reproduction

The bind/status loop alone will NOT reproduce this; plain binding is correctly per-connection. The run-policy traffic is required:

1. One RepoPrompt instance, one workspace, one window.
[...]

PR #372, opened 2026-07-04
Preserve explicit bind_context pins across agent_run start

lemikegao:
## Summary
- Preserve an external connection's explicit `bind_context` pin when `agent_run op=start` (and the shared explore/manage start paths) rebinds the calling connection to the new agent-session tab.
- Add `testAgentRunSpawnRebindPreservesExplicitBindContextPin` (3 scenarios: explicit pin preserved; implicit window-affinity binding still follows the spawn; `.agentModeRun` run-scoped source binding still preserved).
- Add the matching test-suite contract-ledger row.

## Why
An external MCP client (e.g. Claude Code over stdio) that pins its connection with `bind_context op=bind context_id=<own compose tab>` loses that pin as a side effect of `agent_run op=start`: `AgentExternalMCPRunStarter.start` calls `bindCurrentRequestToTab(target.tabID, ...)`, and `bindCurrentRequestToTabIfPossible` had no explicit-binding preservation, so the caller's explicitly chosen tab was displaced to the new agent-session tab. The client's next call carrying its own tab as a per-call hint then fails strict hint validation with `-32602`: "Explicit tab context hint for <tool> targets tab <own tab>, but this connection is already bound to tab <agent tab>". Reproduced deterministically on CE 1.0.22.

The rebind-on-spawn itself is intentional - implicit (window-affinity) callers should follow the session they just started - and a preservation guard already existed for one caller class (`shouldPreserveAgentRunSourceBinding`, `.agentModeRun` sources). This change extends the same seniority to explicit external pins: the guard early-returns (with a debug log) when the connection's existing binding is `explicitlyBound` and its tab differs from the spawn target.

This is the minimal deterministic manifestation of #168 item 2 (explicit `bind_context` bindings are not senior to run-scoped rebinding). The full seniority/provenance overhaul remains #168. Refs #168.

Classic RepoPrompt has the same displacement at its `AgentExternalMCPRunStarter.swift:81`, but it is silent there - Classic has no strict hint validation, so per-call routing continues to work. In CE the strict validation turns the displacement into a hard error.

## Behavior
| Caller's binding at spawn time | Before | After |
|---|---|---|
| Explicit `bind_context` / `_tabID` pin to a different tab | Displaced; next `context_id`-hinted call fails `-32602` | Preserved |
| Implicit window-affinity binding | Follows the spawn | Follows the spawn (unchanged) |
| `.agentModeRun` run-scoped source binding | Preserved | Preserved (unchanged) |

Run control is unaffected: child-session creation and `agent_run` poll/wait/steer resolve by `session_id`, not the caller's binding.

## Known interim limitation
The spawn rebind stamps `explicitlyBound: true`, so after a first spawn rebinds a previously-implicit caller, a second `agent_run start` on that connection preserves the first session's tab instead of following the new spawn. This is a known interim trade-off, not the intended final seniority model: stamping `false` there would break `requireExplicitOrRunScoped` policy resolution for post-spawn calls (the binding has no runID). Proper binding provenance/seniority tracking belongs to the #168 overhaul.

## Validation
- New test fails without the fix at the explicit-pin assertion (guard stashed), passes with it.
- `make dev-test FILTER=TabContextRoutingTests` - 47/47.
- Full `make dev-test` - 3 failures, all pre-existing on clean main (2 flaky pass-on-retry; `GitBlobIdentityServiceTests/testCheckoutAttributesConfigFiltersLFSIdentAndEncodingRequireRawBytes` fails identically on main).
- `make dev-lint` - clean. `verify-ledger` - new ID reconciles; pre-existing main ledger debt untouched.
- Live A/B, same 4-step CLI sequence on one connection (bind explicit pin, `agent_run start`, `bind_context status`, `context_id`-hinted call): 1.0.22 release build shows the displacement and the `-32602` failure; fixed 1.0.23 debug build preserves the pin and the hinted call succeeds.
- `.agents/skills/rpce-contribution-check/scripts/preflight.sh commit` and `push` pass. `pr-ready`: whitespace, secret scans, repository guardrails, SwiftFormat, and SwiftLint strict (0/1384) all pass; the full root-test lane reports only the two pre-existing main failures above (`GitBlobIdentityServiceTests/testCheckoutAttributesConfigFiltersLFSIdentAndEncodingRequireRawBytes` fails identically on clean main; `DurableArtifactIdentityLeaseGCTests/testQuarantineGraceAbandonedWorkCleanupAndDirectOldVersionDeletion` is flaky and passes on focused retry). `TabContextRoutingTests` passed in the same run, including the new test.


provencher (approved):
Looks good put probably too much test code for this

baron (changes requested):
The explicit-binding seniority guard remains directionally sound, but exact head `38b39451e33e9042cf0a8fcc1ca20c29d8208750` is now conflicting with current main. The owning function has moved and the branch carries a ledger row for a ledger system removed from main. Please update onto current main, port only the production guard and focused regression coverage, omit the removed ledger artifact, and obtain fresh exact-head checks.

morluto:
### Audit disposition — **known correctness blocker** (2026-08-14)

This fixes the immediate explicit-pin displacement, but the PR itself documents that a second `agent_run start` can remain bound to the first spawned session because the spawn binding is stamped `explicitlyBound: true`. That is a knowingly incorrect steady state, not merely missing validation.

Please replace the boolean with binding provenance/seniority: user-selected pin, per-call hint, implicit affinity, spawned-run follow, and run-scoped ownership need distinct identities and replacement rules. An operation should only overwrite a binding it owns or one explicitly junior to it. I would not merge another precedence special case while #168’s underlying model remains unresolved.

morluto:

**Disposition: block; the known interim limitation is a correctness bug.** This fixes explicit-pin displacement, but the spawned binding is still stamped `explicitlyBound: true`. A second `agent_run start` can therefore preserve the first spawned tab instead of following the second spawn. That leaves one routing bug in place while fixing another.

Please implement binding provenance/seniority rather than extending the boolean: distinguish user pins, per-call overrides, spawned-run affinity, inherited window affinity, and run-scoped source bindings, with explicit replacement rules. I would not merge a routing change that knowingly directs subsequent starts to the wrong tab; the #168 provenance work should be the prerequisite or be included here.

PR #374, opened 2026-07-05
Add impacted XCTest lane and harden test tooling

morluto:
## Summary

- Add an impacted root XCTest lane backed by the curated test contract ledger.
- Default impacted selection to include both `origin/main...HEAD` and current worktree changes.
- Add optimizer support for impacted selection, shard planning, focused-cost diagnostics, and append-only timing evidence.
- Harden hosted CI XCTest routing so suites run against matching built `.xctest` bundles.
- Consolidate shared XCTest helpers for async waits, git/worktree fixtures, codemap seams, and scale gates.
- Update testing docs and ledger rows for the new workflow.
- Gate opt-in benchmark/scale tests and remove the unused SwiftyJSON dependency.
- Fix follow-up edge cases in live ledger validation, split test-root selection, and codemap seam accounting.

## Why

The full root XCTest suite takes roughly ~10 minutes per run, which makes each iteration cycle expensive. This branch keeps full-suite validation available, but adds a safer targeted inner loop for changes with narrower test impact.

Local timing evidence from this branch shows an approximately 16% reduction in repeated test-cycle cost for the measured workflow.

## Validation

- `python3 Scripts/test_test_suite_optimizer.py` — 44 tests passed
- `python3 Scripts/test_ci_app_test_runner.py` — 28 tests passed
- `make dev-test FILTER=WorkspaceFileContextStoreTests` — 123 tests passed
- `make dev-format-check` — passed
- `.agents/skills/rpce-contribution-check/scripts/preflight.sh commit` — passed

## Notes

- This is test/tooling infrastructure; no production app behavior changes are intended.
- The impacted lane is an iteration aid, not a replacement for full PR-ready validation.
- Local timing artifacts should remain append-only and should only be committed when paired with retained artifacts under `docs/test-suite-optimizer/artifacts/`.


baron:
## Review findings from a local PR worktree

I reviewed this PR in an isolated worktree at PR head `23a8e09c` against `origin/main` (`fd28ee6c`) and also asked a Claude Fable 5 reviewer, bound to the same worktree, for a second opinion. The Python/tooling validations I ran all passed, but I found a few localized correctness issues in the new impacted-test lane that look worth fixing before relying on `make dev-test-impacted` as a gate.

### Validations run locally

All of these passed in the PR worktree:

```bash
python3 Scripts/test_test_suite_optimizer.py          # 44 tests
python3 Scripts/test_ci_app_test_runner.py            # 28 tests
python3 Scripts/test_contribution_preflight.py        # 10 tests
python3 Scripts/test_conductor_lifecycle.py           # 66 tests
python3 Scripts/test_xcode_workspace_generator.py     # 21 tests
python3 -m py_compile Scripts/ci_app_test_runner.py Scripts/conductor.py Scripts/generate_xcode_workspace.py Scripts/test_suite_optimizer.py Scripts/test_ci_app_test_runner.py Scripts/test_conductor_lifecycle.py Scripts/test_contribution_preflight.py Scripts/test_test_suite_optimizer.py Scripts/test_xcode_workspace_generator.py
make guardrails
```

The SwiftyJSON removal, benchmark-test gating, CI XCTest single-bundle routing, conductor `--test-product` plumbing, and guardrails consolidation all looked consistent from this pass.

---

### 1. `impacted --run` on a broad boundary currently runs nothing and exits 0

In `Scripts/test_suite_optimizer.py`, `impacted_tests()` returns early when any changed path matches `BROAD_IMPACT_PATH_PREFIXES`:

```py
if broad_reasons:
    full_count = len(rows)
    return {
        ...
        "selection_mode": "full_root_required",
        "full_root_required": True,
        ...
        "command": [str(repo_root / "conductor"), "test"],
        "run": None,
    }
```

But `Makefile` wires `dev-test-impacted` with `--run` unconditionally:

```make
dev-test-impacted:
	@python3 Scripts/test_suite_optimizer.py impacted --ledger Scripts/Fixtures/test-suite-contract-ledger.tsv --range "$(if $(RANGE),$(RANGE),default)" --run$(if $(INCLUDE_HEAVY), --include-heavy)
```

So for broad-boundary changes, `make dev-test-impacted` can succeed without executing any tests. I confirmed this directly on the PR branch:

```bash
python3 Scripts/test_suite_optimizer.py impacted \
  --ledger Scripts/Fixtures/test-suite-contract-ledger.tsv \
  --range origin/main...HEAD \
  --run \
  --skip-live-list-validation
```

Summary of the JSON result:

```json
{
  "selection_mode": "full_root_required",
  "full_root_required": true,
  "command": [".../conductor", "test"],
  "run": null
}
```

Process exit was `0`.

Suggested fix: if `run_selected` is true and `full_root_required` is true, either actually invoke `run_conductor(repo_root, "root")` and propagate failures, or return nonzero with an explicit message telling the caller to run the full root suite.

---

### 2. Focused `impacted --run` records failing conductor results but still exits 0

For the non-broad path, `impacted_tests()` records the conductor result:

```py
if run_selected and filter_value:
    run = run_conductor(repo_root, "root", filter_value=filter_value)
    run_payload = {
        "command": run.command,
        "process_exit_code": run.process_exit_code,
        "state": run.result.get("state"),
        "exit_code": run.result.get("exitCode"),
        ...
    }
```

But nothing checks `process_exit_code`, `state`, or `exit_code`, and `main()` prints the payload and returns `0` for every successful JSON-producing command:

```py
print(json.dumps(payload, indent=2, sort_keys=True))
return 0
```

That means a failing impacted test run can be represented in JSON but still make `make dev-test-impacted` pass. This undermines the target as a validation gate.

Suggested fix: after a `--run`, raise `OptimizerError` or otherwise return nonzero unless the conductor run has `process_exit_code == 0`, `state == "completed"`, and `exit_code == 0`. Add unit coverage for the failing-run case.

---

### 3. The default impacted range misses staged-only worktree changes

The PR description says the default impacted selection includes branch changes and current worktree changes. The current implementation does:

```py
branch_changed = changed_files_for_git_diff(repo_root, ["origin/main...HEAD"], ...)
worktree_changed = changed_files_for_git_diff(repo_root, [], "worktree")
return sorted(set(branch_changed).union(worktree_changed))
```

`changed_files_for_git_diff(..., [])` runs plain:

```bash
git diff --name-only --diff-filter=ACMRT --
```

That is worktree-vs-index, so staged-but-uncommitted changes are invisible. This is important for the usual flow of staging intended changes and then running validation before commit.

I verified with a temp git repo:

```text
git diff --name-only: []
git diff HEAD --name-only: ['staged.txt']
changed_files_for_range(default): []
```

Suggested fix: use a local leg that includes staged + unstaged changes, e.g. `git diff --name-only --diff-filter=ACMRT HEAD --`, or union explicit staged and unstaged diffs. Consider whether untracked files should also be included, since “current worktree changes” may reasonably imply them.

---

### Lower-priority notes

- `--test-bundle-name` appears to be an escape hatch only; no current CI caller uses it. Its current target-name mismatch check would reject a SwiftPM combined bundle name like `RepoPromptCEPackageTests.xctest`, but auto-discovery covers the live CI path and is tested.
- The full-root payload reports `selected_count` as the total ledger row count while `selected` is `[]`; that is understandable but could be confusing for consumers.
- Large shard/impacted filters generate very large regex CLI args. Current scale is probably fine on macOS, but a guard/comment might help future-proof it.

Bottom line: I did not see problems in the shipped app/build paths, but the three impacted-lane issues above can produce false-green validation results. They should be small, localized fixes in `Scripts/test_suite_optimizer.py` plus tests in `Scripts/test_test_suite_optimizer.py`.

morluto:
All three primary points were accurate, and I've addressed them plus the lower-priority notes.

**1. Broad boundary + `--run` no longer exits 0 silently.** When `run_selected` is true and a broad boundary is detected, `impacted_tests()` now raises `OptimizerError` with a message directing the caller to `make dev-test`, and `main()`'s existing exception branch returns 1. I went with the "raise + tell the caller" variant rather than auto-invoking the full root suite, since silently escalating a focused lane into a full-suite job would defeat the point of `full_root_required` and could surprise callers with a much heavier job than they asked for.

**2. Focused `--run` failures now propagate.** After the focused conductor run, the same triple check used by `listed_root_test_ids` (`process_exit_code == 0`, `state == "completed"`, `exitCode == 0`) is applied; any failure raises `OptimizerError` and `main()` returns 1. This makes `dev-test-impacted` an actual validation gate instead of a green-on-failure reporter.

**3. Default range now captures staged changes.** The worktree leg now runs `git diff --name-only --diff-filter=ACMRT HEAD --` instead of the worktree-vs-index form, so staged-but-uncommitted changes are included. Untracked files are intentionally excluded (documented in the comment): they include scratch artifacts and local investigation docs that will never be committed, and the branch leg already excludes them, so keeping the two legs consistent. Makefile help text updated to "branch, staged, and unstaged changes."

**Lower-priority notes:**
- `selected_count` in the broad payload now reports `0` (consistent with `selected: []` and with the focused path's `len(selected_ids)`), with a separate `full_root_total_count` for the total ledger row count. No existing consumer depended on the old overloaded value.
- Added a comment at `exact_xctest_filter` noting the current scale ceiling and a future guard/chunking path past ~1k selections.
- `--test-bundle-name` left unchanged; agreed it's an escape hatch with no current CI caller.

**Test coverage added:** `test_impacted_tests_broad_boundary_with_run_raises`, `test_impacted_tests_focused_run_failure_raises`, `test_impacted_tests_focused_run_success_attaches_payload` (guards against over-tightening the new check), plus updates to the existing broad-boundary and default-range tests for the new field semantics and `HEAD --` form.

provencher (approved):

provencher:
Fyi @morluto i merged this but didn't realize tests started failing before doing so. Please take a look at stabilizing ci in a follow up pr

morluto:
Sounds good, I will do that now

morluto:
that issue seems to be transient locally, still hunting it down

shipping some test / ci fixes & optimizations + leak fixes in a PR shortly, should help @provencher

morluto:
follow up PR here @baron @provencher 
https://github.com/repoprompt/repoprompt-ce/pull/379

PR #375, opened 2026-07-05
Add keyboard shortcuts for agent model and effort picker

### LINKED ISSUE #363: Keyboard shortcuts for the Agent/Model and Effort pickers
## Problem Statement

When I'm working in Agent Mode, switching the active session's agent and model means reaching for the mouse to click the composer's provider/model chip. There's no keyboard shortcut to open that picker, and the Settings → Keyboard Shortcuts screen only lets me remap existing predefined actions — none of which touch model selection. The same is true for Codex's reasoning-effort control. For a keyboard-driven workflow this is a constant, small friction: I know exactly which model I want, and once the menu is open I can navigate it with the keyboard, but I can't get the menu open without leaving the keyboard.

## Solution

Add a remappable keyboard shortcut that opens the active Agent session's existing Agent/Model picker — the same composer menu I open today by clicking. It ships with a sensible default binding (`⌘⇧M`) and lives in Settings → General → Keyboard Shortcuts under the "Agent & layout" section, so I can rebind it like any other action.

Add a second, optional remappable shortcut that opens the effort picker for the active session. Effort is not tied to one provider: Codex exposes a reasoning-effort control, Claude-tooling agents expose an effort-level control, and other harnesses/models may expose their own. The shortcut opens whichever effort control the currently selected agent exposes. It ships with no default binding (effort is provider/model-specific and less universal than model selection), and appears in Settings right next to the model-picker shortcut so it's discoverable when I want it.

Both shortcuts only _open existing controls_. They don't introduce new model-switching logic, don't add a shortcut per model, and don't change any defaults. Selection, provider-lock handling, validation, and persistence continue to flow through the exact same paths the mouse uses today. The shortcuts are scoped to the focused Agent window, so they act on the session I'm actually looking at.

## User Stories

1. As a keyboard-driven Agent Mode user, I want a shortcut that opens the active session's Agent/Model picker, so that I can switch models without reaching for the mouse.
2. As a user with a specific model in mind, I want the picker to open at the same menu I use today, so that I can navigate it with the keyboard and pick the model I need.
3. As a user, I want the model-picker shortcut to have a sensible default binding, so that it works out of the box without configuration.
4. As a user who dislikes the default, I want to remap the model-picker shortcut in Settings, so that it fits my personal keymap.
5. As a user, I want the model-picker shortcut to appear in the "Agent & layout" section of Keyboard Shortcuts, so that I can discover it alongside related actions.
6. As a user running multiple windows, I want the shortcut to open the picker for the window I'm focused on, so that I never accidentally change a different session's model.
7. As a user in a session where the provider is locked, I want the shortcut to behave exactly like clicking the chip — surfacing the same lock message or disabled rows — so that the shortcut is never a back door around locking rules.
8. As a user, I want a model chosen via the shortcut to persist and save the same way as a mouse-selected model, so that the input method never changes the outcome.
9. As a user, I want the shortcut to do nothing harmless when no agent providers are available or model controls are disabled, so that it never opens an empty or broken menu.
10. As a user, I want the shortcut to never open the sidebar "Models" defaults popover, so that it stays focused on the active session rather than global defaults.
11. As a user who presses the shortcut repeatedly, I want each press to reliably open the picker, so that a second press after dismissing the menu still works.
12. As any Agent Mode user, I want an optional shortcut that opens the effort picker for my active session, so that I can change effort without the mouse regardless of which provider I'm on.
13. As a Codex user, I want the effort shortcut to open the Codex reasoning-effort menu, so that the shortcut respects my provider's effort control.
14. As a Claude-tooling user, I want the same effort shortcut to open the Claude effort-level menu, so that one shortcut works across providers rather than being Codex-only.
15. As a user, I want the effort shortcut to open the same effort menu I use today and persist through the same path, so that behavior is identical to clicking.
16. As a user, I want the effort shortcut to ship unbound by default, so that it doesn't consume a global hotkey unless I choose to assign one.
17. As a user, I want to bind the effort shortcut in Settings next to the model-picker shortcut, so that I can opt in when I want it.
18. As a user whose selected agent exposes no effort control, I want the effort shortcut to do nothing harmless, so that it never opens an irrelevant or empty menu.
19. As a user, I want neither shortcut to fire when a RepoPrompt window isn't focused or keyboard shortcuts are disabled, so that global key presses aren't hijacked.
20. As a user, I don't want a shortcut for every individual model, so that my keymap stays small and the picker remains the single place I make the choice.

## Implementation Decisions

**Single presentation seam.** The design centers on one seam: a token-based programmatic-open request on the existing stable menu component that backs the composer's provider/model chip. Adding an optional "open request" token to that component lets any caller trigger the same presentation path a click uses, without duplicating menu-building, availability, disabled-state, or selection logic. This is the highest-level seam available and keeps the change count near one.

**Shallow shortcut coordinator.** The keyboard-shortcut layer stays thin. Registering the new shortcut(s) resolves the focused window (through the existing focus-gating used by all Agent shortcuts) and posts a window-scoped presentation request. The coordinator reads and mutates no model state.

**Window-scoped presentation notification.** Because the app's notification delivery is broadcast, every picker-open notification carries the target window's identity in its user info, and the mounted composer filters on it. The composer already has its own window identity available, so it only reacts to requests aimed at it. This guarantees multi-window correctness.

**Composer owns applicability.** The mounted composer decides whether to present, matching the exact render/disabled conditions of the visible chip: agent providers must be available and model controls must be enabled. For the effort shortcut, the composer resolves which effort control the selected agent exposes — Codex's reasoning-effort control, a Claude-tooling agent's effort-level control, or a future provider's equivalent — and presents that one, with non-empty effort options; if the selected agent exposes no effort control, the shortcut is a no-op. All lock messages, disabled rows, model-menu snapshotting, and selection actions remain local to the existing picker implementations and are not modified.

**Effort is provider-general, not Codex-specific.** The composer already renders separate effort controls per provider family (a Codex reasoning-effort menu and a Claude effort-level menu today), each with its own options source and selection action. The effort shortcut is a single command that routes to whichever of these applies to the active agent, rather than a Codex-only shortcut. New providers that add their own effort control participate by being resolved in the same applicability step.

**Distinct tokens per press.** Repeated shortcut presses assign a fresh unique token each time, so the "open request" is not deduplicated by the UI framework and a second press reliably re-opens the menu. Programmatic presentation is deferred off the current UI update pass to avoid presenting the native menu inside a view-update transaction. A request that arrives before the anchor exists safely does nothing.

**Modules built/modified.**

- Global keyboard-shortcut identifiers: add `openAgentModelPicker` (default `⌘⇧M`) and, for the follow-up, `openAgentEffortPicker` (no default). `⌘⌥M` is deliberately avoided because macOS commonly reserves it for Minimize All.
- Keyboard-shortcut coordinator: register the new shortcut(s) and post the window-scoped presentation notification(s) via the existing focus-gated helper.
- App notification names: add window-scoped `openAgentProviderModelPicker` and (follow-up) `openAgentEffortPicker`, each documenting a window-id value in user info.
- Stable menu component: add an optional open-request token input and present via the existing click path when it changes.
- Keyboard Shortcuts settings: add one remappable row for the model picker under "Agent & layout"; add a second row for the effort picker only when that follow-up is implemented.
- Agent composer view: receive the notification, filter by window id, gate on the chip's own render/disabled conditions, and pass a fresh token to the stable menu.

## Testing Decisions

**Test external behavior, not internals.** Good tests here assert on observable outcomes of the seam — "given a focused window and an available picker, an open request results in a presentation request the composer acts on" — rather than on private view state or menu internals.

**What to cover.**

- The shortcut coordinator posts a window-scoped presentation request only when a RepoPrompt window is focused and keyboard shortcuts are enabled, and includes the correct window identity.
- The composer's applicability gate: a request is honored only when providers are available and model controls are enabled; ignored otherwise; and, for the effort shortcut, it resolves to the effort control matching the selected agent (Codex reasoning effort, Claude effort level, or another provider's), presenting only when that control has non-empty options and is a no-op when the agent exposes none.
- Window scoping: a request targeted at one window's identity is ignored by a composer bound to a different window.
- Repeated requests each produce a distinct open token (no dedup), so a second press re-opens.
- Selection through the shortcut-opened menu exercises the same selection/persistence path as a click (no divergent branch).

**Prior art.** Follow the existing keyboard-shortcut registration/focus-gating tests and existing Agent composer/view-model tests as the pattern for these. Reuse existing model-selection test scaffolding rather than introducing a parallel harness.

## Out of Scope

- A separate shortcut for each individual model or provider (explicitly rejected — the picker remains the single selection surface).
- "Cycle model", "cycle provider", "cycle effort", or "use last model" power-user commands.
- Changing global defaults: sub-agent role defaults, Oracle/Plan model, or Context Builder agent. The sidebar "Models" popover remains the home for defaults and is untouched.
- Any change to model availability logic, provider-switch locking rules, validation, or persistence.
- New selection UI. Both shortcuts only open existing controls.

## Further Notes

- Effort is provider-general: it is not limited to Codex. Today the composer exposes a Codex reasoning-effort control and a Claude effort-level control, and other harnesses/models may add their own. The single effort shortcut resolves to whichever applies to the active agent.
- The effort follow-up involves converting the affected provider effort controls from framework menus to the shared stable menu component. A small, acceptable visual shift results: menu selection indication moves from framework-drawn checkmarks to native menu-item selection state. This is intended and part of that slice.
- Default binding rationale: `⌘⇧M` is proposed for the model picker; `⌘⌥M` is avoided due to the common macOS Minimize All reservation. The effort shortcut intentionally ships unbound to avoid claiming a second global hotkey by default.

aAtila:
The agent composer's model picker and provider effort picker can now be opened from remappable keyboard shortcuts, routed to the currently focused composer.

## Changes

- Add remappable global shortcuts that open the active composer's model picker and effort picker.
- Route shortcut requests through the focused window to the active composer, with teardown-safe menu presentation guards so a menu never opens on a composer that's tearing down.
- Snapshot Codex effort options into composer props so the shortcut-opened effort menu has stable state.
- Align Codex and Claude effort picker trigger styling with the model picker.
- Cover shortcut catalog registration and composer open-request filtering with tests.

The model picker keeps its default binding (Cmd+Shift+M) and the effort picker ships unbound; both remain remappable and scoped to the active composer, so triggering a picker never fires a menu on the wrong window. Provider effort controls now follow the same focused-composer shortcut path as the model picker.

## Testing

All green via the coordinated daemon:

- `make guardrails` — passed (SwiftPM notice inventory matches Package.resolved, 50 packages).
- `make dev-lint` — passed (0 violations, 0 formatting diffs across 1386 files).
- `make dev-test` for the related suites — **55 tests, 0 failures**:
  - `AgentModelPickerOpenRequestGuardTests` (7)
  - `AgentEffortPickerOpenRequestGuardTests` (13)
  - `KeyboardShortcutCatalogTests` (2)
  - `AgentModeChatSwitchActivationTests` (4)
  - `AgentModeStopSubmitTargetTests` (24)
  - `WorktreeMergeReviewStateTests` (5)

Fixes #363

### DIFF
--- a/Sources/RepoPrompt/App/GlobalKeyboardShortcutsCoordinator.swift
+++ b/Sources/RepoPrompt/App/GlobalKeyboardShortcutsCoordinator.swift
@@ -218,6 +218,7 @@ final class GlobalKeyboardShortcutsCoordinator {
         register(.nextParentAgentSession) { [weak self] in self?.focusAdjacentParentAgentSession(forward: true) }
         register(.showCurrentWindowAgentNavigationHUD) { [weak self] in self?.showAgentNavigationHUD(mode: .currentWindow) }
         register(.showAllAgentsNavigationHUD) { [weak self] in self?.showAgentNavigationHUD(mode: .allAgents) }
+        register(.showAgentModelPicker) { [weak self] in self?.showAgentModelPicker() }
     }
 
     private func startNewAgentSessionFromShortcut() {
@@ -272,6 +273,15 @@ final class GlobalKeyboardShortcutsCoordinator {
                 AgentNavigationHUDNotificationUserInfoKey.windowID: win.windowID,
                 AgentNavigationHUDNotificationUserInfoKey.mode: mode.rawValue
             ]
+        )
+    }
+
+    private func showAgentModelPicker() {
+        guard let win = guardedFocusedWindowState() else { return }
+        NotificationCenter.default.post(
+            name: .showAgentModelPicker,
+            object: nil,
+            userInfo: ["windowID": win.windowID]
         )
     }
 }
--- a/Sources/RepoPrompt/App/GlobalKeyboardShortcutsCoordinator.swift
+++ b/Sources/RepoPrompt/App/GlobalKeyboardShortcutsCoordinator.swift
@@ -219,6 +219,7 @@ final class GlobalKeyboardShortcutsCoordinator {
         register(.showCurrentWindowAgentNavigationHUD) { [weak self] in self?.showAgentNavigationHUD(mode: .currentWindow) }
         register(.showAllAgentsNavigationHUD) { [weak self] in self?.showAgentNavigationHUD(mode: .allAgents) }
         register(.showAgentModelPicker) { [weak self] in self?.showAgentModelPicker() }
+        register(.showAgentEffortPicker) { [weak self] in self?.showAgentEffortPicker() }
     }
 
     private func startNewAgentSessionFromShortcut() {
@@ -280,6 +281,15 @@ final class GlobalKeyboardShortcutsCoordinator {
         guard let win = guardedFocusedWindowState() else { return }
         NotificationCenter.default.post(
             name: .showAgentModelPicker,
+            object: nil,
+            userInfo: ["windowID": win.windowID]
+        )
+    }
+
+    private func showAgentEffortPicker() {
+        guard let win = guardedFocusedWindowState() else { return }
+        NotificationCenter.default.post(
+            name: .showAgentEffortPicker,
             object: nil,
             userInfo: ["windowID": win.windowID]
         )
--- a/Sources/RepoPrompt/App/Notifications/AppNotifications.swift
+++ b/Sources/RepoPrompt/App/Notifications/AppNotifications.swift
@@ -95,4 +95,8 @@ extension Notification.Name {
     /// ⌘1–⌘9 handlers before they fall through to regular compose-tab switching.
     /// userInfo: ["windowID": Int, "resultIndex": Int, "handledRequest": AgentNavigationHUDHandledRequest]
     static let selectAgentNavigationHUDResult = Notification.Name("selectAgentNavigationHUDResult")
+
+    /// Request that Agent Mode open the Agent/Model picker for the target window.
+    /// userInfo: ["windowID": Int]
+    static let showAgentModelPicker = Notification.Name("showAgentModelPicker")
 }
--- a/Sources/RepoPrompt/App/Notifications/AppNotifications.swift
+++ b/Sources/RepoPrompt/App/Notifications/AppNotifications.swift
@@ -99,4 +99,8 @@ extension Notification.Name {
     /// Request that Agent Mode open the Agent/Model picker for the target window.
     /// userInfo: ["windowID": Int]
     static let showAgentModelPicker = Notification.Name("showAgentModelPicker")
+
+    /// Request that Agent Mode open the active provider's effort picker for the target window.
+    /// userInfo: ["windowID": Int]
+    static let showAgentEffortPicker = Notification.Name("showAgentEffortPicker")
 }
--- a/Sources/RepoPrompt/Features/AgentMode/ViewModels/AgentModeViewModel+ComposerUI.swift
+++ b/Sources/RepoPrompt/Features/AgentMode/ViewModels/AgentModeViewModel+ComposerUI.swift
@@ -44,6 +44,7 @@ extension AgentModeViewModel {
             selectedModelDisplayName: selectedModelDisplayName,
             selectedReasoningEffortRaw: selectedReasoningEffortRaw,
             selectedReasoningEffortDisplayName: selectedReasoningEffortDisplayName,
+            codexReasoningEffortOptions: reasoningEffortOptionsForCurrentSelection(),
             availableAgents: availableAgents,
             isProviderPickerLockedForCurrentTab: isProviderPickerLocked(tabID: tabID),
             lockedAgentSelectionMessage: lockedAgentSelectionMessage(tabID: tabID),
--- a/Sources/RepoPrompt/Features/AgentMode/ViewModels/UI/AgentComposerUIModels.swift
+++ b/Sources/RepoPrompt/Features/AgentMode/ViewModels/UI/AgentComposerUIModels.swift
@@ -234,6 +234,7 @@ struct AgentComposerProps: Equatable {
     let selectedModelDisplayName: String
     let selectedReasoningEffortRaw: String?
     let selectedReasoningEffortDisplayName: String
+    let codexReasoningEffortOptions: [CodexReasoningEffort]
     let availableAgents: [AgentProviderKind]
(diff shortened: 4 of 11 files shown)


baron (approved):
Approved on exact head `d0167db9d855da950e5fa26dc9e10d37e66e276f`. The shortcut routing is guarded by the target window/tab and the detached menu presentation has a stable ownership guard; I found no code blocker. App shard 4 still timed out in unrelated `WorktreeAPISmokeHarnessTests`, so rerun the required check successfully on current main before merge—no retries or timeout inflation.

aAtila:
Thanks, Baron. Since your review of `d0167db9`, I ran an additional code-review pass with PRO (web) and addressed its findings:

- deferred picker presentation now revalidates composer ownership/lifecycle before opening;
- picker menus now use fresh state instead of state captured before presentation;
- focused tests cover these stale-state and teardown paths.

Regarding the unrelated `WorktreeAPISmokeHarnessTests` timeout: after several CI reruns, the shard finally passed unchanged. The same intermittent timeout also appears on unrelated `main` runs, while three focused local runs passed. I added no retry logic or timeout inflation.

All required checks are now green on the current head (`2f796567`). Since the reviewed head has advanced, another look would be appreciated.

morluto:
### Audit disposition — **merge candidate after final current-base checks** (2026-08-14)

I did not find a code-level blocker in the focused-window routing, active-composer filtering, teardown-safe menu presentation, or stable effort-option snapshot. The design is appropriately scoped to the focused composer rather than a global menu owner.

Final gates: refresh onto current main, obtain exact-head CI, and manually exercise two windows, rapid focus transfer, disabled/unavailable effort controls, composer teardown, and remapped/unbound shortcuts. Subject to those checks, this looks suitable to merge.

morluto:

**Disposition: merge candidate after current-head gates.** The focused-composer routing, window/tab identity checks, teardown-safe presentation guards, and stable effort-option snapshot form a coherent contract. I did not find a code-level blocker.

Final validation should include a manual multi-window smoke: change focus between composers, invoke both shortcuts during/after tab teardown, verify disabled or unavailable effort controls do not open stale menus, and confirm remapping/unbound behavior. With exact-head CI green, this is appropriately narrow and ready for non-author approval.

PR #376, opened 2026-07-05
Coalesce Oracle token estimate updates

baron:
## Summary
- Coalesce Oracle upcoming-input token estimate work instead of canceling/restarting every update.
- Snapshot the latest compose/conversation state after a short coalescing delay.
- Use the prompt view model's cached token total so the passive "Next msg" estimate no longer rebuilds full chat context/clipboard during typing or restore bursts.

## Notes
- This intentionally trades some estimate precision for responsiveness when the chat preset's resolved context differs from the composer cached total, especially git-inclusion cases.
- No tests were added: Fable 5 review called out that testing this private UI-label coalescing path would require disproportionate refactoring.

## Validation
- `swiftformat --config .swiftformat Sources/RepoPrompt/Features/Chat/ViewModels/Oracle/OracleViewModel.swift`
- `swiftformat --config .swiftformat --lint Sources/RepoPrompt/Features/Chat/ViewModels/Oracle/OracleViewModel.swift`
- `swiftlint lint --strict --config .swiftlint.yml Sources/RepoPrompt/Features/Chat/ViewModels/Oracle/OracleViewModel.swift`

## Review
- Claude Code Fable 5 (`claudeCode:claude-fable-5`) performed a final read-only minimization review and recommended GO after trimming redundant defensive cancellation machinery.

### DIFF
--- a/Sources/RepoPrompt/Features/Chat/ViewModels/Oracle/OracleViewModel.swift
+++ b/Sources/RepoPrompt/Features/Chat/ViewModels/Oracle/OracleViewModel.swift
@@ -499,34 +499,41 @@ class OracleViewModel: ObservableObject {
 
     @MainActor
     private func updateUpcomingInputTokenText() {
-        // Cancel any previous run
-        upcomingTokenEstimateTask?.cancel()
+        if upcomingTokenEstimateTask != nil {
+            upcomingTokenEstimatePending = true
+            return
+        }
 
-        // Quick snapshot
-        let inputTextSnap = inputText
-        let promptTextTokensSnap = TokenCalculationService.estimateTokens(for: promptViewModel.promptText)
-        let duplicateFactorSnap = promptViewModel.duplicateUserInstructionsAtTop ? 2 : 1
+        let task = Task.detached { [weak self] in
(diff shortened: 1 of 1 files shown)


provencher:
Actually oracle token counting isnt really used anywhere. We should just remove this altogether. it made sense in the IDE mode chat in classic to show the token count there. Here we dont really need this.

baron:
Closing this because the desired product direction changed, rather than because the coalescing implementation is mechanically unsound.

The maintainer feedback on this PR is explicit: the Oracle upcoming-message token estimate is not useful in the current CE chat experience and should be removed rather than optimized. Current main still contains the estimate machinery, but `upcomingInputTokenText` and `updateUpcomingInputTokenText()` have no consumers outside `OracleViewModel.swift`, which supports treating this as dead UI plumbing.

Keeping this optimization PR open would invest further complexity in machinery that is intended to disappear. The cleaner follow-up is a small removal PR against current main that deletes the published property, update task/state, subscriptions, and call sites, with focused Oracle/Chat validation.

Thanks for investigating and improving the update behavior; the implementation work helped clarify the underlying cost and the preferable removal direction.

PR #249, opened 2026-06-18
Make Finder debug launcher just work without a signing identity

morluto:
## Issue

Double-clicking `Launch RepoPrompt CE.command` on a Mac without an Apple Development signing identity drops into a `r/s/x/q` retry loop that never succeeds. The launcher looks like a simple app launcher, but it is really a small retry UI around a build+package+relaunch pipeline:

```text
Finder double-click
-> Launch RepoPrompt CE.command
-> ./conductor app relaunch
-> Scripts/run.sh
-> Scripts/package_app.sh debug
-> swift build RepoPrompt
-> swift build repoprompt-mcp
-> stop/open the app only after packaging succeeds
```

Because packaging runs before the app lifecycle step, any deterministic packaging failure returns the launcher to the same retry prompt. Pressing `r` reruns the same environment and hits the same failure again. There is no hint that the user is expected to set an environment variable in a separate Terminal.

## What is confusing

The launcher presents itself as a one-click "build and relaunch" button, so a retry loop that never recovers reads as a broken app, not as a missing signing identity. The actual error (`Debug ad-hoc signing is disabled by default. Set ALLOW_ADHOC_SIGNING=1 ...`) is buried in the build log and the launcher never surfaces it as actionable guidance. The README documents the `ALLOW_ADHOC_SIGNING=1 ./conductor app relaunch` workaround, but a Finder-double-click user has no reason to read the README before clicking the launcher.

There were three layered blockers in that one pre-launch path, each masking the next:

1. On Macs without an Apple Development signing identity, debug packaging refused ad-hoc signing unless `ALLOW_ADHOC_SIGNING=1` was set. The Finder launcher did not set that, so every retry failed at signing preflight.
2. Once ad-hoc signing was enabled, packaging reached SwiftPM and failed compiling `KeyboardShortcuts` because its design-time `#Preview` blocks required preview macro support that the command-line build does not load.
3. After that, first-party SwiftUI `@Entry` environment values hit the same class of command-line macro-plugin loading failure.

So even a user who found the `ALLOW_ADHOC_SIGNING=1` workaround in the README would still hit two more build failures before the app opened.

## Fix: the launcher just works

`Launch RepoPrompt CE.command` now detects the no-identity case up front and exports `ALLOW_ADHOC_SIGNING=1` for its own conductor relaunch jobs, with a clear warning that debug secure storage will be in-memory. The double-click flow no longer requires the user to open Terminal or read the README first — the launcher just works.

[...]

PR #414, opened 2026-07-07
fix(doctor): detect missing Swift macro plugins on CLT-only hosts (#174)

### LINKED ISSUE #174: CLT-only builds fail despite doctor.sh passing — Xcode-only macro plugins
On a CLT-only host (no Xcode.app), `Scripts/doctor.sh --quiet` passes but `swift build -c release --product RepoPrompt` cannot succeed. Three distinct failures observed on macOS 26.5, CLT 26.5 (`swift-driver 1.148.6`, Swift 6.3.2), repo @ 6f8fcbb:

**1. PackageDescription module mismatch (environment-specific; fixed by CLT reinstall — noted for completeness).** `swift package dump-package` failed on any tools-version-6.x manifest: the CLT `.swiftinterface` advertises the `swiftLanguageModes:` init under `@available(_PackageDescription 6)`, but `libPackageDescription.dylib` only exports the legacy `swiftLanguageVersions:` symbol — compile error when the manifest uses the new label, link error when it doesn't. A clean CLT reinstall resolved it (likely a mismatched mid-cycle update). Relevant here because `doctor.sh` passed throughout: it checks tool presence, not whether the toolchain can actually compile a manifest.

**2. `PreviewsMacros` plugin missing from CLT.** KeyboardShortcuts 2.3.0 (`Recorder.swift:172,177,182`) has three `#Preview` blocks. CLT ships only `libObservationMacros.dylib` and `libSwiftMacros.dylib` under `/Library/Developer/CommandLineTools/usr/lib/swift/host/plugins/`; `PreviewsMacros` ships with Xcode.app only. Repro on a CLT-only host:

```
swift build -c release --product RepoPrompt
# error: external macro implementation type 'PreviewsMacros.SwiftUIView' could not be found
#        for macro 'Preview(_:body:)'; plugin for module 'PreviewsMacros' not found
```

KeyboardShortcuts `main` still carries these blocks, so a dependency bump would not fix this.

**3. `SwiftUIMacros` plugin missing from CLT.** With (2) worked around, the build fails again on CE's own source: `@Entry` is used at 8 sites across 4 files (`App/FontPreset.swift`, `Features/AgentMode/Views/AgentMessageBubble.swift`, `Features/AgentMode/Views/ToolCards/ToolCardContainer.swift`, `Infrastructure/UI/Markdown/MarkdownFileLinkInteraction.swift`):

```
grep -rn '@Entry' Sources/RepoPrompt | wc -l   # 8
# error: external macro implementation type 'SwiftUIMacros.EntryMacro' could not be found
#        for macro 'Entry()'; plugin for module 'SwiftUIMacros' not found
```

The pattern: CLT alone is structurally insufficient for this codebase, and `doctor.sh` currently can't tell a contributor that.

CI doesn't surface any of this because GitHub's macos-26 runners ship Xcode.app, so the CLT path is never exercised.

Two suggestions, either of which would close the trap:
- `doctor.sh` gains a macro-compile probe (and ideally a tiny `dump-package` probe for (1)) — compile a snippet using `#Preview`/`@Entry` and hard-fail with "install Xcode.app" guidance when the plugins are absent;
- or the docs state plainly that Xcode.app is required to build.

I have a working fix for blocker (2) following the existing `Scripts/patches` mechanism (same 045cf17 pin, idempotency marker, mirrored release-tooling harness coverage) — happy to PR it if a CLT path is in scope. It deliberately does not claim to restore CLT builds on its own, since (3) remains.

pnascimento9596:
## Summary
`Scripts/doctor.sh --quiet` passes on Command Line Tools–only macOS hosts, but the build then fails at macro expansion. The check verified tool presence and a SwiftUI *symbol* typecheck (the Liquid Glass probe) but never exercised macro-plugin availability, so it reported success on a host that cannot build the app.

## Root cause
RepoPrompt CE requires two Swift macro plugins that ship only with full `Xcode.app`, never with Command Line Tools:
- `SwiftUIMacros` — CE's own source uses `@Entry` at 10 sites across 5 files.
- `PreviewsMacros` — the pinned `KeyboardShortcuts` 2.3.0 (`045cf174…`) uses `#Preview` in `Recorder.swift`.

CLT alone is structurally insufficient, and the old `doctor.sh` could not tell a contributor that.

## Fix
- **`Scripts/doctor.sh`**: adds a macro-plugin probe (mirrors the existing Liquid Glass probe) that typechecks a snippet using both `@Entry` and `#Preview`. Because macro expansion runs during type-checking, an absent plugin fails the probe, which then hard-fails with guidance to install Xcode 26. Catches both blockers — and any future macro dependency — in one place.
- **`README.md`**: corrects the source-build requirements, which incorrectly listed Command Line Tools as a supported alternative to Xcode.

## Validation
- **Xcode 26 host**: standalone probe typechecks clean (exit 0, no extra imports needed); full `doctor.sh` exits 0 with `OK: toolchain provides Swift macro plugins (@Entry, #Preview)`.
- **CLT-only host** (`DEVELOPER_DIR=/Library/Developer/CommandLineTools`): probe fails with `external macro implementation type 'SwiftUIMacros.EntryMacro' … plugin for module 'SwiftUIMacros' not found` and the equivalent for `PreviewsMacros`; `doctor.sh --quiet` exits non-zero with the install-Xcode guidance. Notably the Liquid Glass probe still passes under CLT — i.e. the old check would have reported success — confirming this reproduces and closes #174.
- `bash -n Scripts/doctor.sh` passes; `python3 Scripts/test_release_tooling.py` → 52 tests, all pass.

## Scope
- Blocker #1 from the issue (mid-cycle `PackageDescription` mismatch) is environment-specific and self-resolves via CLT reinstall; intentionally not probed to avoid false positives.
- This does not attempt to *restore* CLT-only builds. Blocker #3 is first-party `@Entry` usage; making CLT builds succeed would mean removing an idiomatic SwiftUI macro codebase-wide and guarding it in CI — a poor trade for a native macOS app. The honest fix is to detect the unsupported host and say so.

## Known limitation / follow-up
CI does not exercise the CLT-only path (GitHub `macos-26` runners ship Xcode), so this probe is validated by contributors, not CI. A job that selects CLT and asserts `doctor.sh` *fails* would close that gap but is awkward on hosted runners; proposed as a separate change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,11 @@ another Mac or redistributed.
 ### Source-build requirements
 
 - macOS 26 or later
-- Xcode 26, or matching Command Line Tools with the macOS 26 SDK
+- Xcode 26 — full `Xcode.app`, not Command Line Tools alone. RepoPrompt CE and
+  its dependencies use the `@Entry` and `#Preview` Swift macros, whose plugins
+  (`SwiftUIMacros`, `PreviewsMacros`) ship only with `Xcode.app`. Its bundled
+  macOS 26 SDK satisfies the SDK requirement. Run `./Scripts/doctor.sh` to
+  verify your toolchain before building.
 
 ### Develop in Xcode
 
--- a/Scripts/doctor.sh
+++ b/Scripts/doctor.sh
@@ -104,6 +104,28 @@ SWIFT
 ARCH="$(uname -m)"
 run xcrun swiftc -typecheck -parse-as-library -target "${ARCH}-apple-macos14.0" "$TMP/GlassProbe.swift"
 log "OK: toolchain can compile SwiftUI Liquid Glass symbols."
+phase "Compiling Swift macro plugin probe"
+cat > "$TMP/MacroProbe.swift" <<'SWIFT'
+import SwiftUI
+
+// Mirrors real RepoPrompt CE usage: `@Entry` (SwiftUIMacros) in an
+// EnvironmentValues extension, plus `#Preview` (PreviewsMacros) carried by
+// dependencies such as KeyboardShortcuts. Both macro plugins ship only with
+// full Xcode.app; Command Line Tools alone cannot expand them, so a typecheck
+// (which runs macro expansion) is enough to detect their absence.
+extension EnvironmentValues {
+    @Entry var repoPromptDoctorMacroProbe: Bool = false
+}
+
+#Preview("doctor macro probe") {
+    Text("OK")
+}
+SWIFT
+if ! xcrun swiftc -typecheck -parse-as-library -target "${ARCH}-apple-macos14.0" "$TMP/MacroProbe.swift" 2>"$TMP/MacroProbe.err"; then
+    (( quiet )) || cat "$TMP/MacroProbe.err" >&2
+    fail "Swift macro plugins unavailable (SwiftUIMacros / PreviewsMacros). RepoPrompt CE uses the @Entry and #Preview macros, whose plugins ship only with full Xcode.app — Command Line Tools alone cannot build it. Install Xcode 26 and select it: sudo xcode-select -s /Applications/Xcode.app"
+fi
+log "OK: toolchain provides Swift macro plugins (@Entry, #Preview)."
 if (( install_debug_cli )); then
     phase "Installing debug CLI"
     run "$ROOT_DIR/Scripts/install_debug_cli.sh" install --build


davidgovea (approved):
Also ran into this. Had xcode installed but needed to select. 

Ran the debug launcher on your branch & got the fast-fail and helpful error 👍

baron:
Current-main refresh on exact head `df25992a12417c0ea7036b32027a79ad6b52f085`: the two-file doctor macro probe remains useful and the existing exact-head approval should be retained, but the branch now conflicts and is far behind. Please refresh the patch and rerun doctor/release-tooling CI; no duplicate approval is needed until the new head exists.

morluto:
### Audit disposition — **preferred direction; rebase and finish validation** (2026-08-14)

This is the cleaner architectural answer to CLT-only build failures: exercise actual macro expansion in `doctor` and state honestly that full Xcode is required, rather than maintaining source/dependency macro-removal patches.

Please rebase onto current main and rerun both sides of the contract: a full-Xcode host must pass, while `DEVELOPER_DIR=/Library/Developer/CommandLineTools` must fail early with the intended guidance. Recheck README/release tooling as well. This should supersede the macro-workaround portion of #249.

morluto:

**Disposition: preferred direction; refresh and merge after current checks.** This is cleaner than maintaining source/dependency patches for a CLT-only build: the repository actually requires macro plugins, so `doctor` should exercise macro expansion and report the unsupported host honestly. The probe covers the real failure mode that the existing symbol/tool checks miss.

Please rebase, rerun both full-Xcode success and CLT-selected expected-failure paths, and ensure the diagnostic distinguishes missing Xcode from a broken/mis-selected `DEVELOPER_DIR`. With those current-head checks, I found no code-level blocker. This should own the toolchain policy, and the macro-workaround portion of #249 should not be merged alongside it.

PR #418, opened 2026-07-08
fix(settings): persist global defaults across restarts and propagate across windows

### LINKED ISSUE #417: Global settings silently reset on restart; window-local changes don't propagate across windows
**Symptom**

Global user settings revert to defaults every time RPCE restarts, and changes made in one window don't appear in other open windows. Affected settings include:

- Oracle / planning model
- Context Builder agent selection
- Sub-agent role overrides (explore / engineer / pair / design)

Setting one of these in window A leaves other open windows showing the old value until the app is restarted (e.g. A shows Oracle = `fable` while every other window keeps `gpt-5.5`).

**Root cause**

1. **An unsupported schema is indistinguishable from "no settings."** A `globalSettings.json` whose `schemaVersion` is newer than the running build supports decodes the same as a missing file, so the store silently falls back to in-memory defaults on every launch. Settings appear to never persist, with no signal to the user that the file is being ignored.
2. **No live cross-window re-sync.** Each window caches global settings locally and the shared `GlobalSettingsStore` doesn't publish changes to observing windows, so a change in one window only appears in the others after a restart.

**Proposed fix**

- Detect an on-disk schema newer than supported and **surface it to the user**: a banner explaining settings won't save, with one-click recovery that backs up the offending file and restores fresh current-schema defaults. The app must never auto-overwrite a schema it did not write.
- Publish `objectWillChange` on the global-defaults setters so windows observing the shared store re-sync live.

**Status**

Fix implemented with tests (cross-window propagation + JSON-only persistence / schema-block recovery), green locally. PR to follow and will reference this issue.

moonray:
Closes #417.

## Problem
Global user settings — Oracle/planning model, Context Builder agent selection, and sub-agent role overrides — reverted to defaults on every restart, and a change made in one window didn't propagate to other open windows.

Two root causes:
1. A `globalSettings.json` whose schema is newer than the running build supports decoded the same as a missing file, so every launch silently fell back to in-memory defaults — with no signal to the user that their file was being ignored.
2. Each window cached global settings locally and the shared `GlobalSettingsStore` didn't publish changes to observing windows, so cross-window changes only appeared after a restart.

## Fix
- **Surface unsupported schemas:** `GlobalSettingsFileStore` detects an on-disk schema newer than supported and exposes a `persistenceBlockReason`; the file is preserved and saves stay blocked (the app never auto-overwrites a schema it did not write). `GlobalSettingsPersistenceBlockBanner` shows the user why settings won't save, with a one-click recovery that backs up the offending file and restores fresh current-schema defaults.
- **Live cross-window propagation:** global-defaults setters publish `objectWillChange` so windows observing the shared store re-sync immediately.

## Tests (all green locally, 22 tests)
- `GlobalSettingsCrossWindowPropagationTests` — Oracle change propagates across windows live; cross-window re-sync does not feedback-loop.
- `SettingsJSONOnlyPersistenceTests` — global-defaults setters publish; newer-schema file exposes the block reason; newer-schema file is preserved with saves blocked; user-initiated recovery backs up, resets, and unblocks.


baron:
Maintainer note after looking at the issue comments and the live failure mode:

Please let us handle the settings-file/schema recovery path in this PR rather than layering additional migration scripts or ad-hoc JSON edits on top of it. The apparent "settings don't save" symptom is broader than the original PR title made it sound, and the safe fix is a little more defensive than a simple schema-version bump.

What is going on:

- The affected settings are mostly stored in one shared file: `~/Library/Application Support/RepoPrompt CE/Settings/globalSettings.json`.
- If that file cannot be trusted/decoded, the app falls back to in-memory defaults. That makes many unrelated settings look like they never save: Oracle/planning model, Context Builder agent/model, Agent Models/role defaults, context budget, appearance, MCP settings, etc.
- The tricky part is that there are real user/dev environments with **unlineaged** `globalSettings.json` files at schema v3/v4. Those can come from classic/internal builds, manual migration, or dev builds that share the live CE Application Support folder.
- Therefore `schemaVersion` alone is not a safe identity. An unlineaged `schemaVersion: 4` must not be treated as "future CE v4" forever, and it also must not become silently accepted later when CE eventually reaches numeric v4 for real.

The latest maintainer commit (`c87b54e9`) handles this by splitting schema identity into two axes:

- `schemaLineage` = who wrote this settings family.
- `schemaVersion` = compatibility within that lineage.

The important safety rules are:

- Newly written CE files get `schemaLineage: "repoprompt-ce.global-settings"`.
- Unlineaged v1/v2 files are accepted as legacy CE.
- Unlineaged versions above v2 are treated as incompatible/foreign **permanently**, even after CE catches up numerically.
- Same-lineage future CE files are preserved and blocked instead of being down-converted by older builds.
- Valid JSON that we cannot identify is preserved; we do not auto-overwrite it.
- Recovery/import is user-initiated and backs up the existing file first.
- Saves re-check the on-disk header so a newer/foreign file that appears after launch is not clobbered.

That is why the PR now has a banner/recovery path instead of just silently rewriting the file. It is deliberately conservative: the worst outcome should be "settings are blocked until the user chooses import/reset," not "a dev or downgrade build quietly destroys a live settings file."

A couple of asks while this settles:

1. Please avoid manual edits, force-sync scripts, or extra schema migrations against `globalSettings.json` while this PR is being validated. Those make it much harder to reason about whether recovery/import is behaving correctly.
2. If you are testing this after the fix lands and see the banner, please use the banner action rather than hand-editing the JSON. Then report any setting that still does not persist **after** that one-time recovery/import step.
3. The reported "Tools" toggles may be separate from `globalSettings.json` depending on which controls are meant. If those still fail after global settings recovery, we should track that as a separate focused bug with repro steps rather than expanding this PR blindly.

The goal here is not to win an argument about schema nomenclature; it is to prevent data loss across new installs, manually migrated users, classic/internal settings files, dev builds sharing live folders, and future upgrade/downgrade loops. This PR now has tests and docs specifically guarding against the fake-v4/real-future-v4 trap so future contributors do not simplify it back into a clobber bug.

Issue #419, opened 2026-07-08
Multi-window load starves foreground git work — workspace switch and agent worktree dispatches stall


moonray:
## Impact & scope

With many workspace windows open (observed at ~6), interactive foreground work across the app stalls:

- **Workspace switch** can hang indefinitely on the "Switching workspace…" overlay; the overlay's hide path is never reached because the switch's root-load phase is parked behind process-wide concurrency pools that the other windows have saturated.
- **Agent Mode dispatches** that create a worktree (`agent_run` start with `worktree_create`) stall during worktree preparation; the dispatch hangs and the session never launches.

In both cases the main run loop stays healthy (the app is not deadlocked and remains interactive); the foreground `Task` is genuinely suspended at an `await`.

Root mechanism: switch-critical and agent-start git work both run at the foreground admission tier, but the process-wide git admission pool has no reserved lane for foreground work, so sustained background codemap/metadata git demand from multiple windows can keep all slots busy and starve foreground waiters.

## Expected vs actual

- **Expected:** switching workspaces completes promptly (overlay hides within ~1–2s) and agent worktree dispatches start promptly, even with many windows open.
- **Actual:** at ~6 open windows, a switch sits on the overlay for tens of seconds to effectively forever, and agent worktree dispatches stall in preparation. CPU stays partly busy (background work) while the foreground work waits.

## Reproduction

Workspace switch:
1. Open N workspace windows (N ≈ 6), including at least one large repo (hundreds of MB, thousands of filesystem entries, git submodules).
2. Switch into the large workspace from a window.
3. The "Switching workspace…" overlay (with Cancel) appears and does not dismiss.

Agent dispatch:
1. Under the same multi-window load, start an agent run that creates a worktree.
2. The dispatch hangs in worktree preparation; the session does not launch.

`sample <pid>` of the running app in either case shows: main thread mostly idle in the run loop; worker threads parked in `GitService.runAdmittedGitData` at `withCheckedThrowingContinuation` (admission wait), with `WorkspaceCodemapBindingEngine.runProjectionPreload` / `processProjectionBatch` feeding git demand. For the dispatch case, a thread is parked for the whole sampling window in `AgentRunMCPToolService.executeStart` → `AgentMCPStartWorktreeCoordinator.prepare(...)`; for the switch case, one in `WorkspaceFileContextStore.loadRoot`.

[...]

PR #437, opened 2026-07-09
Split RepoPrompt into thin executable and app target

provencher:
## Summary

- keep the shipped `RepoPrompt` product, executable target, emitted binary, bundle, packaging paths, and native product scheme identities unchanged
- make `Sources/RepoPromptExecutable` a one-file `@main` shell that delegates to a new internal `RepoPromptApp` library target owning the existing app implementation
- move root app tests and Objective-C bridging-header ownership to `RepoPromptApp` while leaving the separate `RepoPromptMCP` executable dependency intact
- update source-layout guardrails, generated-Xcode manifest contracts, style inputs, and architecture documentation for the split

`RepoPromptApp` remains an internal SwiftPM target; it is not exposed as a new library product or convenience scheme.

## Validation

Passing validation completed for this implementation:

- `make dev-lint`
- `make guardrails`
- `make dev-swift-build PRODUCT=RepoPrompt`
- `make dev-swift-build PRODUCT=repoprompt-mcp`
- coordinated debug packaging/build
- `make xcode-generator-test`
- `make xcode-validate`
- focused regression checks and repeated focused reruns for the affected Agent, Codemap, and DurableArtifact suites
- commit preflight: whitespace, staged-index secret scan, and guardrails passed
- push preflight: clean tree, outgoing-range review, outgoing-range secret scan, and guardrails passed

The required post-commit `pr-ready` invocation passed its safety phase and coordinated Swift lint. Its full-root lane completed with the order-sensitive failures described below, so the script stopped before repeating the already-passing product-build and generated-Xcode lanes.

## Full-root test disclosure

Two earlier full-root runs completed but hit unrelated order-sensitive failures in untouched suites. Prompt failures reproduced on clean `main`; Agent, Codemap, and DurableArtifact failures passed repeated focused reruns in the implementation checkout.

The post-commit `pr-ready` run added a third full-root attempt and likewise completed with Prompt preassembly timeouts plus a Codemap/MCP code-structure assertion. This target split changes test imports/module ownership but does not modify those behaviors.

PR #443, opened 2026-07-09
fix(mcp): bound agent_run provisioning stalls (#419)

moonray:
## Summary

Bounds two unbounded awaits in `agent_run` / `agent_explore` provisioning that turned dispatch stalls into ~30-minute MCP client idle timeouts under multi-window load (#419).

### `HeadlessAgentConnectionGate` — the bootstrap amplifier
The gate is a process-wide singleton that serializes every agent bootstrap, and its `acquire` wait had **no deadline**. When one start's bootstrap stalled, every subsequent start across all windows queued indefinitely behind it — the "concurrent multi-window load" trigger. (Operationally corroborated: restarting only the MCP server didn't clear it; a full app restart was required — pinning the stuck state to the app process, exactly where this singleton lives.)

- `acquire` / `acquireWithDiagnostics` / `waitForClearConnection` now take a bounded `deadline` (default 120s, `MCPTimeoutPolicy.headlessAgentConnectionAcquireDeadline`). On expiry the queued waiter resumes with `.timedOut` → returns `acquired: false` → the runner tears the start down cleanly instead of parking forever.
- `WindowStateComposition.make` gains a `workspaceSearchService` injection seam (mirrors the existing `workspaceFileContextStore` seam) for deterministic tests.

### `WorkspaceManagerViewModel.loadWorkspaceFolders` — the convergence loop
The search-readiness convergence loop had **no iteration cap** and was **uncancellable by construction** (its guards checked hydration but never `Task.isCancelled` — which is why the MCP execution watchdog's cancellation couldn't rescue the stall). Under multi-window catalog churn the live `catalogGeneration` keeps advancing, so `rebuildIndex` chased a moving target indefinitely.

- Added `maxSearchIndexConvergenceRebuilds = 3` cap + a `!Task.isCancelled` guard. On cap, `break` to the existing `.ready` / `.degraded` decision — safe because `startKeepingFresh` (subscribed before the loop) self-heals in the background and `isStale` reporting stays honest.
- `WorkspaceSearchService` gains a DEBUG-only `setRebuildIndexReportedGenerationTransformForTesting` hook (idiomatic; mirrors `setSearchDidCaptureGenerationHandler`) for deterministic non-convergence tests.

`loadRoot`'s directory walk is intentionally **not** bounded here — it is a finite computation already covered externally by the watchdog; only a dead mount would hang it, which a Swift cap does not fix.

## Design note
`agent_run` stays `.lifecycleManagedCancellable` (no dispatch watchdog) by deliberate contract (`testAgentRunAndExploreUseLifecycleManagedExemption`). This PR bounds the two specific unbounded awaits rather than reversing that contract; an inline watchdog wrap of the `@MainActor` provisioning path was rejected as fighting the isolation.

## Tests
- `HeadlessAgentConnectionGateDeadlineTests` — a queued acquire with a short deadline returns `false` and the waiter is removed from the queue; uncapped it parks forever.
- `WorkspaceSearchIndexConvergenceCapTests` — a deterministic non-converging rebuild drives the loop; asserts `.switched`, elapsed < 10s, and exact `rebuildIndex` count `== 1 + maxSearchIndexConvergenceRebuilds`.

Both green on `main` (post #437 target split — tests import `RepoPromptApp`). Red state re-confirmed on `main`: the gate has no `deadline` and the convergence loop has no cap / no `Task.isCancelled` guard, so the fix is still needed. Regression suites green (`ContextBuilderRunLifecycleTests` 10/10, `WorkspaceSwitchRecoveryTests` 22/22).

## Follow-ups
- Contract ledger: `verify-ledger` reports `missing=2` for these new methods (main itself is currently `missing=22`, so it's a curation tool, not a CI gate). Happy to add the two rows if you want them curated now.
- The convergence loop is a candidate for future extraction into a bounded helper; an Oracle review judged that over-engineering for a single call site, so this ships the minimal inline cap.

Refs repoprompt/repoprompt-ce#419.

### DIFF
--- a/Sources/RepoPrompt/App/WindowStateComposition.swift
+++ b/Sources/RepoPrompt/App/WindowStateComposition.swift
@@ -32,6 +32,7 @@ enum WindowStateCompositionFactory {
         contextBuilderProviderFactory: ContextBuilderAgentViewModel.ProviderFactory? = nil,
         aiQueriesServiceFactory: ((_ keyManager: KeyManager) -> AIQueriesService)? = nil,
         workspaceFileContextStore injectedWorkspaceFileContextStore: WorkspaceFileContextStore? = nil,
+        workspaceSearchService injectedWorkspaceSearchService: WorkspaceSearchService? = nil,
         workspaceSwitchTimingPolicy: WorkspaceSwitchTimingPolicy = .production,
         loadStoredAPISettingsDataOnInit: Bool = true,
         codexModelPollingService: CodexModelPollingService = .shared
@@ -45,7 +46,7 @@ enum WindowStateCompositionFactory {
             let defaultWorkspaceFileContextStore = WorkspaceFileContextStore()
         #endif
         let workspaceFileContextStore = injectedWorkspaceFileContextStore ?? defaultWorkspaceFileContextStore
-        let workspaceSearchService = WorkspaceSearchService()
+        let workspaceSearchService = injectedWorkspaceSearchService ?? WorkspaceSearchService()
         let workspaceFilesViewModel = WorkspaceFilesViewModel(workspaceFileContextStore: workspaceFileContextStore)
 
         // 2) AI queries
--- a/Sources/RepoPrompt/Features/Workspaces/ViewModels/WorkspaceManagerViewModel.swift
+++ b/Sources/RepoPrompt/Features/Workspaces/ViewModels/WorkspaceManagerViewModel.swift
@@ -5936,6 +5936,13 @@ class WorkspaceManagerViewModel: ObservableObject {
         3
     }
 
+    /// Maximum extra `rebuildIndex` passes the search-readiness convergence loop in
+    /// `loadWorkspaceFolders` will spend chasing a live catalog generation before
+    /// breaking out and letting the keep-fresh listener finish catch-up. Bounds the
+    /// otherwise unbounded rebuild chase under multi-window catalog churn.
+    /// See repoprompt-ce #419.
+    nonisolated static let maxSearchIndexConvergenceRebuilds = 3
+
     nonisolated static func boundedWorkspaceRootLoadLimit(forRootCount rootCount: Int) -> Int {
         guard rootCount > 0 else { return 0 }
         return min(rootCount, maxConcurrentWorkspaceRootLoads)
@@ -6328,10 +6335,20 @@ class WorkspaceManagerViewModel: ObservableObject {
             _ = await warmedGeneration
         #endif
         guard isHydrationGenerationCurrent(hydrationGeneration, workspaceID: workspace.id) else { return }
+        var convergenceRebuilds = 0
         while true {
             let currentCatalogGeneration = await fileManager.workspaceFileContextStore.catalogGeneration(rootScope: .visibleWorkspace)
-            guard isHydrationGenerationCurrent(hydrationGeneration, workspaceID: workspace.id) else { return }
+            guard !Task.isCancelled,
+                  isHydrationGenerationCurrent(hydrationGeneration, workspaceID: workspace.id) else { return }
             guard currentCatalogGeneration != indexGeneration else { break }
+            guard convergenceRebuilds < Self.maxSearchIndexConvergenceRebuilds else {
+                // The keep-fresh listener (subscribed above) continues catch-up in the
+                // background and queries report `isStale` until it lands. Bounding this
+                // readiness barrier avoids an unbounded rebuild chase under multi-window
+                // catalog churn (repoprompt-ce #419).
+                break
+            }
+            convergenceRebuilds += 1
             #if DEBUG
                 let loopSearchCatalogSnapshotStartMS = WorkspaceRestorePerfLog.timestampMSIfEnabled()
             #endif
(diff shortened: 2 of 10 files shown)


moonray:
Re: scope — **yes, "deadline + convergence caps" is the intended scope here; the git-pool foreground lane is deliberately deferred, not omitted.** Three reasons:

**1. The git-admission slice is no longer the hang vector.** `GitProcessAdmissionController` already ships foreground deadline/EDF admission with `.deadlineExceeded` fail-fast (v1.0.24; `GitService` auto-attaches a deadline for `rootBootstrap`/`userInitiatedAuthority` at `GitService.swift:4006/4789/6864`). So git admission no longer parks a foreground waiter indefinitely — it errors within the deadline. (Confirmed on `main`: there's still **no** reserved foreground lane — the controller's own doc says "Active subprocesses remain non-preemptive" — but the deadline is what stops the hang.)

**2. This PR closes the residual non-git hang surface that the git fix exposed.** #419's `sample` was captured on v1.0.24 — *with* the git deadline — and still stalled. The stall had moved to the non-git awaits in provisioning: the `HeadlessAgentConnectionGate` bootstrap wait (process-wide singleton, no deadline → the multi-window amplifier) and the `loadWorkspaceFolders` convergence loop (no cap, and *uncancellable by construction* — its guards never checked `Task.isCancelled`, which is why the dispatch watchdog's cancellation couldn't rescue it). `agent_run` stays `.lifecycleManagedCancellable` (no dispatch watchdog) per the existing contract (`testAgentRunAndExploreUseLifecycleManagedExemption`), so any unbounded await there degenerates to the ~30-min client-idle hang. Bounding those two is exactly what this PR does.

**3. The foreground lane is a complementary, distinct change.** It targets foreground git *throughput* (work completing promptly under saturation) — #419's named root cause — whereas the admission deadline converts "hang" → "fail-fast" and this PR converts "non-git hang" → "bounded." A reserved/preemptive foreground lane is a deeper admission-fairness change that alters cross-window scheduling for *all* windows, not just the `agent_run` path.

**Suggested resolution:** keep this PR scoped as the hang-bounding fix, and track the foreground lane as the core #419 work (separate follow-up PR, or directly under #419). If you'd rather close #419 in one move, I'm happy to expand this PR to add a reserved foreground lane to `GitProcessAdmissionController` — but I'd treat that as a separate decision since it changes admission semantics broadly.

baron (changes requested):
Requesting changes because this does not fix the primary stall documented in #419.

## Blocking findings

1. **The reported `worktree_create` stall occurs before the gate this PR changes.**

   `AgentRunMCPToolService.executeStart` awaits `startWorktreeCoordinator.prepare(...)` before provider startup. The provider startup is the later phase that eventually reaches `HeadlessAgentConnectionGate`. Issue #419's sample places the blocked operation in worktree preparation / Git admission, so the new gate timeout cannot bound that observed path.

   This PR does not change `GitProcessAdmissionController`, and therefore does not satisfy the issue's foreground Git-admission acceptance criterion. Please either fix that starvation mechanism or rescope this PR as secondary hardening that does not close #419.

2. **The gate timeout is per wait-cycle rather than an aggregate deadline.**

   `acquireWithDiagnostics` schedules a fresh full `Duration` on every iteration of its `while activeConnectionID != nil` loop. If a resumed waiter loses the freed gate to another caller, it can wait another complete 120 seconds. Under contention, the total acquisition time is still unbounded.

   Please compute one absolute deadline at entry and use the remaining duration, or transfer ownership directly to the resumed FIFO waiter. The regression test should cover repeated wake-and-steal cycles.

3. **The typed timeout is not propagated consistently through actual Agent Mode starts.**

   `lastAcquireWasTimeout` is interpreted by `AgentRunCoordinator.prepareAndInstallPolicy`, but the Codex, Claude, headless, and ACP Agent Mode runners consume `MCPBootstrapLease.acquire()` as a plain Boolean. Codex can silently return; the others report generic acquisition/routing failures. The timeout-specific error therefore does not consistently reach the `agent_run` paths this PR claims to improve.

4. **The convergence cap publishes `.ready` despite not converging.**

   After the cap is reached, `loadWorkspaceFolders` publishes `.ready` whenever root loading had no failures, even though the live catalog and index did not converge. Please expose a capped/degraded state or otherwise preserve the fact that the index is stale.

The convergence test also transforms `rebuildIndex`'s reported generation rather than reproducing actual catalog-generation churn. It validates cap arithmetic, but not the claimed production mechanism. Since stores are per-window, the multi-window causal narrative needs either evidence from a churn-based reproduction or softer wording.

The timeout and cap may be useful defensive hardening, but in their current form they add complexity without fixing the root starvation path, and the timeout itself is not truly bounded.

baron:
Recommend closing exact head `feeffdd067299b325117c0f428550bd9d340f429`. The acquisition loop still creates a fresh full deadline on each iteration, so repeated wake/steal cycles remain unbounded, and the patch does not cover #419's full foreground/worktree path. The branch is deeply stale, conflicting, and red; any remaining #419 work should be a focused current-main fix with an aggregate deadline.

baron:
Recommend closing exact head `feeffdd067299b325117c0f428550bd9d340f429`. The acquisition loop still creates a fresh full deadline on each iteration, so repeated wake/steal cycles remain unbounded, and the patch does not cover #419's full foreground/worktree path. The branch is deeply stale, conflicting, and red; any remaining #419 work should be a focused current-main fix with an aggregate deadline.

PR #448, opened 2026-07-09
Fix Context Builder teardown finalization

### LINKED ISSUE #444: Context Builder drops final context on non-orderly discovery-child close and leaks discovery agents on run supersession
## Summary

Context Builder's finalization/teardown pipeline assumes an **orderly, single, same-window peer-EOF** discovery-child lifecycle, and *drops* state rather than reconciling it whenever that assumption is violated. This produces two faults that share one teardown pipeline and are evidenced together by a single incident:

- **Fault A — final context is dropped on any non-orderly discovery-child close.** After discovery genuinely completes, the run fails at commit with `Context Builder final context is missing for run <RUN_ID> (connection none)` because the run→connection mapping was destroyed without being preserved. Final-context preservation is gated on an orderly peer-EOF close of the discovery child; every other close cause discards the finalized context.
- **Fault B — discovery `claude -p` children (and their launch-config leases) leak when a run record is superseded.** Child-process termination is owned solely by the run's finalize path, which early-returns for superseded/stale records. Rapid re-invocation on the same tab supersedes older records before their teardown runs, so their children are never SIGTERMed and their `discovery-*.json` lease files are never deleted. There is no independent reaper.

These belong in one issue: they share the same teardown pipeline (`finalizeContextBuilderRun → scheduleRunTeardown → provider.dispose()` plus the connection-removal path), and a single failing run exhibits both — it drops its final context *and* leaks its child.

Observed on **release 1.0.27 (build 28)**, reproduced on **both the MCP and CLI transports**.

## Expected behavior

- A Context Builder run that completes discovery should publish its final context and write its export regardless of *how* the discovery child's transport closed (orderly peer EOF, server-initiated terminate, write-hangup, transport error, or TTL). Final context should be preserved for any discover-run close that still has committed tab context, or captured at stream-completion time before teardown.
- Every retirement of a run record — including records that are superseded or go stale before finalize runs — must guarantee `provider.dispose()` (child SIGTERM + launch-config lease/file cleanup). No discovery child or `discovery-*.json` lease file should outlive its run.

## Actual behavior

**Fault A — `missingFinalContext (connection none)`.** A CLI (`rpce-cli`) run completed end-to-end in ~94s (`exit 0`); the progress log emitted `Context Builder run finalization completed in 0.001s (total 93.686s)` and `Context builder complete`, then the call failed with `Context Builder final context is missing for run <RUN_ID> (connection none)`, `Total tokens: 0`, and **no oracle export file was ever written**. This is not a client timeout — discovery finished and finalization reported complete; the app-side publish/commit is what failed. The same failure reproduced on the MCP transport.

Mechanism (verified by source read on the 1.0.27 tree):

1. At commit, `commitTabContextForAgent` resolves the run's connection via `mcpServer.contextBuilderFinalContextConnectionID(runID:)` (`ContextBuilderAgentViewModel.swift`, ~`:2651`). When that returns `nil`, the only constructor on that branch hard-codes the connection id:
   - `ContextBuilderAgentViewModel.swift:2718-2721` → `.missingFinalContext(runID: runID, connectionID: nil)`.
   - Rendered as `... (connection none)` at `ContextBuilderAgentViewModel.swift:2391-2395` (the `connection none` text appears iff `connectionID == nil`).
2. The lookup returns `nil` because the run→connection association was destroyed at connection-removal time without being preserved into the detached fallback. In `MCPConnectionManager.removeConnection` (~`:5956-5966`), the final-context ownership transfer (`detachContextBuilderRunID`) is set **only** when all three hold: `context.reason == peerEOF` **and** `context.initiator == .peer` **and** the connection's run purpose `== .discoverRun`. Any other close cause yields `detachContextBuilderRunID == nil`, so the tab context is removed with **no** detached fallback and the subsequent lookup is `nil`.
3. The discovery child's real MCP peer is a `repoprompt_ce_cli` grandchild of `claude -p`. When `claude -p` exits and tears that peer connection down as a non-orderly close, the orderly-peer-EOF gate is missed and the finalized context is dropped even though discovery succeeded.
4. This preservation gate was introduced by commit `f2850ecc` "Fix Context Builder completion after discovery peer EOF (#254)". The gate is correct for the orderly-EOF shape — but it is *narrow*: it is the only path that preserves final context, so every other close cause silently drops it.

**Fault B — leaked discovery children and lease files.** In one observation window, **5 orphaned headless discovery agents** (`claude -p … --mcp-config …/MCP/LaunchConfigs/discovery-R-<UUID>.json`, each with a `repoprompt_ce_cli` child still connected back to the app socket) were left sleeping at 0% CPU — one still alive ~4.5 hours after its run. On disk, **5 `discovery-R-*.json` lease files were undeleted, a 1:1 match** with the 5 leaked children. Four of the launches occurred ~20s apart within one minute — consistent with **client-side** re-invocation of `context_builder` (there is no app-side discovery retry loop), each retry leaking a child.

Mechanism (verified by source read):

1. Discovery-child termination is owned *solely* by `finalizeContextBuilderRun → scheduleRunTeardown → provider.dispose()` (SIGTERM over the process registry; the `discovery-*.json` lease/config file is deleted only by the lease `release()`/`deinit` invoked from that same disposal). There is **no** independent reaper, watchdog, or OS supervision.
2. `finalizeContextBuilderRun` early-returns *before* `scheduleRunTeardown` when `acceptsEvents(from:record)` is false (or when `claimTerminal` fails). `acceptsEvents` (`ContextBuilderRunLifecycle.swift:279-285`) requires the record to still own the tab slot: `recordsByRunID[runID] === record && activeRunIDByTabID[tabID] == record.runID && currentSession === record.session && !record.isTerminal && session.activeRunOwnership == record.ownership`.
3. When a newer run supersedes the tab slot (rapid re-invocation on the same tab), the older record fails `acceptsEvents`, so its `provider.dispose()` never fires — its `claude -p` child is never SIGTERMed and its lease file is never deleted.

**Why the two faults correlate.** The same nil association behind Fault A also empties `connectionIDs(forRunID:)`, so the *secondary* connection-scoped cleanup (`clearTabContextForAgent`) finds nothing to terminate. But the *primary* process disposal is independent of that mapping — it is skipped by record supersession (Fault B), not by the nil lookup. On a run that both closes non-orderly and is superseded, both the connection-scoped cleanup (empty mapping) and the process-scoped disposal (skipped finalize) miss, compounding the leak.

## Reproduction sketch

1. Use a RepoPrompt CE build at release 1.0.27 (build 28) or equivalent, with Context Builder available.
2. Invoke `context_builder` (plan mode, `export_response: true`) via either the MCP or CLI (`rpce-cli`) transport, ideally with more than one app window open.
3. Let discovery run to completion (progress log reaches `Context Builder run finalization completed …` / `Context builder complete`).
4. Observe the call fail with `Context Builder final context is missing for run <RUN_ID> (connection none)`, `Total tokens: 0`, and no export file on disk — despite discovery having finished. (Fault A: the discovery child's `repoprompt_ce_cli` peer closed via a non-orderly cause, missing the orderly-peer-EOF preservation gate.)
5. Re-invoke `context_builder` on the same tab a few times in quick succession (~20s apart). After the runs settle, enumerate processes for surviving `claude -p … discovery-*.json` children and the `LaunchConfigs/` directory for undeleted `discovery-*.json` lease files. Observe leaked children with a 1:1 match of undeleted lease files. (Fault B: superseded records skipped their `provider.dispose()`.)

## Impact

- **Context Builder is unusable for its primary purpose on affected runs**: ~94s of real discovery work is thrown away at the last step, returning 0 tokens and no export, so callers (agents and humans) get a hard failure after a long wait.
- **Resource leak**: orphaned `claude -p` discovery children accumulate across retries (observed alive for hours), each holding a live `repoprompt_ce_cli` peer connection and a launch-config lease open; undeleted `discovery-*.json` files accumulate in parallel. Because there is no reaper, these must be killed manually.
- **Invisible in release builds**: Context Builder routing diagnostics are debug-only and CB runs are not registered as first-class agent sessions, so neither the drop nor the leak surfaces in logs or any operator-visible list — the failure is silent until discovered via `ps`.

## Related issues checked

- **#170** (Context Builder stall / transport-closure attribution): nearest sibling — **same pipeline family, different mechanism**. #170's one open item is a discovery-phase *stall* (a run that hangs). This is a **hard finalization drop after discovery successfully completed** (finalization reported complete in 0.001s, then reported the context missing). #170's remediation landed as PR #171 ("Fix MCP transport timeout recovery") and **is present in the failing build** — it made the transport cancellation-safe but does not address the close-cause preservation gate or the disposal-on-supersession gap.
- **#169** (per-call `context_id` / wrong-tab tripwire swallowed by `try?`): **not on this code path**. CB finalization resolves its connection through `contextBuilderFinalContextConnectionID(runID:)` on the run's own `mcpServer`, not through `resolveTabContextSnapshot`, so the swallowed wrong-tab tripwire does not drive `missingFinalContext`. Same multi-window family, distinct bug.
- **#26** (abort/cancellation propagation to Oracle/Context Builder): distinct — this failure is not an abort; discovery ran to completion and finalization reported success before the context was dropped.
- **#312** (Agent Mode offers Context Builder when no Oracle model is configured): different failure mode — #312 is a **configuration error** surfaced up front; here the discovery phase ran a full ~94s with a configured model, then dropped the result at commit.
- **#254 / commit `f2850ecc`** ("Fix Context Builder completion after discovery peer EOF"): **introduced the narrow preservation gate this issue widens**. Its orderly-peer-EOF handling is correct but is the *only* path that preserves final context, so it does not cover non-orderly closes.

No existing upstream issue or PR references `missingFinalContext` or "final context is missing" — this exact failure is unfiled.

## Acceptance criteria

- **Fault A:** A Context Builder run whose discovery completed publishes its final context and writes its export for **any** discovery-child close cause that still has committed tab context — not only orderly peer EOF. Equivalent alternative: final context is captured at discovery stream-completion time (before teardown) so it does not depend on the connection still being mapped at commit. The `missingFinalContext(connection none)` path no longer fires after a successful discovery.
- **Fault B:** `provider.dispose()` (child SIGTERM + launch-config lease/file cleanup) is guaranteed on **every** run-record retirement, including superseded and stale records that currently fail `acceptsEvents`/`claimTerminal`. No `claude -p` discovery child and no `discovery-*.json` lease file outlives its run. The connection-scoped cleanup (`clearTabContextForAgent`) does not depend on mappings that finalization may already have destroyed.
- **Regression tests** cover (a) a discovery run whose child transport closes via a non-orderly cause (terminate / write-hangup / transport error / TTL) still yields a preserved final context and a written export, and (b) an older run record superseded on the same tab before finalize still has its child disposed and its lease file removed.
- Optional (observability): a runID → {pid, leaseURL, connectionID} registry so both the drop and the leak are enumerable in release builds, where CB diagnostics are otherwise debug-only.

baron:
## Summary

Fixes #444.

- Preserve Context Builder final context for every discovery-run teardown path, not just orderly peer EOF.
- Ensure stale/superseded Context Builder run records still retire safely, cancel pending publication, resolve continuations, release active ownership when appropriate, and dispose the provider/lease via teardown.
- Add regression coverage for non-orderly discovery-child close causes and stale run retirement/provider disposal.

## Validation

- `make dev-test FILTER=ContextBuilderRunLifecycleTests` — passed 11 tests, 0 failures (conductor client later timed out, direct job log showed the suite passed)
- `make format-check` — passed
- `swiftlint lint --strict` — passed
- `git diff --check` — passed
- `.agents/skills/rpce-contribution-check/scripts/preflight.sh commit` — passed before commit
- `.agents/skills/rpce-contribution-check/scripts/preflight.sh push` — passed before push

## Notes

I did not run the full `pr-ready` lane; focused lifecycle tests and Swift style checks covered the changed boundary for this fix.

### DIFF
--- a/Sources/RepoPrompt/Features/ContextBuilder/ViewModels/ContextBuilderAgentViewModel.swift
+++ b/Sources/RepoPrompt/Features/ContextBuilder/ViewModels/ContextBuilderAgentViewModel.swift
@@ -464,6 +464,19 @@ final class ContextBuilderAgentViewModel: ObservableObject {
         func isRunTeardownPendingForTesting(runID: UUID) -> Bool {
             runRegistry.record(runID: runID)?.isTeardownPending == true
         }
+
+        func retireStaleRunRecordForTesting(
+            _ record: ContextBuilderRunRecord,
+            waiterResolution: ContextBuilderRunWaiterResolution = .snapshot,
+            cancelExecution: Bool = true
+        ) {
+            retireContextBuilderRunRecordWithoutPublishing(
+                record,
+                waiterResolution: waiterResolution,
+                cancelExecution: cancelExecution,
+                source: "contextBuilder.testing.staleRetirement"
+            )
+        }
     #endif
 
     // MARK: - Published session-scoped proxies
@@ -1939,9 +1952,20 @@ final class ContextBuilderAgentViewModel: ObservableObject {
         saveHistory: Bool,
         source: String
     ) -> Bool {
-        guard acceptsEvents(from: record) else { return false }
+        guard acceptsEvents(from: record) else {
+            retireContextBuilderRunRecordWithoutPublishing(
+                record,
+                waiterResolution: waiterResolution,
+                cancelExecution: cancelExecution,
+                source: source
+            )
+            return false
+        }
         flushAssistantPreview(for: record)
-        guard record.claimTerminal(outcome) else { return false }
+        guard record.claimTerminal(outcome) else {
+            scheduleRunTeardown(record, cancelExecution: cancelExecution)
+            return false
+        }
 
         let session = record.session
         session.lastAgentOutput = record.output.fullOutput()
@@ -2018,6 +2042,47 @@ final class ContextBuilderAgentViewModel: ObservableObject {
             continuation?.resume(throwing: CancellationError())
         }
         return true
+    }
+
+    /// Retires records that have lost active ownership before their terminal path runs.
+    ///
+    /// Superseded/stale records must not publish previews, mutate current tab state, or
+    /// restore run-start configuration over a newer run. They still own provider/process
+    /// resources, continuations, and possibly a registry record, so retirement must always
+    /// schedule teardown and resolve any waiter exactly once.
+    private func retireContextBuilderRunRecordWithoutPublishing(
+        _ record: ContextBuilderRunRecord,
+        waiterResolution: ContextBuilderRunWaiterResolution,
+        cancelExecution: Bool,
+        source: String
+    ) {
+        record.previewPublicationTask?.cancel()
+        record.previewPublicationTask = nil
+
+        let didClaimTerminal = record.claimTerminal(.cancelled)
+        record.session.endRunAttempt(ifCurrent: record.ownership, source: "\(source).staleRetirement")
+        if runRegistry.releaseActiveSlot(for: record) {
+            tabsWithActiveContextBuilderRun.remove(record.tabID)
+        }
+
+        let continuation = didClaimTerminal ? record.takeContinuation() : nil
+        let completion = MCPContextBuilderRunCompletion(
+            runID: record.runID,
+            tabID: record.tabID,
+            terminalDisposition: .cancelled,
+            agentOutput: record.output.fullOutput(),
+            usedAgentOutputAsPrompt: false,
+            committedTab: nil
+        )
+
+        scheduleRunTeardown(record, cancelExecution: cancelExecution)
+
+        switch waiterResolution {
+        case .snapshot:
+            continuation?.resume(returning: completion)
+        case .cancellationError:
+            continuation?.resume(throwing: CancellationError())
+        }
     }
 
     private func scheduleRunTeardown(
--- a/Sources/RepoPrompt/Features/ContextBuilder/ViewModels/ContextBuilderAgentViewModel.swift
+++ b/Sources/RepoPrompt/Features/ContextBuilder/ViewModels/ContextBuilderAgentViewModel.swift
@@ -465,6 +465,10 @@ final class ContextBuilderAgentViewModel: ObservableObject {
             runRegistry.record(runID: runID)?.isTeardownPending == true
         }
 
+        func replaceSessionForTesting(tabID: UUID) {
+            sessions[tabID] = TabSession(tabID: tabID)
+        }
+
         func retireStaleRunRecordForTesting(
             _ record: ContextBuilderRunRecord,
             waiterResolution: ContextBuilderRunWaiterResolution = .snapshot,
@@ -1473,8 +1477,10 @@ final class ContextBuilderAgentViewModel: ObservableObject {
                 session.followUpOracleSessionID = nil
             }
 
-            // 3. Logically cancel the active run without waiting for teardown.
-            if let record = runRegistry.activeRecord(tabID: tabID) {
+            // 3. Logically cancel every registered run for the tab without waiting for teardown.
+            // A superseded record may no longer own the active slot but still owns a provider,
+            // execution task, waiter, and launch-config lease that tab close must retire.
+            for record in runRegistry.records(tabID: tabID) where !record.isTerminal {
                 debugLog("handleComposeTabsWillClose: cancelling run \(record.runID) for tab \(tabID)")
                 cancelRun(
                     record,
@@ -2384,6 +2390,10 @@ final class ContextBuilderAgentViewModel: ObservableObject {
                 )
                 guard !Task.isCancelled, acceptsEvents(from: record) else { return .cancelled }
                 debugLog("Routing result for run \(runID): routed=\(routed)")
+                if routed {
+                    record.finalContextConnectionIDForDiagnostics =
+                        mcpServer.contextBuilderFinalContextConnectionID(runID: runID)
+                }
 
                 if !routed, record.origin.isMCP {
                     let timeoutSeconds = TimeInterval(ContextBuilderDefaults.mcpRoutingTimeoutMilliseconds) / 1000
@@ -2609,7 +2619,15 @@ final class ContextBuilderAgentViewModel: ObservableObject {
         waiterResolution: ContextBuilderRunWaiterResolution,
         saveHistory: Bool
     ) {
-        guard acceptsEvents(from: record) else { return }
+        guard acceptsEvents(from: record) else {
+            retireContextBuilderRunRecordWithoutPublishing(
+                record,
+                waiterResolution: waiterResolution,
+                cancelExecution: true,
+                source: "contextBuilder.cancel.staleRetirement"
+            )
+            return
+        }
         guard record.canAcceptCancellation else {
             debugLog("Cancel ignored after final context commit claim for run \(record.runID)")
             return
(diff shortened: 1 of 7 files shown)


mplibunao:
Thanks for taking this on. I walked the diff against the #444 reproduction. The discover-run detach gate is correctly widened at MCPConnectionManager.swift:6002, with the live-context and ownership guards intact at MCPServerViewModel+TabContext.swift:4021-4026. Stale retirement also disposes the provider and resolves a waiter when its helper runs. The preview-task cancellation, conditional slot cleanup, and close-cause diagnostics add useful hardening.

I see three remaining gaps before this covers the reported failure modes:

1. **The context is preserved, but production finalization still rejects it after every non-orderly close.** MCPConnectionManager.swift:5913-5925 publishes `.detachedWithoutOrderlyPeerEOF`, and ContextBuilderRunLifecycle.swift:57-63 accepts only `.peerEOFDetached`. The new test at ContextBuilderRunLifecycleTests.swift:592-690 confirms that a read-error handoff never reaches commit. The table at :784-827,876-909 calls `commitContextBuilderTabContext` directly, so it proves the stored context can be committed but bypasses the production resolver. For #444, a discovery that has already completed successfully still needs to publish/export regardless of the subsequent close shape; a cancellation or failure before discovery completion should remain unsuccessful. A completion-aware signal seems safer than using close cause alone as the success gate.

2. **A silent superseded run still has no production retirement trigger.** The finalize early-return is fixed at ContextBuilderAgentViewModel.swift:1955-1962, but `cancelRun` still returns at :2612 when ownership has already flipped. If the displaced child stream stays silent, its own finalize never runs, and cancel/tab-close also skips retirement. The new test at ContextBuilderRunLifecycleTests.swift:1014-1085 calls the DEBUG retirement hook directly, so it validates the helper rather than this production wiring case.

3. **Lease cleanup and the diagnostic ID need stronger coverage.** The fake-provider test verifies that `dispose()` is requested, but it can't show that a real `discovery-*.json` lease is deleted while stream cleanup is stalled. The missing-final-context constructor also still passes `connectionID: nil` at ContextBuilderAgentViewModel.swift:2797.

Suggested regression scenarios from the reporter reproduction:

- Run the non-orderly close table through production finalization after marking discovery complete, and assert commit/export succeeds; retain a negative case where discovery didn't complete.
- Flip session ownership while a provider event/stream is parked, then exercise cancel-by-run-ID or tab close. Assert provider disposal occurs once, the MCP waiter finishes within a bound, cleanup completes, and a real lease file disappears.
- Force the same ownership flip but let the stream finish, covering the finalize early-return separately.
- Capture a live discovery connection ID, remove its mapping, and assert the resulting `missingFinalContext` diagnostic names that ID.

The current diff has the right storage handoff and a useful retirement path. The remaining work is to make both reachable under the incident shapes and make the successful-discovery decision independent of transport close cause.

baron:
@mplibunao  thanks for testing and the excellent feedback, will address these issues as well

baron:
Thanks — these were good catches. I updated the PR at `09e06343` to address them:

- Non-orderly close is no longer treated uniformly as failure. Cleanup now captures the transport's response-delivery state before removal. A retained discovery context may finalize successfully after a bad close only when all accepted requests have delivered responses; a pending response or unavailable snapshot remains a failed finalization. The production finalizer tests now cover both the successful drained/read-error path and an explicit `pendingRequestCount: 1` failure twin.
- Stale/superseded cancellation now retires the registered run instead of returning when active ownership has flipped, and tab close retires every non-terminal record for that tab. The production cancellation test forces the stale-owner path, waits for provider disposal, and verifies an actual `discovery-*.json` lease file is deleted.
- The routed final-context connection ID is remembered on the run record and reused in missing-final-context diagnostics if live/detached routing state has already disappeared.

I also had the integrated follow-up diff reviewed specifically against these concerns; no remaining must-fix issue was found. Fresh CI is running on the updated head now.

PR #450, opened 2026-07-09
fix: update Codex app-server text payload key

### LINKED ISSUE #449: Codex app-server turn dispatch hangs indefinitely: payload uses textElements instead of text_elements
## Summary

Every Agent Mode session backed by a Codex CLI model hangs indefinitely at "Thinking…" because the app-server turn dispatch payload uses a camelCase key, `"textElements"`, that the Codex app-server does not recognize — the correct key is snake_case, `"text_elements"`. The turn is built and sent, but Codex never acts on it, so the session waits forever with no error surfaced.

## Expected behavior

A Codex-backed Agent Mode session (including a Codex-configured Context Builder discovery agent) dispatches its turn and Codex begins responding within the normal request/response latency for that model.

## Actual behavior

The app-server turn payload for a text input element is built with the key `"textElements": []` instead of `"text_elements": []`. Codex CLI rejects or silently ignores the malformed payload, so the turn is never dispatched and the session hangs at "Thinking…" with no timeout or error. Reproduced on release 1.0.27: two agent sessions hung until manually cancelled; switching the same roles to Claude models resolved the hang immediately with no other change. A Codex-configured Context Builder discovery agent hangs the same way during discovery, since it drives the same app-server turn-dispatch path.

## Reproduction sketch

1. Configure an Agent Mode role (or Context Builder discovery) to use a Codex CLI model via the app-server integration.
2. Start a session with any text prompt.
3. Observe the session enters "Thinking…" and never progresses — no response, no error, no timeout.
4. Cancel and switch the same role to a Claude model with no other change; the session responds normally.

## Impact

- Any Codex-configured Agent Mode role is unusable — every session hangs until manually cancelled.
- A Codex-configured Context Builder discovery agent hangs identically during discovery, blocking context-building workflows that route through Codex.
- The hang presents with no error, so it is easy to misattribute to an unrelated timeout or infrastructure issue rather than a payload-key mismatch.

## Related issues checked

This is distinct from #444, which covers Context Builder final-context/teardown handling and is provider-agnostic (it applies regardless of which model backs discovery). It is plausibly related to #170's unexplained 300-second discovery-only stall — a Codex-configured discovery agent would hang exactly like this — though that is a hypothesis, not a confirmed link, since #170's stall was observed independently of provider configuration.

## Acceptance criteria

- The app-server turn dispatch payload uses `"text_elements"` for text input elements.
- A regression test covers turn-dispatch payload construction so the key cannot silently regress back to the camelCase form.
- A Codex-configured Agent Mode session (and Codex-configured Context Builder discovery) dispatches turns and receives responses normally.

mplibunao:
## Summary

The app-server turn dispatch payload for a text input element used the camelCase key `"textElements"` instead of the snake_case `"text_elements"` that Codex CLI expects. Codex rejects or silently ignores the malformed payload, so the turn is never dispatched and the session hangs indefinitely at "Thinking…". Every Codex-configured Agent Mode session (including a Codex-configured Context Builder discovery agent) is affected.

## Fix

One-line key rename in the turn dispatch payload builder:

```swift
- "textElements": []
+ "text_elements": []
```

## Validation

- `CodexNativeSessionControllerTurnDispatchTests` (new, 9 methods covering turn start/steer/interrupt/cancellation dispatch): all pass.
- `make dev-swift-build PRODUCT=RepoPrompt`: succeeds.
- `swiftlint lint --strict` scoped to the changed files: 0 violations.
- Live use on a patched debug build since 2026-07-09: Codex-configured Agent Mode sessions dispatch and respond normally; unpatched builds reproduce the hang.
- Contract ledger (`Scripts/Fixtures/test-suite-contract-ledger.tsv`) reconciled to the exact live test method set for the new suite.

Fixes #449

https://claude.ai/code/session_01AsixXS1UnDxsVA7f2111kg


baron (approved):
Reviewed current head e4377612 against main including #545. Protocol schema/source verified, focused regression coverage reviewed, unsupported HTTP image fixture repaired to inline data URL, and independent Fable/engineer/pair concerns reconciled. Approval remains contingent on exact-head required checks completing green.

PR #455, opened 2026-07-09
Prevent unbounded MCP buffering and silent receive drops

morluto:
## Problem

MCP messages are separated by newlines. Previously, a peer could continue sending data without a newline and make the pending frame buffer grow without a bound. Some framing paths also copied the unread remainder after every message, while a full CLI receive queue could silently drop a message.

## What This PR Changes

This PR introduces a shared newline-frame accumulator with a 256 MiB per-message safety limit. It advances a cursor through consumed data and compacts occasionally instead of copying the remaining buffer after every frame.

Oversized frames and full receive queues now terminate the affected connection with an explicit error rather than continuing with incomplete or silently lost input.

The limit is a defensive ceiling, not an expected message size. It remains above RepoPrompt's existing 128 MiB Git-output limit. Ordering, cancellation, fairness, EOF handling, and independently owned delivered frames remain unchanged.

## Changes

- Add a shared cursor-based `MCPNewlineFrameAccumulator`.
- Enforce the per-frame limit with a typed oversized-frame error.
- Compact consumed storage instead of rebuilding the remainder after each frame.
- Continue returning independently owned `Data` values for delivered frames.
- Detect dropped `AsyncThrowingStream` yields and terminate ingress.
- Preserve ordering, cancellation, fairness, EOF, replay, and large-message behavior.
- Clamp non-positive socket-reader chunk sizes.
- Add accumulator, boundary, overflow, fairness, replay, and cancellation coverage.

## Review Order

1. `NewlineFrameAccumulator.swift`.
2. Both `NewlineDelimitedSocketReader` copies.
3. Socket, bootstrap, replay, and proxy integrations.
4. Boundary and overflow tests.

## Validation

- Branch autoreview completed with no actionable findings.
- Formatting, syntax checks, standalone framing probes, reader-copy equality, and `git diff --check` passed.
- Repository guardrails and contribution preflights passed.
- Added test IDs were reconciled with the curated test ledger.

The receive queues remain frame-count bounded. A byte-budgeted queue is separate follow-up work.


baron (changes requested):
### Intent Analysis
**Problem:** MCP newline framing accumulated unterminated frames in growable `Data` buffers with no size cap, and the CLI's bounded `AsyncThrowingStream` could silently drop frames under consumer lag — protocol corruption without explicit failure.
**Before → After:** Ad-hoc `Data` scanning/remainder copying is replaced by a shared cursor-based accumulator with a 256 MiB per-frame cap and compaction (`Sources/RepoPromptShared/MCP/NewlineFrameAccumulator.swift:23-137`). Oversized frames and dropped yields are intended to terminate the connection with explicit diagnostics (`UnixSocketMCPTransport.swift:1429-1463`, `BootstrapSocketMCPTransport.swift:475-505`, `main.swift:1696-1759`).
**Evidence:** Two independent code-first reviews converge on the same files, line ranges, and invariant; all eight CI checks pass on head `a59c965`; PR is MERGEABLE with no prior reviews.
**Confidence:** High on intent and the app-side implementation. Medium on the CLI-side fail-closed guarantee — one review presents a specific, unrefuted race in exactly that guarantee.
**Red Flags:** No linked issue or review discussion. The new CLI overflow test keeps the peer open (`Tests/RepoPromptTests/MCP/Control/BootstrapSocketMCPReceiveOverflowTests.swift:19-23`), so green CI does not cover the disputed teardown-ordering scenario.

### Findings
**F1 — Blocking: CLI overflow terminal cause is race-dependent.** When a yield drops a frame, `BootstrapSocketMCPIngressGate` records only a boolean `isTerminal` (`BootstrapSocketMCPTransport.swift:90-98`) and defers teardown to an unstructured actor task (`:482-489`). EOF/error teardown is independently scheduled (`:494-498`, `:579-604`) with no ordering guarantee. If EOF teardown wins, the stream finishes cleanly (or with an unrelated read error) and the later overflow handler no-ops (`:607-637`) — a genuinely dropped MCP frame surfaces as successful completion. This defeats the PR's own invariant on a realistic path (client bursts, then closes). The app transport in this same PR shows the correct pattern: synchronous authoritative-cause storage resolved at every teardown (`UnixSocketMCPTransport.swift:113-126, 865-890`).
*Required fix:* persist the overflow cause in the CLI ingress gate (mirroring the app side) and resolve it on every teardown path; add a regression test that overflows and immediately closes/half-closes the peer.

**F2 — Non-blocking (follow-ups):** frame queues bounded by count (1,024) not bytes; proxy poll interval tightened 1 s → 100 ms (idle wakeups); no compaction benchmark; no dedicated oversized bootstrap-handshake or reconnect-replay tests (shared accumulator coverage makes this acceptable).

### Maintainer-guidance check
- **Invariant:** accepted frames delivered in order; oversized/truncated/dropped frames must fail explicitly, never appear as successful completion. F1 violates the second half nondeterministically on the CLI side.
- **Root-cause confidence:** confirmed for the problem the PR fixes; F1's residual cause (no authoritative terminal cause in the CLI gate) is mechanistically specific and cited.
- **Authority:** accumulator owns framing limits; the fix belongs in the ingress gate and actor teardown — within this PR's scope, not a follow-up.
- **State-safety:** app transport preserves terminal-cause precedence; CLI transport does not — an asymmetry within one PR signals the CLI side is incomplete.
- **Validation boundary:** existing framing/overflow tests plus a new overflow-then-peer-close race test; live `rpce-cli-debug` saturation/disconnect smoke after the fix.

### Verdict
**REQUEST_CHANGES** — narrowly, on F1 only. The change is otherwise coherent, well-scoped, and well-tested, and CI is fully green. But F1 is neither hypothetical nor stylistic: it claims the PR's central new guarantee (fail-closed on dropped frames) is unenforced on the CLI side under a realistic overflow-plus-peer-close ordering, with a concrete mechanism, exact citations, an in-PR counterexample of the correct pattern, and an identified test gap explaining why CI can't see it. The competing review never engages this code path, so it provides no basis to override the finding. The fix is small; once the CLI gate persists and resolves an authoritative overflow cause (with the race test), this should merge without further rounds. If verification against head `a59c965` shows the gate already persists the overflow cause (i.e., the finding misread `:90-98`/`:607-637`), F1 dissolves and the verdict flips to RECOMMEND_MERGING — that single check is the decision hinge.

morluto:
Thanks @baron — addressed the REQUEST_CHANGES item on head for this push:

- **CLI overflow terminal cause race:** `BootstrapSocketMCPIngressGate` now records the overflow error synchronously on `.dropped` yield (first-writer-wins). `finishStreamIfNeeded` always prefers that authoritative cause over a racing nil/EOF teardown, matching the app transport pattern.
- **Regression:** `testOverflowThenImmediatePeerCloseStillFailsWithOverflow` overflows capacity-1 buffering then immediately closes the peer; stream must still terminate with `BootstrapSocketMCPReceiveBufferOverflowError`.
- **Ledger:** registered the new executable test ID.

**Sentry:** no `sentry[bot]` threads on this PR.

**Validation:** full suite of `Scripts/test_ci_app_test_runner.py` is N/A here; local XCTest cannot run in this Linux agent (no Darwin/macOS SDK). Hosted CI will re-run on the new head. Please re-review when green.

**Deferred (non-blocking):** F2 items (byte-bounded frame queues, compaction benchmark, etc.) left as follow-ups.

morluto:
Follow-up on non-blocking **F2** from the review (blocking overflow-cause fix already on `2b6fac5f`):

Investigated and intentionally deferred these as out-of-band follow-ups rather than expanding this PR:

- **Frame queue bound is count (1,024) not bytes** — count cap still fail-closes via overflow; a byte budget would need a separate capacity model and admission metrics. Prefer a focused follow-up if measured buffer RSS warrants it.
- **Proxy poll 1s → 100ms** — intentional responsiveness change for backpressure; idle wakeup cost is bounded and acceptable for CLI bootstrap.
- **Compaction benchmark / oversized handshake / reconnect-replay tests** — shared accumulator already covers framing correctness; extra benches/tests are valuable canaries but not required for the fail-closed invariant this PR owns.

Happy to open a small follow-up PR for any of these if you want them tracked as issues.

morluto:
Follow-up for @baron and the hosted failures: pushed `b6d33e5b722adec54a0d05bd9881ce92847c5b68` with the two shard fixes. Shard 1 sampled the virtual-token refresh counter while setup refresh work was still active; the test now drains setup refresh tasks before taking its baseline. Shard 2 persisted the selected paths but invoked MCP before explicitly synchronizing that compose tab into `PromptManager`; it now loads the updated tab before asserting bounded admission. Assertions and production contracts are unchanged. Evidence: exact hosted failure logs inspected, both edited Swift files pass repository formatting, diff/whitespace inspection passes, and mandatory commit/push contribution preflights pass. Fresh hosted checks are now authoritative.

morluto:
### Audit disposition — **resource-safety blocker** (2026-08-14)

The per-frame cap and explicit handling of dropped stream yields are worthwhile, but the receive queue is still bounded only by frame count. With 1,024 queued frames and a 256 MiB frame ceiling, the theoretical retained payload is roughly 256 GiB. That does not close the unbounded-memory class described by the title.

Please add a cumulative queued-byte budget with backpressure or immediate connection termination when it is exceeded. The accumulator and silent-drop fixes could be split and merged separately, but I would not merge the overall “bounded buffering” claim until total retained bytes are bounded.

morluto:

**Disposition: block pending an aggregate memory bound.** A 256 MiB per-frame ceiling fixes the unbounded no-newline accumulator, and treating dropped stream yields as fatal is correct. However, the receive queues remain count-bounded at up to 1,024 frames. In the worst case that still permits roughly 256 GiB of queued payload, before object overhead, so the connection remains effectively unbounded by memory.

Please add a cumulative queued-byte budget with explicit backpressure or connection termination when exceeded, and test many large independently owned frames—not only one oversized frame. The cursor accumulator and explicit receive-drop handling could be split and merged separately if useful, but the PR should not claim bounded buffering until both frame and queue bytes are bounded.

PR #458, opened 2026-07-09
fix(agent-mode): stop tool-result updates from remounting unrelated cards

morluto:
## Problem

Agent Mode keeps complete tool-result payloads separately from the compacted transcript so result cards can still show their detailed content.

Those retained payloads previously shared one presentation revision. That revision was also used as the SwiftUI identity for every visible result card. When one retained payload changed, every visible result card received a new identity.

Changing `.id` discards the existing result subtree and its local `@State`. Unrelated cards could therefore lose expansion state and recreate their AppKit text or diff views even though their content had not changed.

## Change

Retained payload revisions now remain keyed by tool-result item ID throughout the presentation path.

Each result card receives only its own revision:

- changing result A refreshes result A;
- result B keeps its existing identity and state.

Payload-only changes still publish a transcript update. Raw payload retention, transcript compaction, persistence, provider behavior, and tool output remain unchanged.

## Impact

Retained payload updates no longer fan out across every visible result card.

This preserves state and backing views for unrelated cards and reduces unnecessary reconstruction during tool-heavy Agent Mode runs.

## Tests

Adds regression coverage with two retained results to verify that:

- updating one result changes only its render identity;
- the other result keeps its previous identity;
- both cards continue receiving the correct retained payload.

### DIFF
--- a/Sources/RepoPrompt/Features/AgentMode/Models/AgentTranscriptModels.swift
+++ b/Sources/RepoPrompt/Features/AgentMode/Models/AgentTranscriptModels.swift
@@ -1008,7 +1008,7 @@ struct AgentTranscriptPresentationSnapshot: Equatable {
     let hydratedBindingTransitionGeneration: UInt64?
     let performanceSnapshot: AgentTranscriptPerformanceSnapshot
     let metadata: AgentTranscriptPresentationMetadata
-    let rawToolResultPayloadRenderRevision: Int
+    let rawToolResultPayloadRenderRevisionByItemID: [UUID: Int]
 
     init(
         tabID: UUID? = nil,
@@ -1028,7 +1028,7 @@ struct AgentTranscriptPresentationSnapshot: Equatable {
         hydratedBindingTransitionGeneration: UInt64? = nil,
         performanceSnapshot: AgentTranscriptPerformanceSnapshot = .empty,
         metadata: AgentTranscriptPresentationMetadata = .empty,
-        rawToolResultPayloadRenderRevision: Int = 0
+        rawToolResultPayloadRenderRevisionByItemID: [UUID: Int] = [:]
     ) {
         self.tabID = tabID
         self.revision = revision
@@ -1047,7 +1047,7 @@ struct AgentTranscriptPresentationSnapshot: Equatable {
         self.hydratedBindingTransitionGeneration = hydratedBindingTransitionGeneration
         self.performanceSnapshot = performanceSnapshot
         self.metadata = metadata
-        self.rawToolResultPayloadRenderRevision = rawToolResultPayloadRenderRevision
+        self.rawToolResultPayloadRenderRevisionByItemID = rawToolResultPayloadRenderRevisionByItemID
     }
 
     func contentEqualsExcludingPerformance(_ other: Self) -> Bool {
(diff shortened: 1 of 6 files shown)


baron (approved):
### Intent Analysis
**Problem:** A single session-wide maximum retained-tool-result payload revision was used as the SwiftUI `.id` input for every visible result card, so updating one payload recreated unrelated card subtrees and could discard their transient UI state.
**Before → After:** Before: `buildTranscriptPresentation` reduced all visible retained-result revisions to one max scalar shared by every card. After: revisions stay keyed by result item UUID (`[UUID: Int]`) through session → projection → presentation snapshot, and each card consumes only `map[item.id] ?? 0`.
**Evidence:** Two independent code-first reviews converged on identical references: per-item projection `AgentModeViewModel.swift:9309-9314`; session publication `AgentModeViewModel.swift:9564-9569`; presentation model/equality `AgentTranscriptModels.swift:1011-1080`; per-card lookup `AgentModeView.swift:2734-2738`; unchanged `.id` consumer `AgentMessageBubble.swift:282-303`; cleanup of keyed revisions `AgentModeViewModel+TabSession.swift:1267-1306, 1520-1529`; two-result regression test `AgentModeViewModelInactiveRefreshTests.swift:58-138`. All 9 CI checks pass (4 app test shards, provider tests, style, Sentry build, secret scan, contributor, secret scan); PR is MERGEABLE at head `1d78301`.
**Confidence:** High — the causal chain is directly visible in the six-file diff, the fix reuses the existing per-item authority (`ephemeralToolResultPayloadRevisionByItemID`) rather than introducing a new one, and the regression test exercises the exact broken invariant.
**Red Flags:** None. No scope creep; no persistence, migration, credential, or concurrency surface changes; dictionary construction/equality is linear over already-enumerated visible results. main is ~24 commits ahead of the tested base, but none touch the changed files.

### Findings
No blocking findings. Non-blocking note: the regression test verifies per-item revision routing at the view-model boundary (`AgentModeViewModelInactiveRefreshTests.swift:58-138`) but does not mount SwiftUI to directly observe preserved `@State` or reduced reconstruction. Since the `.id` consumer (`AgentMessageBubble.swift:282-303`) is unchanged and now receives an exact per-item input, this is an optional follow-up, not a defect.

### Maintainer-guidance check
- **Invariant:** Updating result A must refresh A without resetting unrelated result B; payload-only changes still publish — directly tested.
- **Root cause:** Confirmed — the shared max scalar drove every result card's identity.
- **Authority:** Existing `TabSession.ephemeralToolResultPayloadRevisionByItemID` remains the single authority; the snapshot is a filtered projection.
- **State-safety:** Main-actor session state intact; keyed revisions cleared on removal/cache reset (`AgentModeViewModel+TabSession.swift:1267-1306, 1520-1529`).
- **Scale:** Bounded O(n) visible-item dictionary replaces an O(n) max reduction and eliminates unbounded unrelated remounts.
- **Scope/validation:** Smallest coherent fix at the owning presentation boundary; focused regression plus full green CI is the appropriate validation boundary.

### Verdict
RECOMMEND MERGING. Both independent code-first reports confirm, with matching exact references, that the PR fixes a demonstrated identity fan-out by projecting the existing per-item revision authority, preserves publication and cleanup behavior, and adds targeted regression coverage. No defensible blocking evidence exists; the only noted gap is a hypothetical validation nicety and is not a blocker.

PR #459, opened 2026-07-09
perf(git): bound diff cache retention with LRU eviction

morluto:
## Problem

`GitDiffEngine.shared` previously cached generated diffs in an unbounded dictionary for the lifetime of the application.

Cache keys include repository status, comparison target, scope, and selected paths. Editing files or changing the comparison can create new entries while older repository states remain cached.

An entry may retain the diff text twice: once as the complete unified diff and again across its per-file strings. Long sessions involving large or frequently changing diffs could therefore retain increasing amounts of obsolete diff data.

## Change

The cache now uses deterministic LRU retention with two initial internal limits:

- 32 entries;
- 64 MiB of retained UTF-8 payload.

Cache hits update recency. The least-recently-used entries are removed when either limit is exceeded.

Accounting includes cache keys, fingerprints, complete diff text, and per-file diff strings. Swift object and collection overhead is not included in the byte total, while the entry limit also bounds growth from many small entries.

Results larger than the byte budget are returned normally but are not cached. `useCache: false` continues to bypass both lookup and admission.

DEBUG metrics expose entry count, retained bytes, hits, misses, bypasses, admissions, replacements, evictions, and oversized rejections. These metrics allow the initial limits to be evaluated and adjusted from observed behavior.

## Impact

Recent diffs still benefit from caching, but obsolete repository states no longer accumulate without limit.

Eviction changes only reuse behavior. It does not change generated diff content, Git commands, branch selection, or snapshot retention.

## Tests

Adds focused coverage for:

- count and byte limits;
- LRU recency after a hit;
- complete and per-file payload accounting;
- replacement and oversized-entry behavior;
- overflow-safe accounting;
- cache bypass;
- concurrent actor requests.


baron (changes requested):
### Intent Analysis
**Problem:** `GitDiffEngine` retained generated diffs in an unbounded, process-lifetime dictionary; distinct repository states/selections create distinct keys, so obsolete diff payloads accumulated for the life of the process.
**Before → After:** Unbounded dictionary → actor-isolated LRU bounded at 32 entries / 64 MiB counted UTF-8 payload. Hits refresh recency; overflow evicts LRU-first; oversized results are returned but not retained. Additionally, `useCache: false` changes from "skip lookup but still admit/replace the recomputed value" to "skip both lookup and admission" — a silent behavioral change to a production-relevant flag.
**Evidence:** `Sources/RepoPrompt/Infrastructure/VCS/GitDiff/GitDiffEngine.swift:5-20, 39-220, 475-727` (admission/eviction accounting :117-156; bypass sites :484-486, :523-525, :556-558, :707-709, :723-725); `Sources/RepoPrompt/Infrastructure/FileSystem/LRUCache.swift:67-76, 102-138`; `Tests/RepoPromptTests/Services/VCS/GitDiffEngineCacheTests.swift:7-196`. Two independent code-first reviews converge on these references.
**Confidence:** High on intent and mechanism (both reviews agree on structure, limits, accounting, and the changed bypass semantics). Medium-high on the blocking finding: only one review traced it, but its chain is fully cited and the other review documents the same semantic change without contradicting any link.
**Red Flags:** No linked issue or prior review discussion. The new bypass semantics are covered only by a test that asserts the new behavior without exercising its consequence. The 32/64 MiB limits are empirically unvalidated (non-blocking).

### Findings
**Blocking — force refresh can strand a stale cache entry that later ordinary requests serve.**
`GitDiffEngine.swift:504-558, 576-629, 689-725` admit recomputed results only when `useCache` is true. The production caller maps force refresh directly onto this flag (`GitStatusActor.swift:617-651`, `useCache: !forceRefreshSnapshot`). Concretely reachable via Jujutsu: `JujutsuBackend.swift:261-279` derives `statusHash` from summary text (not content hashes), and the uncommitted cache key excludes `fingerprint.headSHA` (`GitDiffEngine.swift:45-52, 496-503`). So editing the content of an already-modified file preserves the key: a forced refresh returns the fresh diff but leaves the old resident entry intact, and the next ordinary request serves the stale diff — into prompts, token counts, clipboard, or MCP context. Pre-PR code admitted the recomputed value at exactly this point, so this is a regression, not a pre-existing gap. The new test (`GitDiffEngineCacheTests.swift:97-132`) asserts bypass-without-admission but recomputes identical content and never follows with a normal lookup, so it encodes the regression rather than catching it. All eight green checks on head `1d72290b` are real but none cover this sequence.
**Fix (narrow):** decouple lookup-bypass from admission — on `useCache: false`, still admit/replace (restoring prior semantics) or invalidate the matching key after forced recomputation. Add one regression test: cached value → same-key changed content → forced refresh → ordinary request returns the refreshed diff.

**Non-blocking:** concurrency test (`GitDiffEngineCacheTests.swift:165-196`) uses identical requests (acceptable under actor isolation); health counters are DEBUG-only (`GitDiffEngine.swift:60-86, 182-195, 218-222`); byte accounting measures retained payload, not allocator overhead (intentional, coherent).

### Maintainer-guidance check
- **User impact/invariant:** bound obsolete diff retention without changing which diff users/prompts/MCP receive; a force refresh must not be undone by the next cached request. The blocking finding violates the second clause.
- **Root-cause confidence:** unbounded retention confirmed by both independent reviews; the stale-after-refresh regression is directly supported by cache key + backend fingerprint + caller path. One review missed it by evaluating the cache in isolation; the other traced the full chain — absence of investigation, not a refutation.
- **Authority:** correct — `GitDiffEngine` remains sole cache owner; `LRUCache` holds generic recency mechanics; backends own fingerprints; `GitStatusActor` owns force-refresh semantics.
- **State-safety:** no durable-state/migration risk; entries are derived and disposable. Residual risk is ephemeral-but-visible stale diff context.
- **Recommended scope:** keep the bounded-LRU design as-is; repair admission on forced recomputation; add the one regression test. Limit tuning and production observability remain follow-ups.
- **Validation boundary:** `make dev-test FILTER=GitDiffEngineCacheTests` plus the new same-key refresh regression; existing root CI otherwise adequate. Maintainer spot-check of `GitStatusActor.swift:617-651` and the uncommitted key construction confirms the chain in minutes.

### Verdict
**REQUEST_CHANGES.** The bounded-cache design is focused, correctly scoped, and mechanically well tested, but the changed `useCache: false` semantics are a traced correctness regression: a force refresh can be silently undone by the next cached read under same-key content change (concretely reachable via the Jujutsu fingerprint path), and the accompanying test asserts the regressive behavior rather than guarding against it. Restore admission (or invalidate) on forced recomputation and add the four-step regression test; with that narrow fix this is a clean approve.

morluto:
Thanks @baron — addressed the REQUEST_CHANGES item and the actionable non-blocking note on `199550db926e3be0295d1c2729926ec5688fe55e`:

**Blocking**
- **Force refresh no longer strands stale same-key entries:** `useCache: false` still bypasses lookup (and records bypass), but every recomputed result is admitted/replaced into the LRU. Restores pre-bounded-cache force-refresh semantics while keeping entry/byte caps.
- **Tests:** bypass asserts admission+replacement; same-key unit path (stale → force admit → ordinary read); end-to-end edit → force refresh → ordinary read.
- **Ledger:** renamed/added executable IDs surgically.

**Non-blocking**
- **Concurrency test used identical requests:** updated to race distinct scopes/paths (all + selected A/B/C) so multi-key cache traffic is exercised under actor isolation, still asserting entry/byte bounds.
- **DEBUG-only health counters / payload-byte accounting:** left as intentional design (diagnostics + retained-payload budget, not allocator RSS). Happy to follow up if you want production-facing metrics.

**Sentry:** none on this PR.

**Validation:** local XCTest cannot run on this Linux agent (no Darwin/macOS SDK). Hosted CI will re-run on the new head. Ready for re-review when green.

morluto:
Follow-up for @baron: the blocking force-refresh admission fix and the actionable non-blocking multi-key concurrency coverage remain on head, and commit `d1dda68062e20e2b239e7765f4d6e3c6c2ed227a` fixes the hosted Style failure by applying the repository SwiftFormat rules to the two PR-owned files (`LRUCache.swift` compact conditionals and cache-test trailing commas). Validation: targeted SwiftFormat lint over all three changed Swift files passes; commit and push contribution preflights pass. The pushed SHA is visible on `perf/bounded-git-diff-cache`; hosted checks are rerunning.

baron:
Current-main refresh on exact head `4dfadaf3bf888f36c28c272c90748fd5efcae3a2`: the force-refresh admission fix now looks code-correct and current main still has an unbounded cache, but the branch conflicts and its latest shard 2 is red outside the focused cache path. Please rebase, omit obsolete ledger integration, rerun focused cache/current-head CI, and request re-review; the older change request should not be treated as an exact-head approval.

morluto:
### Audit disposition — **revise the cache-mode contract** (2026-08-14)

The deterministic LRU structure and explicit count/byte accounting look coherent. The remaining concern is the public meaning of `useCache: false`: the PR description promises bypass of both lookup and admission, while the reviewed flow appears capable of behaving like a forced refresh that replaces or admits a result.

Please make bypass versus refresh explicit in the API/name, add caller-level tests proving the intended lookup/admission behavior, and then rebase onto current main. I do not see a fundamental objection to bounded LRU retention once that contract is unambiguous.

morluto:

**Disposition: hold for cache-bypass contract verification.** The deterministic LRU, entry/byte limits, oversized-result rejection, and overflow-safe accounting are sensible. The point that needs proof is `useCache: false`: the PR body promises bypass of both lookup and admission, but the implementation/caller behavior should be verified end to end so a forced refresh cannot replace or admit an entry.

Please add a caller-level regression that observes cache contents/metrics before and after `useCache: false`, including an existing key, an oversized result, and concurrent requests. If bypass truly performs neither lookup nor admission on the final head, I did not find another blocker; otherwise rename the mode or correct the implementation.

baron:
Updated the existing PR at `5ae0b5c37af2c33d2789c86b35ad9940ff420f7a` with a normal merge of current main (`4232fc4b`), without rewriting published history.

- Renamed the lookup flag to `allowCachedResult` and corrected the description: disabling lookup forces recomputation followed by bounded admission/replacement, not no-store behavior.
- Added discriminating same-fingerprint coverage through `GitStatusActor`: same-length edit, restored mtime, equal fingerprint, forced refresh, then an ordinary cache hit with fresh content. Oversized refresh must remove the stale entry and make the next ordinary request recompute.
- Preserved main's deletion of the retired test ledger; GitHub now reports MERGEABLE.

Independent native Astra High source review approved this exact commit with no must-fix findings. Scoped SwiftFormat/strict SwiftLint and mandatory commit/push safety gates passed. Focused local tests and the product build remain queued behind shared conductor admission; [exact-head CI](https://github.com/repoprompt/repoprompt-ce/actions/runs/35716971530) is being watched. Keeping this draft until executable validation is complete; the old change request is not being treated as an approval.

baron (approved):
Re-reviewed head 8d107c00efce8ac3dae39be7caa4b5e692792ff0. The July requested-changes finding is addressed: a forced refresh bypasses lookup but admits/replaces its recomputed result, and the caller-level same-fingerprint and oversized regressions cover the subsequent ordinary read. All eight exact-head hosted checks passed. Approved for merge review.

PR #473, opened 2026-07-10
Fix Codex CLI resolution for nvm / .zshrc-only setups (#471)

### LINKED ISSUE #471: Codex resolution misses nvm-installed CLI (non-interactive shell skips .zshrc)
## What
RepoPrompt CE fails to resolve a user's actual `codex` CLI when the toolchain that provides it (nvm, plus the Homebrew/PATH setup nvm depends on) is set up only in `~/.zshrc`. Codex resolution captures the shell environment with a **non-interactive login shell**, which does not source `~/.zshrc`, so for the common "nvm only in `.zshrc`" setup the captured PATH never gains the nvm version bin. RepoPrompt then either resolves a stale bundled binary (older Codex.app) or, after that binary is removed, reports "Codex CLI executable was not found." When a stale binary is launched, newer models such as `gpt-5.6-sol` are rejected by the OpenAI Responses API with HTTP 400 `invalid_request_error`: *"The 'gpt-5.6-sol' model requires a newer version of Codex."* — even though `codex --version` in the user's terminal reports the latest.

## Repro / context
1. Install codex via npm under nvm (binary at `~/.nvm/versions/node/<ver>/bin/codex`); load nvm (and set `HOMEBREW_PREFIX`, custom PATH) only in `~/.zshrc` (not in `.zprofile`/`.zshenv`).
2. Optionally have an older `/Applications/Codex.app` present.
3. Connect Codex in RepoPrompt CE; select a newer model (e.g. `gpt-5.6-sol`).
4. **Observed:** with Codex.app present -> API 400 *"requires a newer version of Codex"* (stale binary launched); after removing/renaming the bundled app -> *"Codex CLI executable was not found."*

**Expected:** RepoPrompt resolves the same `codex` the user's interactive shell resolves. **Actual:** it resolves an older bundled binary or none, because `.zshrc`-only toolchain setup is absent from the non-interactive capture.

## Root cause
`Sources/RepoPrompt/Infrastructure/Process/ProcessEnvironmentBuilder.swift` -> `preferredShellCaptureMode(for:)` uses a **non-interactive** login shell for Codex:

```swift
case .codexAppServer, .codexPreflight:
    .loginShell          // `zsh -l -c env` — does not source ~/.zshrc
default:
    .interactiveLoginShell
```

zsh sources `~/.zshrc` **only for interactive shells**; a non-interactive login shell reads `.zprofile`/`.zlogin`/`.zshenv` but not `.zshrc`. In the affected setup the *entire* toolchain lives in `~/.zshrc` (`HOMEBREW_PREFIX`, the custom `PATH`, and nvm via `${HOMEBREW_PREFIX}/opt/nvm/nvm.sh`). It would still be picked up if it lived in `.zprofile`/`.zshenv` or were inherited, but for the common "nvm only in `.zshrc`" case it is absent from the captured PATH. This applies generally to any CLI whose setup is `.zshrc`-only (nvm/pyenv/Homebrew-only tools).

Two compounding factors prevent recovery:

1. **Static hints don't cover nvm.** `CLILaunchProfile.codexSupplementalSearchPaths` (plus `CLINativePathDefaults` and the `fallbackPathCandidates` PATH enrichment) cover Homebrew / Volta / pnpm / Yarn / Bun / Cargo / mise / asdf / nodenv / pyenv / rbenv / `~/.local/bin` / `~/bin` / `~/go/bin` / `/Applications/Codex.app/Contents/Resources` — **not** nvm's version-specific bin.
2. **The interactive fallback is last-resort and can be preempted.** `CodexProviderHelpers.resolveCodexExecutable` resolves with `shellLookupMode: .fallbackOnly`. In `CommandPathResolver.resolve`, the order is: explicit path -> `locateInSearchPaths` (env `PATH`, then hints) -> (only on miss) interactive shell lookup. So while `/Applications/Codex.app/Contents/Resources` exists, it is found in the hint scan **before** the interactive lookup runs, shadowing it. (Env `PATH` is scanned before hints, so if the captured PATH *did* contain the nvm bin, nvm would win — it does not in this setup.)

The interactive fallback (`CommandPathResolver.lookupViaInteractiveShell`, `zsh -l -i -c 'command -v codex'`, stdin `/dev/null`, `shellLookupTimeout = 5s`) sources `~/.zshrc` and **would typically** resolve nvm's codex (measured ~0.75s in this setup). It is not guaranteed: `.zshrc` may guard on non-TTY or on the `RP_SHELL_LOOKUP=1` marker the resolver sets, stdin is `/dev/null` with no pty, and the resolver may issue up to four queries (each with its own 5s timeout). The one code-supported reason it is *not reached at all* is hint preemption while Codex.app exists. Why it also fails to rescue resolution after Codex.app is removed (-> "not found") is **not established by the code** and needs runtime logs/reproducer — currently a hypothesis (e.g., cold first-spawn latency, a `.zshrc` non-TTY guard, or the Connect flow's resolution path), not a confirmed cause.

### Scope (not Finder-only)
`ProcessEnvironmentBuilder.shouldUseInheritedEnvironment` returns `false` for `.codexAppServer`/`.codexPreflight` even when RepoPrompt is terminal-launched, so Codex **always** re-captures via `.loginShell`. A Finder launch makes the failure more likely (sparser `baseEnv`) but is not the defining condition.

### Why `which codex` looks fine
The user's terminal is an interactive shell that sources `~/.zshrc`, so `which codex` -> newest. RepoPrompt's Codex capture is non-interactive, so its view of PATH differs.

## Evidence
- Version numbers and the API rejection are from direct measurement on the affected machine (not from the reviewed source files): terminal `which codex` -> 0.144.1; `/Applications/Codex.app/Contents/Resources/codex` -> 0.142.5; selecting `gpt-5.6-sol` returned HTTP 400 *"requires a newer version of Codex."*
- Call-sites routing through `resolveCodexExecutable`: `CodexAppServerClient.startProcess` and the preflight behind the API Settings "Connect" flow (`CodexProviderHelpers.preflightCodexExecutable`).
- Replicated the capture condition (GUI-app minimal PATH via `env -i`, stdin `/dev/null`):
  - `zsh -l -c 'command -v codex'` (non-interactive login — Codex's capture) -> **not found**; `HOMEBREW_PREFIX` unset.
  - `zsh -lic 'command -v codex'` (interactive login — the fallback) -> resolves the newest nvm binary; `HOMEBREW_PREFIX=/opt/homebrew`.
  - Interactive fallback latency ~0.72–0.76s over 5 runs; non-interactive capture ~0.01s.
- All toolchain setup referenced only in `~/.zshrc` (`HOMEBREW_PREFIX`, custom PATH, nvm via Homebrew).
- nvm's version bin is confirmed absent from the **static hints and PATH-enrichment candidates**; it is absent from the **captured PATH** specifically for the nvm-only-in-`.zshrc` case (not in general).

## Proposed fix
A **Codex-specific** change is the minimal fix; modifying shared `CommandPathResolver` behavior is not required and risks other-provider regressions. Options:
- Give Codex the interactive capture: map `.codexAppServer`/`.codexPreflight` to `.interactiveLoginShell` in `preferredShellCaptureMode`.
- Or make Codex resolution prefer the shell: switch `resolveCodexExecutable` to `shellLookupMode: .preferShell`.
- Or stop letting a bundled hint preempt the shell lookup for Codex (reorder ChatGPT.app/Codex.app resources after the interactive lookup, or pick the highest-version codex among candidates).

Investigation points: `ProcessEnvironmentBuilder.swift` (`preferredShellCaptureMode`, `shouldUseInheritedEnvironment`), `CodexProviderHelpers.swift` (`resolveCodexExecutable`), `CommandPathResolver.swift` (`resolve`, `lookupViaInteractiveShell`). Validate with the existing Codex integration tests plus a live CE MCP smoke.

## Workarounds (until fixed)

> **Heads-up — `node` is required too.** The npm/nvm `codex` is a `#!/usr/bin/env node` **shim**, not a standalone binary. RepoPrompt finding `codex` is not enough: it must also resolve `node` to exec the shim, otherwise the codex process never starts and you get *"app-server process isn't running."* So any workaround below must make `node` reachable as well. (Symlinking only `codex` reproduces that error: `env: node: No such file or directory`.)

- **Symlink the nvm bin — including `node` — into a hint-covered dir** (`~/.local/bin` is in RepoPrompt's search hints). Link `codex` **and** `node`/`npm`/`npx`; codex alone is insufficient:
  ```sh
  NVM_BIN="$HOME/.nvm/versions/node/v23.5.0/bin"   # adjust to your Node version
  mkdir -p ~/.local/bin
  for b in codex node npm npx; do ln -sf "$NVM_BIN/$b" "$HOME/.local/bin/$b"; done
  ```
  Re-run if you change Node versions.
- **Install codex via Homebrew** (`brew install codex` -> `/opt/homebrew/bin/codex`, covered by `homebrewBins`). The Homebrew package does not go through the nvm node-shim, so it does not require `node` on RepoPrompt's PATH.
- **Move the nvm/PATH (and `HOMEBREW_PREFIX`) setup into `~/.zprofile`** (sourced by non-interactive login shells, which is what RepoPrompt captures today). This puts the entire nvm bin — `codex` **and** `node` — onto RepoPrompt's captured PATH at once, and is the closest manual equivalent of the fix in the linked PR:
  ```sh
  # ~/.zprofile
  export HOMEBREW_PREFIX="/opt/homebrew"
  export NVM_DIR="$HOME/.nvm"
  [ -s "$HOMEBREW_PREFIX/opt/nvm/nvm.sh" ] && . "$HOMEBREW_PREFIX/opt/nvm/nvm.sh"
  ```

*Verified 2026-07-13 against prod 1.0.28 + codex 0.144.3 on an nvm host: the node-inclusive symlink makes the codex app-server start and stay up under a RepoPrompt-like (non-interactive) PATH; a codex-only symlink fails with `env: node: No such file or directory`.*

## Acceptance criteria
- [ ] RepoPrompt resolves an nvm-installed `codex` even when nvm/Homebrew setup is in `~/.zshrc` only.
- [ ] No regression in Codex app-server launch latency for the common (already-on-PATH) case.
- [ ] Regression test covers "CLI visible only via interactive-shell PATH is still resolved".

## Open questions
- Why does the last-resort interactive fallback not rescue resolution after the bundled app is removed? Needs a runtime reproducer (capture a `CodexPreflight` debug log showing the resolved command + status) rather than further code inference.

## Detail / links
- Code: `Sources/RepoPrompt/Infrastructure/Process/ProcessEnvironmentBuilder.swift`, `Sources/RepoPrompt/Infrastructure/Process/CLIEnvironmentCache.swift`, `Sources/RepoPrompt/Infrastructure/Process/CLILaunchProfile.swift`, `Sources/RepoPrompt/Infrastructure/Process/CLINativePathDefaults.swift`, `Sources/RepoPrompt/Infrastructure/Process/CommandPathResolver.swift`, `Sources/RepoPrompt/Infrastructure/AI/Providers/Codex/Shared/CodexProviderHelpers.swift`

## Source
Reported by user 2026-07-09; root cause isolated against the live machine. Body verified and tightened 2026-07-10 (claims checked against source by an independent review pass). Filed by a fork maintainer (local fork `moonray/repoprompt-ce`).

moonray:
## What
Fixes Codex CLI resolution when `codex` is installed via **nvm** — or any toolchain set up only in `~/.zshrc`. RepoPrompt previously captured a **non-interactive** login shell for Codex, which does not source `~/.zshrc`, so it could not see an nvm-installed `codex`. It fell back to a stale bundled `Codex.app` binary (whose older version the OpenAI Responses API rejects for newer models such as `gpt-5.6-sol` — *"requires a newer version of Codex"*) or reported *"Codex CLI executable was not found"*.

Closes #471.

## Root cause
`ProcessEnvironmentBuilder.preferredShellCaptureMode` mapped `.codexAppServer` / `.codexPreflight` to `.loginShell` (non-interactive). zsh sources `~/.zshrc` only for **interactive** shells, so the captured PATH missed the nvm version bin (and `HOMEBREW_PREFIX` / the custom PATH when those are `.zshrc`-only). The supplemental hints do not cover nvm's version-specific bin, so a stale `/Applications/Codex.app` hint shadowed the interactive shell fallback. Full analysis and replication evidence are in #471.

## Change
Map `.codexAppServer` / `.codexPreflight` → `.interactiveLoginShell` so the captured PATH includes `.zshrc`-only toolchain setup. The capture is **cached per mode** by `CLIEnvironmentCache`, so this does **not** spawn a shell per launch — Codex now shares the interactive cache slot every other provider already uses. Codex was the only `.loginShell` producer, so the mapping collapses to a single returned mode; the `.loginShell` enum case and its `CLIEnvironmentCache` capture branch are retained (dormant) to keep blast radius small.

`preferredShellCaptureMode` is made `internal` for direct unit testing.

## Tests
- `ProcessEnvironmentBuilderCaptureModeTests` — `.codexAppServer` / `.codexPreflight` map to `.interactiveLoginShell` and share the cache slot other providers use.
- `CodexProviderHelpersResolutionTests` — a codex on the captured PATH resolves, and **outranks** a stale supplemental-hint codex (env PATH is searched before hints).
- `test-suite-contract-ledger.tsv` — rows added for the two new suites (CI `--strict-ledger` reconciles at the suite level).

## Validation
- `make dev-test` — 5/5 new tests green.
- `make dev-lint` — clean (0/1433 files).
- `make dev-swift-build PRODUCT=repoprompt-mcp` — exit 0.
- Contribution preflight (`commit` + `push`) — passed.

## Risk / rollback
- Codex-only capture change; every other provider already used interactive capture (unaffected).
- Operational risks (accepted): cold-start / force-refresh interactive-spawn latency; arbitrary `~/.zshrc` side effects; shared interactive-cache invalidation coupling — all bounded by the existing 5s capture timeout + 2s grace and the `pathLooksInsufficient` retry.
- Rollback: revert this commit and invalidate/restart the shell-environment cache (a cached interactive snapshot may persist until then). Rollback restores the known nvm resolution failure.

## Known limitations (intentional)
- No fresh non-interactive capture on interactive-capture failure (existing graceful fallback to cached/enriched env + hints remains). Narrowed scope; a retry is a welcome follow-up.
- End-to-end smoke (connect Codex on an nvm host, confirm `model/list`) is host-dependent and not CI-gated; recommended as manual verification.

🤞 Spec, deep plan, and conformance matrix were authored for this change (kept local per repo guardrails); details in #471.


baron (approved):
Reviewed exact head `60d136a4e869502009864b63bbaf26715ee4e993`. The interactive login-shell capture is cached, preserves deterministic fallback behavior, and directly covers the `.zshrc`/nvm resolution failure from #471. The focused resolution tests and exact-head hosted checks are green, and I found no blocking code issue. Approving. Before merge, refresh current-base hosted checks and perform the documented nvm/Codex live smoke.

baron:
Current-main recheck on exact head `60d136a4e869502009864b63bbaf26715ee4e993`: the approved Codex shell-resolution change still has merit, but the branch now conflicts only around the test-suite ledger fixture deleted on main. Please refresh while retaining main's deletion and rerun CI. Because resolving the conflict changes the head/integration, the approval should then be reconfirmed on the new exact head before merge.

morluto:
### Audit disposition — **rebase plus host-dependent smoke** (2026-08-14)

Using the existing interactive-login-shell capture makes nvm/`.zshrc` resolution consistent with the other providers and is a reasonable fix. The accepted trade-offs—arbitrary startup-file effects, cold capture latency, and cache coupling—need current-head operational validation rather than more abstraction.

Please rebase and test nvm, asdf, Homebrew, a noisy/slow `.zshrc`, capture timeout/fallback, and cache invalidation. A real `model/list` connection on an nvm host should be the final gate. I did not find a static correctness blocker.

morluto:

**Disposition: rebase and complete host-level validation.** Mapping Codex to the same interactive-login-shell capture used by other providers is a reasonable fix for nvm/`.zshrc`-only PATHs, and the existing timeout/cache bounds limit the operational risk. The remaining risk is that interactive startup files execute arbitrary user code, print noise, or block.

Please validate nvm, asdf, Homebrew, a noisy `.zshrc`, and a deliberately slow/failing shell; confirm output parsing cannot treat startup text as environment data and that a failed refresh preserves a known-good cached environment. With those current-head checks and a real `model/list` smoke on an nvm host, I did not find a code blocker.

PR #485, opened 2026-07-10
fix(workspaces): prevent ephemeral workspace data from reaching disk

### LINKED ISSUE #468: Temporary workspaces are written to disk but never indexed or cleaned up
## Summary

RepoPrompt marks some workspaces as temporary (`ephemeral`). These workspaces are supposed to exist only in memory.

The initial creation path correctly avoids saving them, but later autosave paths can still create a workspace directory and `workspace.json`. Temporary workspaces are excluded from the workspace index, so these files become orphaned and are not managed or cleaned up normally.

## Impact

Aggregate evidence from one RepoPrompt CE debug installation:

- 3 workspaces in `workspacesIndex.json`
- 3,576 `workspace.json` files on disk
- 3,573 of those files marked temporary
- 3,571 referring to temporary repository roots
- 10,184 directories under workspace storage
- 91 MB of workspace storage
- 56 temporary workspace files updated during one full test-suite run

No workspace names, paths, IDs, or contents are included in this report.

This suggests temporary Agent, worktree, or test flows can leave permanent files behind. It also means tests may write into the real RepoPrompt CE Application Support directory.

## Expected behavior

A temporary workspace should never create or update:

- a workspace directory
- `workspace.json`
- a workspace-index entry
- workspace sidecar data

It should remain entirely in memory unless explicitly converted into a persistent workspace.

## Why this happens

Temporary-workspace handling is inconsistent:

- `createWorkspace(..., ephemeral: true)` skips the initial disk write.
- Workspace-index rebuilding excludes temporary workspaces.
- The regular autosave path does not check whether the active workspace is temporary.
- The 30-second workspace poll can therefore save a dirty temporary workspace.

The result is a workspace file that exists on disk but is intentionally absent from the index.

Relevant area:

- `Sources/RepoPrompt/Features/Workspaces/WorkspaceModel.swift`
- `Sources/RepoPrompt/Features/Workspaces/ViewModels/WorkspaceManagerViewModel.swift`

## Minimal reproduction

Use an isolated temporary storage root:

1. Create a temporary workspace and make it active.
2. Change its tab, prompt, or selection state.
3. Mark it dirty.
4. Run `pollAndSaveStateAsync`.
5. Inspect the isolated storage root.

Current result: a workspace directory and `workspace.json` can be created.

Expected result: no persistent files are created.

## Acceptance criteria

- Autosave and direct-save paths refuse to persist temporary workspaces.
- Temporary workspaces remain absent from `workspacesIndex.json`.
- A dirty temporary workspace can pass through the 30-second poll without creating files.
- Tests cover both poll-driven and direct-save attempts.
- Tests use isolated storage and never touch the real Application Support directory.
- Persistent-workspace saving and cross-window synchronization continue to work.
- Cleanup of existing orphaned directories is handled separately, with conservative ownership checks.

## Related

#381 proposes an overrideable Application Support root for isolated testing. Better isolation would reduce the impact, but the product should still enforce the rule that temporary workspaces are never persisted.

morluto:
## Problem

Ephemeral workspaces are meant to exist only for the current app session, but their creation previously entered the normal persistent-workspace flow before the workspace was marked ephemeral. That allowed directory creation or a save to be scheduled during the transition.

Lower-level APIs could also write workspace-owned sidecars from a workspace model or storage URL without first proving that persistence was allowed. A temporary workspace could therefore leave chats, attachments, agent sessions, Git metadata, or merge artifacts on disk.

Fixes #468.

## Approach

Treat ephemeral behavior as a storage contract rather than a creation-time flag.

The workspace is now marked ephemeral before creation can schedule persistence. Lower-level storage paths require an authorized `WorkspacePersistentStorage` capability, and the writer checks authoritative workspace state again immediately before writing. Creation-time checks alone would not cover later sidecar writes or work already queued by a stale model.

## Changes

- Mark ephemeral workspaces before any save can be scheduled.
- Centralize persistence disposition and the typed `ephemeralWorkspace` error.
- Restrict construction of workspace-owned persistent storage to manager-authorized paths.
- Reject ephemeral persistence before directory creation, disk reads, JSON encoding, writer enqueue, normalization writeback, and atomic replacement.
- Discard queued writes when the authoritative workspace becomes ephemeral.
- Protect chats, agent sessions, attachments, `_git_data`, Prompt/MCP Git artifacts, and worktree merge artifacts.
- Preserve persistent workspace save and reload behavior.

## Performance effect

Ephemeral workspaces now stop before entering the persistence pipeline. They no longer perform workspace-owned directory creation, disk reads, JSON encoding, writer enqueueing, normalization writeback, or atomic replacement.

The same rule prevents chats, attachments, agent sessions, and Git artifacts from scheduling filesystem work for an ephemeral workspace. The result is zero workspace-owned persistence for ephemeral workspaces, removing unnecessary serialization and filesystem I/O.

## Testing

Regression coverage includes creation and save races, queued writes, stale persistent references, normalization, sidecar storage, attachments, Git artifacts, and persistent workspace reload behavior.

- `WorkspaceEphemeralPersistenceTests`: 13 passed

## Not included

- Removing files left behind by older versions
- Adding a cleanup migration

### DIFF
--- a/Sources/RepoPrompt/Features/AgentMode/Runtime/AgentSessionDataService.swift
+++ b/Sources/RepoPrompt/Features/AgentMode/Runtime/AgentSessionDataService.swift
@@ -858,6 +858,7 @@ actor AgentSessionDataService {
         for workspace: WorkspaceModel,
         mode: AgentSessionMetadataIndexLoadMode
     ) async throws -> AgentSessionMetadataIndex? {
+        guard workspace.persistenceDisposition == .persistent else { return nil }
         let folder = try ensureAgentSessionsFolder(for: workspace)
         switch mode {
         case .fast:
@@ -874,6 +875,7 @@ actor AgentSessionDataService {
     }
 
     func fastMetadataRecordsIfAvailable(for workspace: WorkspaceModel) async throws -> FastMetadataRecordsResult? {
+        guard workspace.persistenceDisposition == .persistent else { return nil }
         let folder = try ensureAgentSessionsFolder(for: workspace)
         let key = canonicalMetadataFolderKey(folder)
         if let cached = metadataIndexCacheByFolder[key] {
@@ -912,6 +914,7 @@ actor AgentSessionDataService {
     }
 
     func metadataRecordForSessionID(_ id: UUID, for workspace: WorkspaceModel) async throws -> AgentSessionMetadataRecord? {
+        guard workspace.persistenceDisposition == .persistent else { return nil }
         let folder = try ensureAgentSessionsFolder(for: workspace)
         let fileURL = agentSessionFileURL(id: id, in: folder)
         guard FileManager.default.fileExists(atPath: fileURL.path) else {
@@ -940,6 +943,7 @@ actor AgentSessionDataService {
     }
 
     func sidebarStreamMetadataRecords(for workspace: WorkspaceModel) async throws -> [AgentSessionMetadataRecord] {
+        guard workspace.persistenceDisposition == .persistent else { return [] }
         let folder = try ensureAgentSessionsFolder(for: workspace)
         if let index = await readMetadataIndexIfAvailable(folder: folder) {
             scheduleMetadataIndexReconciliationIfNeeded(
@@ -963,6 +967,9 @@ actor AgentSessionDataService {
         preparation: SavePreparation = .canonicalize,
         trustedCanonicalItemCount: Int? = nil
     ) async throws -> URL {
+        guard workspace.persistenceDisposition == .persistent else {
+            throw WorkspacePersistenceError.ephemeralWorkspace
+        }
         let agentSessionsFolder = try ensureAgentSessionsFolder(for: workspace)
 
         let filename = "AgentSession-\(session.id.uuidString).json"
@@ -1140,6 +1147,7 @@ actor AgentSessionDataService {
 
     /// Returns a list of AgentSession files in the workspace's AgentSessions folder, sorted by mod date desc.
     func listAgentSessions(for workspace: WorkspaceModel) async throws -> [URL] {
+        guard workspace.persistenceDisposition == .persistent else { return [] }
         let agentSessionsFolder = try ensureAgentSessionsFolder(for: workspace)
 
         let contents = try FileManager.default.contentsOfDirectory(
@@ -1291,6 +1299,7 @@ actor AgentSessionDataService {
 
     /// Load an agent session by ID without scanning the session directory.
     func loadAgentSession(id: UUID, for workspace: WorkspaceModel) async throws -> AgentSession? {
+        guard workspace.persistenceDisposition == .persistent else { return nil }
         let agentSessionsFolder = try ensureAgentSessionsFolder(for: workspace)
         let filename = "AgentSession-\(id.uuidString).json"
         let fileURL = agentSessionsFolder.appendingPathComponent(filename)
@@ -1457,6 +1466,9 @@ actor AgentSessionDataService {
 
     /// Creates (if needed) and returns the "AgentSessions" subfolder for the given workspace.
     private func ensureAgentSessionsFolder(for workspace: WorkspaceModel) throws -> URL {
+        guard workspace.persistenceDisposition == .persistent else {
+            throw WorkspacePersistenceError.ephemeralWorkspace
+        }
         let baseFolder = try workspaceFolderURL(for: workspace)
         let agentSessionsFolder = baseFolder.appendingPathComponent("AgentSessions")
 
--- a/Sources/RepoPrompt/Features/AgentMode/ViewModels/AgentModeViewModel+WorktreeMerge.swift
+++ b/Sources/RepoPrompt/Features/AgentMode/ViewModels/AgentModeViewModel+WorktreeMerge.swift
@@ -423,7 +423,9 @@ extension AgentModeViewModel {
         )
         try await requireAuthorizedWorktreeMergeEndpoint(targetEndpoint, roots: authorizedRoots, label: "Target")
         let directory = workspaceDirectory
-            ?? workspaceManager?.activeWorkspace?.customStoragePath
+            ?? workspaceManager.flatMap { manager in
+                manager.activeWorkspace.map { manager.featureArtifactDirectory(for: $0) }
+            }
             ?? FileManager.default.temporaryDirectory
         let preview = try await VCSService.shared.previewGitWorktreeMerge(.init(
             source: source,
--- a/Sources/RepoPrompt/Features/Chat/Services/ChatHistoryManager.swift
+++ b/Sources/RepoPrompt/Features/Chat/Services/ChatHistoryManager.swift
@@ -120,6 +120,9 @@ actor ChatDataService {
         _ session: ChatSession,
         for workspace: WorkspaceModel
     ) async throws -> URL {
+        guard workspace.persistenceDisposition == .persistent else {
+            throw WorkspacePersistenceError.ephemeralWorkspace
+        }
         // 1) Get "Chats" folder
         let chatsFolder = try ensureChatsFolder(for: workspace)
 
@@ -282,6 +285,7 @@ actor ChatDataService {
 
     /// Returns a list of "ChatSession-xxx.json" files in the workspace’s Chats folder, sorted by mod date desc.
     func listChatSessions(for workspace: WorkspaceModel) async throws -> [URL] {
+        guard workspace.persistenceDisposition == .persistent else { return [] }
         let chatsFolder = try ensureChatsFolder(for: workspace)
 
         let contents = try FileManager.default.contentsOfDirectory(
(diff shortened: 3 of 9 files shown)


baron (changes requested):
### Intent Analysis
**Problem:** Ephemeral workspaces were excluded from the index at creation but not from later persistence paths — autosaves, direct saves, chat history, Agent-session storage, attachments, and published Git artifacts could still create orphaned `workspace.json` and sidecar directories (issue #468).
**Before → After:** Persistence prevention moves from creation-time convention to enforced guards: manager-level save rejection, a typed `WorkspacePersistentStorage` capability gating sidecar writes, and a disk-writer barrier on ephemeral conversion. The change also goes beyond persistence: chat/Agent-session *reads* now throw `ephemeralWorkspace`, and attachments and Git artifact publication are disabled in temporary workspaces.
**Evidence:** Two independent code-first reviews of head `189c5db` converging on `WorkspaceManagerViewModel.swift` (save/enqueue/conversion, ~6785–6793, 6838–6853, 7108–7127, 8207–8219), `AgentSessionDataService.swift:1377-1380`, `ChatHistoryManager.swift:283-285, 442-445`, `MCPGitToolProvider.swift:1117-1127`, `AgentModeViewModel+DraftsAttachments.swift:103-117`; new `WorkspaceEphemeralPersistenceTests`; CI failure signatures matching these paths exactly.
**Confidence:** High on the leak diagnosis, the manager-level boundary design, and the read-path regression (found independently twice, confirmed by CI). Medium on the sidecar capability race (single-source, mechanism-plausible) and on whether feature disablement was intended.
**Red Flags:** All four required app-test shards failing at head; new tests codify the overly broad rejection (`WorkspaceEphemeralPersistenceTests.swift:93-109`) rather than catching it; scope expands across 14 files into chat, Agent Mode, attachments, and MCP Git without an established product decision; no human review discussion.

### Findings
**F1 — High (Confirmed): Reads conflate path resolution with persistence authorization, breaking live MCP semantics.** `AgentSessionDataService.swift:1377-1380` makes `ensureAgentSessionsFolder` throw `ephemeralWorkspace`, but the helper backs reads/lookups (`:1101-1103`, `:1217-1225`, `:1253-1259`): `agent_run steer` on an unknown session returns `ephemeralWorkspace` instead of "not found," and `AgentManageMCPToolService.swift:163-175` aborts before merging live in-memory sessions at `:192+`, so `agent_manage list` fails even for memory-only sessions. Same pattern in `ChatHistoryManager.swift:283-285` guarded at `:442-445`. CI: shard 1 fails at `AgentRunMCPToolServiceSteerResumeTests.swift:177` with exactly `ephemeralWorkspace`. Fix: separate non-mutating resolution from directory creation; ephemeral reads return empty/nil, only mutations reject.

**F2 — High (Confirmed): Established ephemeral Context Builder/Agent functionality disabled without a ratified contract decision.** `MCPGitToolProvider.swift:1117-1127` rejects Git artifact publication; `AgentModeViewModel+DraftsAttachments.swift:103-117` disables attachments. CI: shard 2 fails at `ContextBuilderWorktreeInheritanceTests.swift:160` after "Temporary workspaces cannot publish persistent Git artifacts." Issue #468 requires avoiding orphaned workspace-owned storage, not removing these workflows. Either provide explicitly transient lifecycle-managed storage or make the contract change intentional and maintainer-ratified across callers.

**F3 — Medium (Plausible, single-source): Sidecar authorization is revocable in name only.** `WorkspacePersistentStorage` (`WorkspaceManagerViewModel.swift:236-244`) is a bare URL with no invalidation; `setWorkspaceEphemeral` (`:8208-8218`) awaits the writer barrier *before* setting `isEphemeral`, so a capability issued pre-conversion can write post-conversion. The in-flight test covers only `WorkspaceDiskWriter`, not attachments/artifacts/chats/sessions. Verify and fix, or explicitly accept with a comment — strongly recommended, not the primary blocker.

**F4 — High (Confirmed by metadata): All four required Build-and-Test app shards fail at head `189c5db`.** Shards 1–2 root-cause to F1/F2. Shards 3–4 (`TabContextRoutingTests` ending in `CancellationError` during workspace serialization) are attributed but not fully diagnosed. Head is not merge-ready regardless of finding disposition.

### Maintainer-guidance check
- **User impact/invariant:** Temporary workspaces must create no durable state *while retaining* functional in-memory Agent/MCP/chat behavior; the PR enforces the first half by sacrificing the second.
- **Root-cause confidence:** Confirmed for the original leak and F1/F2 (code path + exact CI signatures); not confirmed for shards 3–4.
- **Authority:** `WorkspaceManagerViewModel` correctly owns durable-storage authorization; read services must not use directory-creating helpers as their authority check — that inversion is the core defect pattern.
- **State-safety:** Workspace-JSON writer barrier is sound and installed pre-transition; the sidecar capability window (F3) remains open under concurrency.
- **Scale/observability:** Disk-leak amplification fixed; but ephemeral operations now emit repeated opaque `ephemeralWorkspace` errors, and enqueue adds a full workspace-JSON decode per save.
- **Recommended scope:** Keep the manager save/autosave boundary, capability type, and writer barrier. Split path resolution from directory creation in `AgentSessionDataService` and `ChatHistoryManager`. Implement transient storage for artifacts/attachments or split that policy into a focused follow-up. Fix affected integration fixtures; address or explicitly defer F3.
- **Validation boundary:** Focused `WorkspaceEphemeralPersistenceTests`, ephemeral Agent MCP steer/list tests, `ContextBuilderWorktreeInheritanceTests`, `TabContextRoutingTests`, all four app shards green, then the live CE MCP smoke flow (Agent Mode/MCP behavior changed).

### Verdict
REQUEST_CHANGES. The centralized persistence boundary is the right design and should be preserved, but the PR demonstrably regresses ephemeral Agent/MCP read semantics (F1), disables established Context Builder functionality without a ratified contract decision (F2), and all four required app-test shards fail at head — two failures directly and mechanically attributable to this patch. These are confirmed behavioral defects with exact file/line and CI evidence, not hypothetical or style concerns. The fix path is well-scoped above.

morluto:
@baron F1 and the associated CI fixture regressions are fixed on visible head `79f1471ff9fb673b2b2db7aaf27d2d40cb771e9d`. Ephemeral Agent-session/chat reads now return empty/nil without creating sidecars, while mutations still fail at the persistence boundary. The MCP steer/list fixtures are restored to temporary workspaces so they exercise the intended live in-memory contract; the persistent-fixture workaround that caused the shard 2/4 failures was removed. Validation: SwiftFormat was run on the touched files; mandatory commit and outgoing-range push preflights passed. Per the maintainer fast-path request, local heavyweight tests were canceled/skipped and fresh hosted CI is authoritative. F2 (transient attachment/Git-artifact behavior) and F3 (revocable sidecar authorization) are not claimed resolved by this commit and remain open for the next scoped follow-up.

morluto:
Correction: the full pushed commit SHA is `79f1471f6a9a12454c9c5a246b49ff7882ab2980` (the short SHA `79f1471f` in the preceding comment was correct). All other validation and scope notes stand.

morluto:
@baron F2 and F3 are now addressed on visible head `480dadbbeed429e573754f2f646e5514658312c4`. F2: attachments and Git/worktree artifacts remain functional in temporary workspaces, but use a workspace-keyed directory under the process temporary root; that transient directory is removed on workspace deletion and persistence-disposition changes, while chat/session persistence remains rejected. F3: persistent sidecar capabilities now carry a lock-protected revocable authorization; conversion marks the workspace ephemeral and invalidates issued capabilities before awaiting the disk-writer barrier, and attachment writes validate the capability at the mutation boundary. Regression coverage exercises transient artifact resolution and a pre-conversion attachment capability rejected after conversion. SwiftFormat was applied to touched files with unrelated formatter churn removed; mandatory commit and push preflights passed. Per fast-path instruction, fresh hosted CI is the build/test authority.

baron (changes requested):
A new exact-head regression exists at `7f93e1a25eea6c1d244ff8eeab1bd79717d26219`: the latest merge resurrects deleted codemap suites without their support types. `CodemapGraphFreezeQueryTests.swift:8` instantiates undefined `CodemapResolutionGate`, matching the hosted compiler failure across all app shards and Sentry build. Reconstruct from current main, retain the ephemeral-storage fixes, do not reintroduce removed suites, resolve conflicts, and rerun exact-head CI.

baron (changes requested):
Blocking on rewritten head `95b2d7e2ba93c36e3fd4bd608d96be3b1d5aecaf`.

The revised ephemeral-storage direction addresses the earlier authority concerns, but the exact head is syntactically unbuildable: the `WorkspaceManagerViewModel` change drops a closing brace, producing `expected '}' in class` and cascading missing-scope errors. Style, every app shard, and the Sentry build fail.

Fix the syntax, retain the single ephemeral/persistent authority, and rerun exact-head ephemeral Agent/MCP plus lifecycle validation.

morluto:
### Audit disposition — **superseded by #814** (2026-08-14)

#814 now covers both sides of this problem: preventing future ephemeral workspace persistence through the domain authority and handling already leaked records through a narrow, reviewed cleanup/tombstone path.

Please port any unique lower-level sidecar/capability tests from this branch that #814 does not already cover, preserve author credit, and close this PR rather than merge two competing persistence models. The original finding was valid; the newer PR is the more complete owner.

morluto:

**Disposition: superseded by #814.** This PR correctly treats ephemeral status as a storage contract and contains useful lower-level sidecar/write-race coverage. However, #814 now owns the broader current architecture: it prevents future domain/catalog persistence, filters persisted ephemeral projections, and provides a narrowly revalidated cleanup path for already leaked records.

Please transfer any unique tests or capability checks from this branch that #814 lacks, credit this work in #814, and close this PR rather than merging two overlapping persistence models. In particular, preserve coverage for queued stale writes and workspace-owned sidecars if it is not already represented in the successor.

morluto:
Superceded

PR #490, opened 2026-07-11
fix(codex): correct managed auth checks and browser login completion

### LINKED ISSUE #479: Codex Test Connection forces token refresh and managed login reports success before OAuth completes
## Summary

RepoPrompt’s Codex connection flow conflates three different states:

1. whether stored Codex credentials exist;
2. whether the current access token works;
3. whether the refresh token can be rotated.

When a valid access token coexists with a stale/reused refresh token, **Test Connection** forces token refresh, receives `refresh_token_reused`, and marks Codex disconnected even though ordinary Codex requests still work.

The subsequent **Login with ChatGPT** flow can then report success before the browser authorization completes because it accepts any pre-existing account returned by `account/read` instead of waiting for the matching `account/login/completed` notification.

## Environment

- RepoPrompt CE debug app at commit `5c1504f1`
- macOS 26.3, arm64
- Codex CLI 0.144.1
- ChatGPT-managed Codex authentication

## Reproduction

Starting state:

- `codex login status` reports `Logged in using ChatGPT`.
- Normal Codex requests work using the existing access token.
- The stored refresh token has already been rotated/consumed.

Steps:

1. Open Settings → CLI Providers → Codex.
2. Press **Test Connection**.
3. Observe that RepoPrompt marks Codex disconnected or requests authentication.
4. Press **Login with ChatGPT**.
5. Leave the browser at the “Choose an account to continue to Codex” screen without selecting an account.
6. Observe that RepoPrompt may nevertheless report that ChatGPT login completed.

## Expected behavior

- Test Connection should distinguish a usable current session from a failed forced refresh.
- A connection-status check should not unnecessarily mutate authentication state.
- Managed browser login should remain pending until the matching OAuth attempt completes.
- RepoPrompt should not silently replace or logout shared Codex CLI authentication.

### Test Connection forces refresh

`CodexManagedAuthRecoveryService.refreshManagedAccount()` sends:

```json
{
  "method": "account/read",
  "params": {
    "refreshToken": true
  }
}
```

The Codex app-server documentation defines this as a forced token refresh.

In the reproduced state, Codex returned:

```text
401 Unauthorized
code: refresh_token_reused
message: Your refresh token has already been used to generate a new access token.
```

The RPC result contained no account, so RepoPrompt set its local Codex connection state to disconnected.

However, immediately afterward:

- `codex login status` still reported ChatGPT login;
- `account/read(refreshToken:false)` returned an existing `chatgpt` account;
- ordinary requests remained usable with the valid access token.

Test Connection therefore presents “refresh credential failed” as “not logged in.”

### Managed login completes prematurely

After `account/login/start`, RepoPrompt subscribes to notifications and records the matching `account/login/completed` result in `LoginNotificationState.successSeen`.

That success state is never consulted.

Instead, RepoPrompt polls:

```json
{
  "method": "account/read",
  "params": {
    "refreshToken": false
  }
}
```

and returns `.authenticated` whenever any account is present. A pre-existing account can therefore satisfy the poll before the new browser OAuth attempt completes.

This was observed with the browser still waiting at the account chooser while RepoPrompt reported successful login.

### Other state inconsistencies

- When `account/login/start` reports “external auth is active,” the recovery service can silently call `account/logout` and retry. This may modify authentication shared with terminal Codex without confirmation.
- RepoPrompt’s visible **Sign Out** action does not call `account/logout`; it only clears RepoPrompt connection state.
- The running RepoPrompt app had eight Codex app-server child processes. Multiple long-lived app-server processes may increase stale authentication-state risk, although the investigation did not establish which process originally consumed the refresh token.

## Impact

- Working Codex credentials appear logged out in RepoPrompt.
- Provider models and controls may be cleared or disabled.
- Users are prompted to authenticate unnecessarily.
- RepoPrompt may report successful login while OAuth is still pending.
- Recovery may unexpectedly affect authentication shared with terminal Codex.
- Once the current access token expires, the stale refresh token cannot obtain a replacement.

## Regression evidence

The embedded changelog says RepoPrompt 2.0.23, released March 19, 2026, introduced “Codex automatic auth recovery.” The current recovery implementation was already present in the May 31 CE root snapshot.

The exact pre-CE introducing commit is not available in this repository’s public history.

## Acceptance criteria

- Test Connection does not force token refresh merely to determine whether an account exists.
- A refresh failure is not presented as a complete logout when the current account/session remains usable.
- Managed browser login waits for a successful `account/login/completed` notification matching the new `loginId`.
- A pre-existing account alone cannot complete a new OAuth attempt.
- Any action that calls `account/logout` clearly warns that shared terminal Codex authentication may be affected and requires explicit confirmation.
- The current local-only Sign Out action is labeled according to its actual effect, or performs a real explicitly scoped logout.
- Regression tests cover:
  - valid existing account plus failed forced refresh;
  - existing account while a new browser login remains pending;
  - mismatched or failed login completion notifications;
  - no implicit logout during recovery.

morluto:
## Summary

Fixes #479.

Corrects RepoPrompt's Codex managed-auth state handling so connection checks no longer invalidate otherwise usable sessions, and browser login cannot report success before OAuth actually completes.

## Changes

- Use passive `account/read(refreshToken: false)` for Test Connection and cached connection-status checks.
- Preserve forced token refresh for genuine authentication recovery.
- Keep passive checks and forced refreshes in separate single-flight tasks.
- Require browser login to receive a successful `account/login/completed` notification matching the returned `loginId`.
- Verify the account passively after the matching completion notification.
- Ignore missing, stale, or mismatched login-completion notifications.
- Remove the implicit `account/logout` fallback when external authentication is active.
- Add focused regression coverage for the managed-auth state machine.
- Stabilize two existing prompt preassembly synchronization tests discovered during PR-ready validation.
- Reconcile the affected test contract ledger entries.

## User impact

Previously, a user could see this sequence:

1. Codex CLI remained usable with its current access token.
2. RepoPrompt's Test Connection forced refresh.
3. A stale or reused refresh token failed with `401 refresh_token_reused`.
4. RepoPrompt incorrectly appeared disconnected.
5. Login with ChatGPT opened a browser flow.
6. RepoPrompt saw the pre-existing account and reported success before the new OAuth flow completed.

After this change:

- status checks do not force token rotation;
- refresh failures are not confused with passive account presence;
- browser login only succeeds after the correlated OAuth completion;
- RepoPrompt does not silently logout shared terminal Codex authentication.

## Validation

- `CodexManagedAuthRecoveryServiceTests`: 7 passed
- `PromptContextPreAssemblyServiceTests`: 16 passed
- Full coordinated root test suite: passed
- Test contract ledger verification: 3,155 tests reconciled
- Repository guardrails and secret scans: passed

### DIFF
--- a/Sources/RepoPrompt/Features/Settings/ViewModels/APISettingsViewModel.swift
+++ b/Sources/RepoPrompt/Features/Settings/ViewModels/APISettingsViewModel.swift
@@ -1283,7 +1283,7 @@ public class APISettingsViewModel: ObservableObject {
     private func probeCachedCodexConnection(ifNeeded: Bool) async -> Bool {
         guard ifNeeded else { return false }
         let publicationToken = codexSessionFence.capturePublicationToken()
-        let refreshResult = await codexManagedAuthRecovery.refreshManagedAccount()
+        let refreshResult = await codexManagedAuthRecovery.checkManagedAccount()
         guard codexSessionFence.allowsAuthenticationPublication(publicationToken) else { return false }
         switch refreshResult {
         case let .recovered(account):
--- a/Sources/RepoPrompt/Infrastructure/AI/Providers/Codex/Shared/CodexManagedAuthRecoveryService.swift
+++ b/Sources/RepoPrompt/Infrastructure/AI/Providers/Codex/Shared/CodexManagedAuthRecoveryService.swift
@@ -14,6 +14,7 @@ protocol CodexManagedAuthRPCClient: Sendable {
 
 protocol CodexManagedAuthRecovering: Sendable {
     func refreshManagedAccount() async -> CodexManagedAuthRefreshResult
+    func checkManagedAccount() async -> CodexManagedAuthRefreshResult
     func managedAccountSnapshot() async -> CodexManagedAccount?
     func startManagedChatgptLogin(
         openURL: @MainActor @escaping @Sendable (URL) -> Void
@@ -249,6 +250,7 @@ actor CodexManagedAuthRecoveryService: CodexManagedAuthRecovering {
     private let sleep: @Sendable (TimeInterval) async throws -> Void
     private let retirementSleep: @Sendable (TimeInterval) async throws -> Void
     private var inFlightRefresh: InFlightRefresh?
+    private var inFlightCheck: InFlightRefresh?
     private var inFlightLogin: InFlightLogin?
     private var inFlightLogout: InFlightLogout?
     private var latestManagedAccount: CodexManagedAccount?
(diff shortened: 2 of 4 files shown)


baron (changes requested):
Exact head `5530d1de92cd2357a5c47dda15ffabe9b20acd8c` conflicts with the current managed-auth authority. The PR's plain recovered/authenticated results and three-task model predate current account projection, logout, device-code, mutation-generation fencing, and richer cancellation state. Please deliberately port the passive `account/read(refreshToken:false)` and correlated login-completion behavior while preserving those current semantics, drop obsolete ledger edits, update call sites, and rerun focused auth plus exact-head CI.

baron (changes requested):
Blocking on rewritten head `2af1613d9a0726061e4b0bc2797281347c876a58`.

The passive account-read direction has merit, but the exact head does not compile: `CodexManagedAuthRecoveryServiceTests.swift:674` places `await` inside `XCTAssertEqual`'s non-async autoclosure. Style also fails. Please evaluate the async value before the assertion, cover interaction between the new `inFlightCheck` and login/refresh ownership, then rerun the full exact-head lanes.

morluto:
### Audit disposition — **authentication concurrency blocker** (2026-08-14)

Passive account checks and forced refreshes use separate in-flight tasks. Because the actor is reentrant across awaits, they can overlap and publish in completion order rather than authority order; a slower passive check can overwrite the result of a successful forced refresh.

Please place both operations behind one publication generation/authentication lane and add deterministic tests for both completion orders. Also distinguish a transient passive transport failure from an authenticated-state transition: “unable to determine” should not clear a known-good account or automatically become “user must log in.” The OAuth completion correlation work is good, but I would not merge with this remaining state race.

morluto:

**Disposition: block pending unified auth publication fencing.** Separating passive checks from forced refresh is correct at the RPC level, but the two operations have independent in-flight tasks. Because the actor is reentrant across `await`, a passive check and forced refresh can overlap; whichever completes last calls `applyRefreshResult` and can overwrite a newer successful refresh with stale/weaker state.

Please put check/refresh/login/logout behind one authentication operation generation or one publication lane, and test both completion orders for simultaneous passive check + forced refresh. Also distinguish transient passive transport failure from confirmed sign-out: an inability to determine account state should not clear a known-good snapshot or automatically become “user must log in.”

PR #498, opened 2026-07-11
Split conductor scheduling into root and provider lanes

morluto:
## Problem

Conductor currently treats root SwiftPM work and provider-package work as the same build resource. An unrelated provider test therefore waits behind a root build or test even though the packages can run independently. Raising the global worker count alone would not help because the shared lane still creates head-of-line blocking.

## Approach

Model the resources that jobs actually contend for. Root and provider SwiftPM operations receive separate lanes, while formatting and lifecycle operations continue to claim every resource they can affect.

## Changes

- Replace the shared build lane with `rootBuild` and `providerBuild`.
- Preserve FIFO ordering between jobs that claim the same resource.
- Allow non-conflicting queued jobs to proceed independently.
- Add queue-age and resource-wait telemetry.
- Freeze timing when queued jobs are canceled.
- Update conductor protocol and documentation.

The global heavy-job capacity remains `1` by default. Additional concurrency still requires explicit `REPOPROMPT_DEV_HEAVY_SLOTS=N`, allowing this scheduler change to land without changing normal machine load.

## Validation

- Focused conductor lifecycle and scheduling tests
- `python3 Scripts/test_conductor_lifecycle.py`
- `.agents/skills/rpce-contribution-check/scripts/preflight.sh pr-ready`


baron:
Recommend closing rather than rebasing mechanically. The conductor has since gained major control-plane, machine-wide heavy-admission, and durable-cache rewrites, while current authority still intentionally uses the shared build lane. If concurrency remains a measured bottleneck, open a fresh focused issue/PR from current measurements.

PR #499, opened 2026-07-11
Reduce Git subprocesses in the codemap layout watcher test

morluto:
## Problem

The codemap layout watcher regression test starts six Git subprocesses while preparing its fixture. Much of that work establishes repository eligibility, but the behavior under test is reclassification after a `.git` layout change. Repeating unrelated production probes makes an already integration-heavy suite slower.

## Approach

Keep the parts that establish the regression contract real: a valid repository, the `.git` transition, watcher delivery, binding-engine readiness, and the preflight assertion. Control only the unrelated eligibility decision through the existing test seam.

## Changes

- Reduce fixture setup from six Git subprocesses to three.
- Use command-scoped Git identity and signing configuration.
- Use the controlled eligibility seam.
- Preserve the real filesystem and watcher path.
- Normalize ledger resource tags without changing scenario count or historical timing.

This changes test setup only; production behavior is unaffected.

## Validation

Repository guardrails and contribution preflights passed. Focused XCTest execution and ledger verification remain macOS CI checks.


baron:
Recommend closing as fully absorbed. GitHub reports 0 changed files, 0 additions, and 0 deletions at exact head `29b737606530f6d8afb034b9ed28dc230d3c5313`. There is no remaining reviewable contribution.

PR #500, opened 2026-07-11
Speed up preflight with impacted lint and test selection

morluto:
## Problem

Local `pr-ready` validation performs repository-scale style and test work even for narrow changes. That increases feedback time and conductor contention, while simply skipping full validation would risk missing changes to shared configuration or test infrastructure.

## Approach

Select validation from staged, unstaged, and untracked files, but fall back to full-root validation whenever the affected boundary is broad or uncertain. Hosted CI remains comprehensive and unchanged.

## Changes

- Add impacted SwiftFormat and SwiftLint targets using one invocation per tool.
- Add impacted test selection with a smoke-test floor.
- Fall back to full validation for tooling, configuration, and other broad boundaries.
- Report timing for successful and failed preflight phases.
- Pin new SwiftFormat defaults so tool upgrades do not rewrite the existing baseline.
- Add focused selection and fallback tests.

## Validation

- `python3 Scripts/test_swift_style.py`
- `python3 Scripts/test_test_suite_optimizer.py`
- `python3 Scripts/test_contribution_preflight.py`
- `make guardrails`


baron (changes requested):
Exact head `0d473964aa43342f47e798fa4c75de61ad4eb075` breaks every SwiftFormat-backed style command. `swift_style.sh` still calls `install_format_tools.sh resolve-swiftformat`, while this diff removes that subcommand and rejects it with exit 2. Restore a supported resolver or use a validated executable directly, add an integration test for the real script contract, resolve current-main conflicts, and rerun the full style lane.

PR #501, opened 2026-07-11
Isolate shared MCP state in the Agent Mode handoff test

morluto:
## Problem

The Agent Mode handoff regression test mutates shared MCP, window, and global application state. Without ownership covering the complete setup, assertion, and cleanup lifecycle, another concurrently scheduled suite could observe transient state or begin before cleanup finishes.

## Approach

Hold the existing shared-server lease for the test's entire lifecycle and finish deterministic window cleanup before releasing it. Also classify the suite with the canonical serial-group tag so the runner and the test agree about the shared resource.

## Changes

- Hold `MCPSharedServerTestLease` across setup, assertions, and teardown.
- Clean up windows before lease release on success and failure.
- Keep test-only helpers behind `#if DEBUG`.
- Use the canonical `mcp_shared_server` serial-group tag.
- Normalize ledger tags without changing scenario count.

This does not make the suite parallel-eligible. It makes its required serialization explicit and its cleanup deterministic.

## Validation

Repository guardrails, suite-plan classification, and contribution preflights passed. Focused XCTest execution and ledger verification remain macOS CI checks.


baron:
Recommend closing this PR in its current form. It conflicts with main and all four app shards fail because the branch reintroduces an obsolete test-contract ledger removed in #704; most changed lines are in that deleted ledger. If the focused lease/cleanup race still reproduces, port it as a small current-main test-only PR without ledger changes.

PR #507, opened 2026-07-11
Fix Agent Mode context-window reporting

### LINKED ISSUE #506: Agent Mode reports incorrect context windows for Claude and Codex sessions
## Summary

Agent Mode can report a context-window denominator that does not match the provider’s effective runtime limit.

This makes the context percentage misleading at exactly the point where users rely on it to decide whether to add more context, compact, or start a new session.

Examples observed in the CE debug app:

- A Claude model with a 1M model maximum and a configured 400K auto-compact window initially reports 1M.
- After the first response, the same Claude session can incorrectly switch to 200K.
- Fresh Codex GPT-5.5 and GPT-5.6 sessions can briefly show the internal 200K fallback before runtime usage reports the model’s actual context window. GPT-5.5 was observed at approximately 258.4K after the first response. In case of GPT 5.6, it was 353.4K.

### Claude

1. Configure `CLAUDE_CODE_AUTO_COMPACT_WINDOW=400000`.
2. Start a new Agent Mode session with a 1M-capable Claude model.
3. Inspect the context tooltip before sending a message, while waiting for the first response, and after the response arrives.

Expected: the effective window remains 400K, while the 1M model maximum is available as separate detail.

Actual: the UI can show 1M initially and later switch to 200K.

### Codex

1. Start a new Agent Mode session with GPT-5.5 or GPT-5.6.
2. Send the first message and inspect the context indicator before the response arrives.
3. Inspect it again after runtime usage is received.

Expected: an unknown denominator is shown honestly until Codex reports its runtime window.

Actual: the internal 200K fallback can appear as though it were the real window, then change when the runtime-reported value arrives. For GPT-5.5, this was observed at approximately 258.4K.

## Additional bug found during investigation

Claude result events can contain `modelUsage` entries for both the main model and background models such as Haiku. The translator previously selected the first positive entry from an unordered dictionary, allowing a background model’s 200K window to be attributed to a 1M main-model session.

The investigation also found that the 200K display fallback was being used as a validation ceiling while the real window was unknown. This could silently discard legitimate context readings above 200K in fresh 1M sessions and after restoring persisted usage.

## Acceptance criteria

- Configured provider limits are used as the effective denominator when known.
- Canonical model capacity remains available separately from the effective limit.
- Multi-model Claude usage events are attributed to the active main model.
- Valid usage readings above 200K are retained while the canonical window is still unknown.
- Unknown windows are shown without presenting an internal fallback as fact.
- Model switches, process restarts, and session/workspace changes do not retain stale window values.
- Focused regression coverage exists for Claude, Codex, persistence, and sidebar presentation.

BetaFold3:
Fixes #506

## Why

The context indicator could use three different denominators during one session:

- a static model maximum before provider activity,
- the internal 200K fallback while waiting for usage,
- and the provider-reported window after the first response.

Besides being confusing, this could produce incorrect percentages and silently discard valid usage readings above 200K.

This was especially visible with Codex GPT-5.5 and GPT-5.6: a fresh session could present the 200K internal fallback before runtime usage reported the actual window. GPT-5.5 was observed reporting approximately 258.4K after the first response. 353.4K in case of GPT 5.6.

## What changed

- Resolve Claude’s configured auto-compact window from the effective launch environment and supported Claude settings files.
- Resolve Codex context configuration from user and project `config.toml`.
- Keep canonical model capacity separate from the effective reporting denominator.
- Show an unknown denominator as `—` instead of exposing the internal 200K fallback as fact.
- Carry configured Claude limits to the sidebar before the first usage event.
- Preserve spawn-authoritative values across asynchronous provisional resolution, model changes, and workspace changes.
- Retain valid large Claude usage readings while the canonical window is still unknown.

## Bug found along the way

Claude result events may report usage for several models in the same session, including background Haiku work.

The translator previously selected the first positive `modelUsage` entry from an unordered dictionary. A background 200K model could therefore replace the main model’s 1M window, causing the sidebar to report 200K after the first response.

This change attributes usage to the tracked main model, handles normalized model variants deterministically, and resets attribution across live model switches and process respawns.

## User-visible behavior

- A 1M-capable Claude model configured to compact at 400K reports 400K consistently, with the 1M model maximum shown separately.
- Codex GPT-5.5 and GPT-5.6 sessions show an unknown denominator before runtime usage, then switch to the provider-reported value when it arrives.
- Percentages and warning colors are suppressed while the denominator is unknown.
- Valid context readings above 200K are no longer dropped before the canonical window is known.

## Validation

- Focused Claude/Codex resolver and context-usage suites
- `make dev-provider-test FILTER=ClaudeSDKNDJSONTranslatorTests`
- `make dev-test FILTER=AgentRuntimeSidebarViewModelTests`
- `make dev-test FILTER=ClaudeContextUsageEstimatorTests`
- `make dev-lint`
- Authoritative root/provider test lists and test-ledger reconciliation
- Manual CE debug-app checks for Claude 400K and Codex unknown-to-runtime-window transitions


BetaFold3:
This PR does not modify the failing test or Codemap implementation. Mind helping with the re-trigger? @baron

BetaFold3:
friendly ping @baron @provencher

baron:
@BetaFold3 I'll look at this and see what we can do.

baron:
Current-main refresh on exact head `3c4c9eb9de78df21b2c3db66195f7f73322709e6`: the context-window reporting issue remains open, but this historical branch has broad Agent/provider conflicts and restores the test-suite ledger authority deleted on main. Please replace it with a focused current-main implementation that omits obsolete ledger changes and rerun exact-head CI.

morluto:
### Audit disposition — **attribution blocker; split the change** (2026-08-14)

The denominator cleanup is valuable, but the main-model attribution still has unsafe fallbacks. When no tracked-model candidate matches, selecting the global maximum can attribute a background model’s window to the active model. Symmetric prefix matching can also associate a neighboring/versioned model incorrectly.

Please fail closed unless there is an exact match, an explicit reviewed alias, or a sticky previously authoritative value. Add mixed `modelUsage` cases with background-only entries and similarly prefixed models. I also recommend separating translator attribution, configuration-file resolution, and UI denominator changes so each authority contract can be reviewed independently.

morluto:

**Disposition: block and split the attribution/config/UI changes.** Removing the fake 200K denominator and separating canonical capacity from effective reporting are good. The main-model attribution still has a fail-open path: when exact/normalized attribution cannot identify the tracked model, selecting the global maximum can assign a background model’s window to the foreground session. Symmetric prefix matching can also conflate distinct versioned model IDs.

The safe fallback is “unknown” unless there is an explicit alias or a previously authoritative sticky main-model value. Please add background-only/no-match/version-collision tests and split translator attribution, Claude/Codex config resolution, and UI presentation into independently reviewable commits or PRs. The current surface is too broad to accept with an attribution ambiguity at its core.

PR #516, opened 2026-07-11
Add title-adjacent Agent chat options menu

baron:
## Summary
- place a native Chat Options ellipsis beside the centered Agent chat title
- build the NSMenu lazily from a menu-open snapshot with pin, rename, stash, and delete actions
- preserve the existing New Session titlebar accessory and avoid observing streaming session state
- hide the menu when no focused Agent session is available and collapse only redundant single-window tab chrome

## Validation
- strict SwiftFormat/SwiftLint: passed
- focused `RepoPrompt` product build: passed
- contribution whitespace, guardrail, staged/outgoing secret scans: passed
- live title-adjacent placement and menu contents were verified during implementation

Hosted exact-head checks will provide the final full test/build gate.

### DIFF
--- a/Sources/RepoPrompt/App/WindowContentView.swift
+++ b/Sources/RepoPrompt/App/WindowContentView.swift
@@ -28,6 +28,7 @@ struct WindowContentView: View {
             // default app-name title over the workspace name whenever it refreshes the
             // window chrome (visible in both window titles and native tab names).
             .navigationTitle(windowState.displayedWindowTitle)
+            .removingSystemToolbarTitle()
             .background(
                 WindowAccessor { newWindow in
                     // IMPORTANT: do not mutate SwiftUI @State here.
@@ -72,5 +73,16 @@ struct WindowContentView: View {
             .sheet(isPresented: $versionManager.shouldShowVersionPopup) {
                 VersionPopupView(isPresented: $versionManager.shouldShowVersionPopup)
             }
+    }
+}
+
+private extension View {
+    @ViewBuilder
+    func removingSystemToolbarTitle() -> some View {
+        if #available(macOS 15.0, *) {
+            toolbar(removing: .title)
+        } else {
+            self
+        }
     }
 }
--- a/Sources/RepoPrompt/App/WindowState.swift
+++ b/Sources/RepoPrompt/App/WindowState.swift
@@ -939,7 +939,8 @@ class WindowState: ObservableObject {
 
     private func collapseRedundantSingleWindowTabBarIfNeeded(for window: NSWindow) {
         guard let tabbedWindows = window.tabbedWindows,
-              tabbedWindows.count == 1
+              tabbedWindows.count == 1,
+              window.tabGroup?.isTabBarVisible == true
         else {
             return
         }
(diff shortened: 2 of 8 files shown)


Sources/RepoPrompt/App/WindowContentView.swift
```diff
@@ -74,3 +75,14 @@ struct WindowContentView: View {
             }
     }
 }
+
+private extension View {
+    @ViewBuilder
+    func removingSystemToolbarTitle() -> some View {
+        if #available(macOS 15.0, *) {
```
provencher, line 82: Do we need this backcompat stuff?

Sources/RepoPrompt/App/WindowState.swift
```diff
@@ -513,9 +528,15 @@ class WindowState: ObservableObject {
     private func configureWindowChrome(for window: NSWindow) {
         // Keep titlebar visually continuous with content (no horizontal separator).
         window.toolbar?.showsBaselineSeparator = false
+        if #unavailable(macOS 15.0) {
```
provencher, line 531: again I dont like back compat stuff. I dont think we need support for ancient versions

Sources/RepoPrompt/App/WindowState.swift
```diff
@@ -773,6 +930,20 @@ class WindowState: ObservableObject {
...
+        }
+    }
+
+    private func collapseRedundantSingleWindowTabBarIfNeeded(for window: NSWindow) {
+        guard let tabbedWindows = window.tabbedWindows,
+              tabbedWindows.count == 1
+        else {
+            return
+        }
+        window.toggleTabBar(nil)
```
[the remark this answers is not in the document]
baron: Fixed in 8aafe862. `collapseRedundantSingleWindowTabBarIfNeeded` now requires `window.tabGroup?.isTabBarVisible == true` before calling the non-idempotent toggle, so the synchronous and deferred invocations can only hide an actually visible single-window tab bar. Formatter, strict lint, and focused RepoPrompt build all pass.

Sources/RepoPrompt/App/WindowContentView.swift
baron: Verified this availability path is still required by the repository’s current supported deployment target. `Package.swift` declares `.macOS(.v14)` and `AppBundle/Info.plist.template` declares `LSMinimumSystemVersion` 14.0. A direct Swift compiler probe targeting macOS 14 rejects `.toolbar(removing: .title)` because `.title` is macOS 15+. Removing this branch would break supported builds, so it is intentionally retained.

Sources/RepoPrompt/App/WindowState.swift
baron: Verified against the current support contract: both `Package.swift` and the packaged Info.plist set macOS 14. The pre-15 `titleVisibility = .hidden` path supplies the corresponding title suppression where SwiftUI’s macOS-15-only `.toolbar(removing: .title)` API cannot be used. Removing it would regress title behavior on a currently supported OS, so it is intentionally retained.

Issue #139, opened 2026-06-10
Release blocker: signed packaged MCP helper aborts during release smoke


baron:
## Summary

The `Publish Release` workflow is failing in the post-signing secret-free smoke job because the signed packaged `repoprompt-mcp` helper aborts during the first real MCP request (`-e 'windows'`).

This blocks publishing releases: the release workflow completes validation, build, signing, notarization, and draft asset creation, but the final signed-helper smoke fails.

## Failing run

- Workflow run: https://github.com/repoprompt/repoprompt-ce/actions/runs/27253740816
- Workflow: `Publish Release`
- Event: `workflow_dispatch`
- Head branch: `main`
- Head SHA: `be74f5073311f487d69e19d3a4aae2612c50603a`
- Failing job: `Smoke Signed Helper Without Release Secrets`
- Failing job URL: https://github.com/repoprompt/repoprompt-ce/actions/runs/27253740816/job/80490475790
- Artifact being smoked: `RepoPrompt-1.0.12-13.zip`

## What passed before the failure

The release did **not** fail during build, signing, notarization, artifact upload, or static validation.

Passed jobs/steps include:

- `Validate Approved Release Ref`
- `Build Approved Source Without Secrets`
- `Sign, Notarize, and Create Draft`
- signed ZIP download
- ZIP checksum verification
- embedded MCP helper layout validation
- universal architecture validation: `arm64,x86_64`
- artifact manifest verification
- packaged helper `--version` smoke

The helper prints:

```text
rpce-cli (repoprompt-mcp) 1.0.12
OK: Signed draft packaged app early helper passed --version smoke.
```

## Failure excerpt

The failure occurs when `Scripts/smoke_packaged_mcp_roundtrip.sh` launches the signed app and then runs the packaged helper against it:

```bash
./trusted-control-plane/Scripts/smoke_packaged_mcp_roundtrip.sh \
  "extracted/RepoPrompt CE.app" \
  "Signed draft packaged app" \
  "$ARTIFACT_MANIFEST_PATH"
```

Relevant log excerpt:

```text
freed pointer was not the last allocationERROR: Signed draft packaged app helper request failed with exit 250: /Users/runner/work/repoprompt-ce/repoprompt-ce/extracted/RepoPrompt CE.app/Contents/MacOS/repoprompt-mcp
--- packaged app log tail ---
BootstrapSocketServer: failed to read handshake from fd 5
BootstrapSocketServer: failed to read handshake from fd 5
[...]

PR #517, opened 2026-07-11
fix(mcp): avoid task-group teardown abort in runTransport retry backoff

### LINKED ISSUE #513: repoprompt-mcp aborts in runTransport retry backoff during teardown (residual swift_task_dealloc crash after #140)
## Summary

`repoprompt-mcp` still aborts with the `swift_task_dealloc` SIGABRT signature that #139 reported and #140 fixed — but at a different site: the retry-backoff `Task.sleep` inside `MCPService.runTransport()`. Crash reports continue on binaries that already contain the #140 fix, so this appears to be a residual second site of the same crash class. Clients experience it as intermittent `MCP error -32000: Connection closed` and tool-call timeouts.

## Evidence

- 28 crash reports in `~/Library/Logs/DiagnosticReports/` from 2026-07-08 10:20 through 2026-07-11 08:52 (+08:00), all one signature: `EXC_CRASH (SIGABRT)`, crashing thread `swift::swift_Concurrency_fatalError` → `swift_task_dealloc`, with `MCPService.runTransport()` and a closure in `MCPService.run()` on the stack.
- Binary: `repoprompt-mcp` from `/Applications/RepoPrompt CE.app/Contents/MacOS/` (Homebrew), across two build UUIDs — 23 crashes on the 1.0.26-era binary (07-08 → 07-09) and 5 on the 1.0.27 binary (07-10 → 07-11). 1.0.27 postdates #140, and `Sources/RepoPromptMCP/main.swift` has only received `CLI_VERSION` bumps since (checked through v1.0.28, `6ceafd39`), so the crash is still live on current main.
- Symbol mapping (nm/otool against the 1.0.27 binary): the reported `MCPService.runTransport()+44` return PC sits immediately after `_swift_task_dealloc`, and the matching allocation/resume continuation is around `ContinuousClock.sleep` — i.e. the abort happens while the compiler-generated frame for the retry-backoff sleep is deallocated on resume.

Line references at `bf9a0b82` (unchanged on main apart from version bumps):

- The backoff sleep: `Sources/RepoPromptMCP/main.swift` — `runTransport()`'s reconnection loop, `try await Task.sleep(for: .seconds(delay))` (~L2316 in the retry loop at ~L2263-2320).
- The cancellation shape: `MCPService.run()` runs `runTransport()` as the third child of a throwing task group and cancels the group when any child completes or throws (~L2113-2153). Cancellation interrupting the backoff sleep is the plausible trigger.

No spawn/exec/pipe frames appear in any report; source-level structured concurrency looks intact (the group scope waits for its children), so the underlying mechanism may be a Swift runtime/compiler async-frame teardown issue exposed by this cancellation/error control-flow shape — same flavor as the race #140 addressed at initialization.

## Impact

- Intermittent `MCP error -32000: Connection closed` mid-session; tool calls dying or timing out; MCP clients (Claude Code and others) losing the server without a clean shutdown.
- Observed to break real sessions: it wedged an agent-delegated investigation mid-run and killed unrelated tool calls the same day.

## Prior art

- #139 / #140: same abort signature at initialization (throwing-task-group timeout race), fixed — this report is the residual site in the reconnection backoff.
- #157 / #163: helper SIGABRT during transport failure, but via stderr `FileHandle.writeData` — different path.
- #301: same signature reported in the app process for `bind_context` — possibly the same class on the app side.
- #170: open umbrella for transport-closure remediation; this crash is a concrete contributor.

## Suggested direction

Harden the backoff sleep against group-cancellation teardown the same way #140 hardened the initialization timeout (e.g. avoid suspending on the compiler-generated sleep frame inside the cancellable group child, or restructure the retry wait so cancellation resumes it deterministically before the frame is torn down).

pnascimento9596:
## What changed

Replaced the three-child throwing task-group race in `MCPService.run()` with an explicit-task race that:

- settles the first returned value or thrown error under a lock;
- cancels all three operations after settlement;
- awaits every task to completion before returning or throwing;
- propagates caller cancellation through the same deterministic teardown path.

`MCPServiceProxyTaskGroupPolicy`, terminal-record persistence, the post-race cancellation check, kill-signal semantics, and `runTransport()` retry/backoff behavior are unchanged.

Added deterministic tests for every child winning with either a value or error, loser cancellation and completion before return, and outer cancellation draining all three children. The SIGABRT is a Swift 6.2 release-runtime interaction and is not unit-reproduced; the tests pin the replacement control-flow structure identified by the issue's crash forensics.

The contract ledger also reconciles three executable IDs that were already missing on current `main`, which was required for exact ledger verification after adding this suite.

## Why

Issue #513 maps the residual `swift_task_dealloc` abort to the retry-backoff sleep in `MCPService.runTransport()` while the surrounding task group is being cancelled. This mirrors the crash class reported in #139 and fixed at the initialization site in #140 by commit `9f881efb`.

Flagged decision for maintainer review: this restructures the second race site away from a value-returning throwing task group, following #140's explicit-task settlement and drain pattern. Please scrutinize the preserved first-completion priority, especially value versus error ordering.

## Validation

Passed:

- `make dev-test FILTER=MCPServiceProxyRaceTests` (3 tests)
- `make dev-test FILTER=MCPProxyTerminalRecordTests` (7 tests)
- failing-first structural run against a temporary cancel-without-await variant
- `make dev-test-list`
- `python3 Scripts/test_suite_optimizer.py verify-ledger --ledger Scripts/Fixtures/test-suite-contract-ledger.tsv` (3175 IDs)
- `make dev-format-check`
- `make dev-lint`
- `make guardrails`
- contribution preflights: `commit` and `push`

Local full-root status: `make dev-test-impacted` correctly required the full root suite because the ledger is a broad boundary. Three coordinated full-root attempts did not complete green due unrelated, nondeterministic codemap/worktree failures. The repeated failure was `PersistentAgentModeMCPReadFileConnectionTests/testPairAgentOwnedNoRangeReadSelectsNonEmptyWorktreeLogicalFile`; it passed immediately when rerun alone. One attempt also failed two other unrelated codemap tests. The issue #513 focused suites stayed green.

Not run:

- release-configuration build or packaged release smoke. The coordinated product build was debug SwiftPM; signed app packaging stopped before compilation because no signing identity or explicit ad hoc signing opt-in was configured.

Fixes #513


provencher:
Thanks for contributing! This looks like a great fix

PR #532, opened 2026-07-12
Fail Codex child sessions closed when RepoPrompt MCP provisioning or routing never confirms

### LINKED ISSUE #514: Codex child agents can start without RepoPromptCE MCP tools when routing/bootstrap failures are swallowed
## Summary

A Codex-backed child agent session spawned via `agent_run` can start "successfully" **without any RepoPromptCE MCP tools** (`mcp__RepoPromptCE__*`), because MCP routing/bootstrap failures during child start are swallowed at three points. The child runs with shell tools only and no surfaced error — from the outside it looks like agent flakiness, and any instruction that references RepoPrompt MCP tools (e.g. `set_status`) makes the child fail or refuse.

## Observed behavior

On 2026-07-12 (debug build at `bf9a0b82`, Codex CLI 0.144.1), one `explore` child dispatched via `agent_run` refused its task because `mcp__RepoPromptCE__set_status` was unavailable in its session, while two sibling explores dispatched the same way at the same time ran fine using shell tools. Nothing in the parent's `agent_run` result or the session state indicated the child had started without its MCP tools.

## Code path (line refs at `bf9a0b82`; none of these files changed on main through v1.0.28)

Children are intended to get RepoPromptCE tools: the server name is the fixed literal `RepoPromptCE` (`Sources/RepoPrompt/Infrastructure/MCP/RepoPromptMCPServerConfiguration.swift:16` — not build-variant-derived), start/resume registers MCP ownership and supplies runtime overrides, and the explore advertisement policy explicitly keeps `set_status` visible.

But each of the three failure signals on that path is discarded:

1. `CodexAgentModeCoordinator.swift:3871-3917` — the coordinator can establish a live Codex thread and then ignore a `false` result from the MCP routing wait.
2. `MCPBootstrapLease.swift:238-275` — `releaseWhenRouted()` returns `false` and clears the pending policy after timeout/failure; callers don't act on it.
3. `CodexNativeSessionController.swift:1074-1088` — `startOrResume()` discards the success Boolean from `ensureCodexServerForDiscovery()`.

Any transient route/bootstrap failure therefore yields a functioning Codex child with native shell tools but no RepoPrompt tools, silently.

## Impact

- Children that need RepoPrompt MCP tools fail or refuse in ways that read as model/agent inconsistency rather than a provisioning defect.
- Parents/orchestrators can't detect the degraded state: the start reports success and nothing marks the session as tool-less.
- Sibling behavior is misleading: children that happen not to touch MCP tools proceed fine, masking whether their tools were present at all.

## Suggested direction

Surface the failure instead of discarding it: fail the child start, or mark the session as MCP-degraded and emit a visible warning in the session state/log when routing/bootstrap doesn't complete. Related but distinct work: #443 bounds provisioning stalls (doesn't detect a child that started without tools), #383/#388 covered worktree-inheritance startup failures, #158 is the Cursor-specific variant of tools never arriving.

mplibunao:
Fixes #514.

## Problem

Agent Mode child sessions (predominantly Codex) intermittently start with shell tools but no RepoPromptCE MCP tools, while the parent's `agent_run` sees the startup as successful. The issue body identifies three silent-swallow layers on the Codex bootstrap path; this PR closes all three so a child that cannot be confirmed to hold RepoPrompt tools fails closed **before its first user turn**, with a typed error surfaced to the parent.

Field evidence while diagnosing: the failure is app-state-dependent rather than uniformly random — during one observed degradation window it went from ~1-in-6 to 3 consecutive tool-less spawns with **zero** bridge crash reports, consistent with app-side swallowed readiness failures rather than the bridge-crash class (#513/#517).

## What changed (one commit per layer)

1. **`f80e2005` — throwing routing readiness on `MCPBootstrapLease`.** New `MCPBootstrapReadinessError` (with `LocalizedError` descriptions) and `requireRouting(timeoutMs:)`, a fail-closed variant of `releaseWhenRouted` that performs the same idempotent cleanup and throws instead of returning a swallowed `false`. Cleanup (policy clear, routing teardown) became memoized joinable operations, so concurrent lifecycle calls join a single in-flight cleanup instead of duplicating or skipping it. The Boolean API is unchanged and all its other callers stay untouched.

2. **`d39eebc6` — provisioning validated before process start.** `CodexNativeSessionController.startOrResume` gains an injectable throwing provisioner (production default validates `ensureCodexServerForDiscovery()`; `success == false` only ever means a real directory-create/config-write failure, so no healthy no-op can misread as failure). For tool-expecting children (`expectedMCPClientName != nil`) it runs after binding and **before** PID registration, app-server launch, or any thread request; a failure or a cancellation during provisioning aborts through the existing catch path with no process started.

3. **`0df1ef7f` — terminal routing gate before the first turn.** The routing wait in `CodexAgentModeCoordinator.ensureCodexNativeSession` (previously `_ = await releaseWhenRouted(...)`) is now terminal on timeout: the started controller is torn down, the readiness error is published through the existing terminal-failure machinery (child transcript + parent-facing failed run state, exactly once), and `startUserTurn` is never dispatched. A cancellation during the wait stays an ordinary run cancellation and also cannot cross the first-turn boundary. A resumed session whose routing is already confirmed does not spuriously fail closed.

**Scope guardrails:** additive only — `releaseWhenRouted` remains for its other callers (Headless, Claude, ACP, ContextBuilder); deferred-routing ACP paths are untouched (`ACPSynchronousMCPStartupTests` green); Claude children keep their existing listener fail-close. A compiler-forced migration of every runner to the throwing API is deliberately left as a possible follow-up. No auto-retry: a redispatch by the parent is safe because every failure here is pre-first-turn.

## Validation

New/updated suites (deterministic, no sleeps; gate/continuation-probe synchronization):

| Suite | Tests | Result | Time |
|---|---|---|---|
| `CodexMCPRoutingReadinessTests` (new; serialized on the shared MCP test lease) | 3 | 0 failures | 0.58s |
| `CodexMCPBootstrapReadinessTests` (new) | 3 | 0 failures | 0.57s |
| `MCPBootstrapLeaseTests` (+5 contract tests incl. joinable-cleanup and cancellation races) | 11 | 0 failures | 0.49s |
| `ACPSynchronousMCPStartupTests` (regression) | 3 | 0 failures | 2.38s |

- Full root suite: 3186 cases; the only 2 failures (`MCPCodeStructureWorktreeTests`, `ProcessLauncherDescriptorInheritanceTests`) have zero file overlap with this diff and pass cleanly in isolation (18/18, 6/6) — pre-existing parallel-load flakes.
- Provider package tests, debug package build, `verify-ledger` (3200, +6 rows): all pass.
- **Local stress evidence (not committed):** the fail-closed scenario run 100 iterations in a single process with fresh run IDs — 100/100 passes, 0 failures, 54.1s; per-iteration asserts covered run failure, zero `startUserTurn`, released controller, and zero leaked gates/waiters/policies. Happy to add it as an `RPCE_ENABLE_BENCHMARK_TESTS`-gated suite if you'd like it in-repo.


provencher:
I haven't encountered this problem at all. Is your codex cli on latest? I suspect it happens if you force enable mcp servers that take too long to initialize. Rp defaults to only using rp.

mplibunao:
Codex CLI is 0.144.1 (latest, same as when we discussed #449).

On the config theory: fair point to check. My global codex config does have other MCP servers enabled (docker-based terraform/slack), and I'll test with those disabled on the codex path. However, occurrences today don't fit config latency: the **context_builder internal discovery agent** spawned without any RP tools (no git/read_file/selection tools, produced an empty selection, couldn't build review context) — and that reproduces with both discovery backends. The cleanest counterexample is the Claude-backed run, which launches with the leased `--mcp-config` + `--strict-mcp-config` containing *only* RepoPrompt — no user-configured MCP servers are in that child's config at all.

What we've observed across ~20 occurrences: failures cluster in windows (one window produced 3 consecutive tool-less spawns; children spawned before the window kept their tools throughout), with zero repoprompt-mcp crash reports during those windows. Restarting the client doesn't clear it; restarting the app does. That pattern reads as app-side connection-admission state rather than per-spawn config latency.

On scope: this PR deliberately doesn't claim to fix the trigger — it only makes the failure visible at `agent_run` instead of reporting success for a tool-less child. If you'd prefer different product behavior on timeout (e.g. warn-and-continue rather than fail), happy to rework it that way; the observability is the part we care about. Can share screenshots/diagnostics from the next occurrence if useful.

PR #545, opened 2026-07-14
Complete Agent Models workspace follow-ups

### LINKED ISSUE #508: Agent Models: complete workspace-scoped wizard and pin-state follow-ups after #275
## Summary

Follow-up to #275, merged as `6ceafd393ea7630dfee65dd5e8d2c9f109527f27`.

@crstdr — thanks again for the substantial scoped Agent Models work. The prior stored-pin blocker was resolved before merge. Final review identified several remaining consistency and hardening items that should be tracked together.

### 1. Recommendation wizard remains global-only

`RecommendationWizardViewModel` computes recommendations without an explicit `AgentModelsEditingScope`, so `AutoRecommendationEngine` uses its `.global` default. The wizard’s apply paths similarly omit scope.

Relevant areas:

- `RecommendationWizardViewModel.recomputeState`
- `applyCurrentStep`
- `applyAllRecommendations`
- `AutoRecommendationEngine.computeRecommendations`
- recommendation apply helpers

For a workspace using `.useWorkspaceOverrides`, the wizard can:

- evaluate satisfaction against the global profile rather than the active workspace profile;
- update global settings that are shadowed by workspace overrides, producing no visible active-workspace change;
- associate wizard completion or notification bookkeeping with a workspace even though the applied profile was global;
- retain recommendations briefly across a debounced workspace switch.

The dedicated Agent Models settings flow already passes `editingScope`, so the wizard should either become scope-aware or be explicitly defined and presented as a global-only onboarding surface.

### 2. Compact popover hides same-as-recommended stored pins

The full Settings view distinguishes:

- `hasStoredOverride`
- `hasCustomOverride`

and exposes **Pinned to recommended**, **Clear Pin**, and section-level reset behavior.

The compact Agent Models popover still gates its reset affordance using `hasCustomOverride`. An explicit pin equal to the current recommendation therefore appears indistinguishable from recommendation tracking and cannot be cleared from that surface.

Relevant file:

- `Sources/RepoPrompt/Features/AgentMode/Views/AgentModelsPopoverView.swift`

### 3. Chat/Oracle synchronization invariant is not enforced at the store boundary

UI setters guard against mirroring blank model values, but global/workspace profile setters can still accept a profile where synchronization is enabled while the planning model is blank or differs from the compose model.

This is currently reachable programmatically rather than through normal UI, but the durable profile boundary should either normalize the invariant or document and validate the caller contract.

### 4. Workspace Agent Models profiles have no lifecycle pruning

`agentModelsSettingsByWorkspaceID` entries are durable but are not pruned when workspaces disappear. This matches some existing workspace-keyed settings behavior, but the expected lifecycle should be decided explicitly to avoid indefinite orphan accumulation.

Any pruning must be conservative and must not delete profiles merely because a workspace is temporarily unavailable.

### 5. Blocked future schemas use the corrupt-file telemetry fallback

`GlobalSettingsFileStore` correctly preserves and blocks future or incompatible schema documents.

However, `GlobalSettingsManager.load()` currently uses:

```swift
let loadedExistingDocument = fileExists ? try? fileStore.load() : nil
let existingFileWasCorrupt = fileExists && loadedExistingDocument == nil
```

This collapses deliberate future/incompatible-schema errors into the same manager-level “corrupt” classification and forces the telemetry mirror to `false`, even though the file was preserved intentionally and the UI receives the correct block reason.

This behavior predates #275, but the schema-v4 transition makes mixed-version development and downgrade scenarios more likely to exercise it.

### 6. Documentation and regression-test polish

- Document why the CE schema moved from 2 to 4 and the intended downgrade-blocking behavior.
- Add an explicit post-clear assertion that `effective == recommended`, rather than relying only on the stored/custom override flags to imply it.

## Acceptance criteria

- [ ] Recommendation computation and application use an explicit profile scope.
- [ ] Switching workspaces cannot apply cached recommendations under the wrong workspace/scope.
- [ ] Notifications accurately describe the scope that was changed.
- [ ] Every stored role pin has a visible clear/reset affordance in both Settings and the compact popover.
- [ ] The durable profile boundary enforces or explicitly validates the Chat/Oracle synchronization invariant.
- [ ] Workspace-profile lifecycle behavior is documented and either safely pruned or intentionally retained.
- [ ] Future/incompatible schema blocks are distinguished from corrupt-file failures when updating the telemetry mirror.
- [ ] Focused regression coverage exercises workspace overrides, same-as-recommended pins, workspace switching, and future-schema telemetry behavior.
- [ ] The schema-version skip and downgrade behavior are documented.

## Non-goals

- Do not introduce a lossy downgrade migration.
- Do not delete workspace profiles without a reliable lifecycle/identity signal.
- Do not redesign the broader recommendation system beyond the scope consistency needed for Agent Models profiles.

crstdr:
## Summary

- make the Setup Wizard compute and apply Agent Models recommendations against the active global/workspace profile, reject stale targets, and label the affected scopes truthfully
- expose and clear every stored role pin consistently, and validate synchronized Oracle/Compose profiles at the durable settings boundary
- distinguish blocked future/incompatible settings schemas from corruption in telemetry handling, with focused regression coverage and persistence documentation
- emit canonical notifications for every durable scope actually changed, including global MCP Presets combined with workspace Agent Models updates

## Current-main integration

This follow-up is based directly on current `main` after #534. It preserves #534's `AgentModelsOperationIdentity`, stale-action revision guards, exact-scope mutation boundaries, and service-tier normalization.

## Validation

- Oracle final review: **APPROVE**
- Fable 5 final review: **APPROVE**
- `RecommendationWizardScopedTargetTests`: 5/5 passed
- `AutoRecommendationEngineScopedSettingsTests`: 8/8 passed
- `MCPAgentRoleDefaultsServiceTests`: 9/9 passed
- `SettingsJSONOnlyPersistenceTests`: 75/75 passed
- exact test ledger verification: 3,257 IDs
- full `make dev-test`: one unrelated durable-artifact GC timing test failed; isolated suite rerun passed 9/9
- targeted SwiftFormat/SwiftLint on touched files: passed
- commit and push contribution preflights: passed
- hosted GitHub CI: all required checks passed; shard 3 passed on rerun

Repo-wide lint remains blocked by existing SwiftFormat baseline findings in untouched files.

Closes #508


crstdr:
- Synced this branch with `main` via normal two-parent merge `9acf049e` (no history rewrite). The sole conflict was the curated test ledger, resolved additively.
- Ledger review heads-up: 3 `WorkspaceProjectedPathSearchTests` rows are intentionally beyond issue-508 scope, reconciling pre-existing drift from mainline `c1429603` (2 method renames plus 1 new method that landed without ledger updates). `verify-ledger` now reports 0 missing / 0 stale.
- Validation: `verify-ledger`, six focused suites, provider tests, and both product builds passed locally; full-root validation completed with its sole unrelated DurableArtifact timing flake passing the isolated-suite rerun. All 8 hosted checks are green; initial shard 1/4 failures moved across files unchanged from main and passed on rerun.

crstdr:
**Review sign-off note for the main-sync merge (`9acf049e`):**

Before the merge commit was created, the resolved index (write-tree `126cfb5d1e32865d8d7d2c1299e57fd6a8c5c3ae`) was reviewed by two independent AI reviewers against the full evidence package (effective-diff audit, ledger reconciliation proof, and validation results):

- **Oracle (GPT-5.6 Sol Max): APPROVE** — no blockers against committing the captured index tree.
- **Fable 5 (independent session): APPROVE** — independently re-verified with read-only git commands: exact 17-path scope vs main, non-ledger blobs identical to the topic commit, `AgentChatTitlebarSafetyTests.swift` sentinel byte-identical to main, exact +18/−3 ledger delta (incl. the c1429603 drift repair with preserved scenario counts and reviewed metadata), and repair-row IDs matching the actual test methods on the merged tree.

The merge commit's tree equals the reviewed write-tree — no content changed after these approvals; subsequent steps (commit/push gates, CI) were procedural.

baron (approved):
Maintainer approval after focused Agent Models persistence/schema/migration/compatibility review, old-state restoration and workspace-integrity validation, independent exact-model second opinion, and successful exact-head hosted checks at a66bf84ac02dc438f21379fd459c442d4169d489.

baron:
Maintainer ship note (state-integrity follow-up)

During the focused persistence/schema/migration review, we found that Agent Models could durably store an impossible synchronized tuple: `syncChatModelWithOracle = true` while the Oracle/planning and Chat/compose models were blank, only one was present, or the two differed. That could survive relaunch, including in dormant workspace override profiles, and later unrelated settings saves could preserve the bad state.

Repair commit `5dc3fe39` centralizes and enforces the invariant that synchronization is enabled only when both model values are nonblank and equal:

- enabling sync snaps Chat to a valid Oracle model; it refuses to enable when Oracle is blank
- legacy scalar and MCP writes preserve the explicitly written model value but disable sync rather than blanking or silently overwriting the sibling into an invalid tuple
- global and workspace profile writes normalize invalid synchronized profiles and retain diagnostics for the attempted invalid state
- load/reload repairs old persisted invalid tuples for both global and dormant workspace profiles, saves the repaired sync flag, and preserves model/provider raw values for forward compatibility
- MCP `app_settings` now delegates to the same store behavior instead of maintaining a divergent sync implementation

Regression coverage exercises blank, one-sided, and divergent tuples; old-state restoration across relaunch and reload; dormant workspace correctness; legacy setters; MCP writes; notification behavior; and unknown/future raw-value preservation.

Final validated pair: base `41e8df8b901830ba27b40569d13afccea9b2f4a2`, head `a66bf84ac02dc438f21379fd459c442d4169d489`. All eight hosted checks passed on the exact head, including the rerun of app shard 1. An independent second-opinion review using `claudeCode:claude-fable-5:high` raised no unresolved material concerns.

PR #566, opened 2026-07-16
Preserve Context Builder pending context through routing

### LINKED ISSUE #562: Context Builder loses pending final context when a routed child exits before resolving tab context
## Summary

A Context Builder MCP run can route successfully, complete an orderly peer EOF teardown, and still fail with:

```text
Context Builder final context is missing for run <redacted> (connection none)
```

The deterministic trigger is a routed discovery child that initializes or lists tools, but exits without calling a RepoPrompt tool that resolves tab context. The frozen context remains pending instead of becoming connection-bound. Teardown therefore has nothing to detach, removes the run mapping, and final commit cannot recover the context.

Minimal and full requests reproduced the same failure on three runs.

## Environment

- RepoPrompt CE debug app / `rpce-cli-debug`
- App version: 1.0.29
- Commit: `0c9eb51151bd5c7f49879f61faabcf26e3eccaf2` on `main`
- Build provenance reported the current checkout and a clean tree
- macOS 26.3, Apple silicon
- Observed provider: Claude Code, although the lifecycle gap is provider-independent

## Reproduction

1. Start a Context Builder MCP run.
2. Allow the discovery child to connect and route successfully.
3. Have the child initialize/list MCP tools but make no RepoPrompt call that resolves tab context.
4. Let the child exit normally.
5. Observe finalization.

## Expected

The exact frozen context installed for the run remains attributable to that run and is committed even if the routed child does not make a context-resolving tool call.

## Actual

The run fails at final commit with `missingFinalContext` and `connection none`.

Bounded debug diagnostics for each reproduced run showed:

1. routing waiter registered
2. client identity observed
3. run route mapped
4. routing waiter signalled as routed
5. policy applied
6. route wait completed with `routed: true`
7. transport close accepted with `cause: peer_eof`, `initiator: peer`
8. teardown resolved with `detached: false`
9. terminal `missingFinalContext` with no connection ID

The connection accepted five protocol frames and dropped none.

## Confirmed mechanism

- Context Builder installs frozen tab context before starting the provider.
- When no connection exists yet, that context is queued by run/client.
- The pending exact-run context is currently bound to a connection only inside tab-context resolution, when the child calls a RepoPrompt tool that needs contextual state.
- Establishing `run_route_mapped` does not itself promote the pending context.
- Teardown detachment requires an existing run-to-connection mapping and connection-bound tab context.
- If the child exits before a context-resolving call, peer EOF removes the mapping while the context is still pending. Detachment returns false, and final commit has neither a live connection nor detached context.

Relevant owning code is in:

- `ContextBuilderAgentViewModel` run setup, routing wait, and final commit
- `MCPServerViewModel+TabContext` pending-context binding, detachment, and final-context lookup
- `MCPConnectionManager` discovery-run transport teardown

## Why this is not #444

#444, fixed by #448, covered non-orderly discovery-child close causes and stale/superseded run disposal. This reproduction uses an **orderly peer EOF** and reaches the generalized detach path, but `detached` is false because the pending context was never connection-bound.

The missing invariant is earlier: once routing establishes the exact run/connection association, final-context ownership must no longer depend on a later contextual tool call.

## Impact

A successful route and normal child exit can discard the entire Context Builder result at the final step. Retrying is not corrective because the same discovery behavior reproduces the same ownership gap.

## Acceptance criteria

- An exact pending run context becomes durably attributable when routing maps the run to a connection, or an equivalent ownership mechanism preserves it without waiting for a contextual tool call.
- A routed child that performs initialization/list only and then exits by orderly peer EOF still commits the frozen final context.
- Teardown and commit preserve exactly-once ownership with no stale context leak across runs or windows.
- Existing non-orderly-close and supersession behavior from #448 remains covered.
- Regression coverage includes:
  - route mapped, no context-resolving tool call, peer EOF
  - normal contextual tool call followed by peer EOF
  - missing/different run mapping
  - teardown and commit racing without duplicate detach or publication

## Related

- #170 — broader Context Builder transport/lifecycle tracker
- #444 / #448 — prior final-context teardown fix for a different close-state path

morluto:
## Summary

- promote exact queued Context Builder context when pending-policy routing establishes the run/connection mapping
- preserve it through orderly peer EOF even when the child never calls a context-resolving tool
- restore promoted context exactly once if tentative policy mapping rolls back
- preserve read-file auto-selection handover lineage for replacement connections
- add lifecycle, rollback, and reconnect regression coverage

## Root cause

Frozen run context was queued before the child connection existed, but it was only bound during a later context-resolving tool call. A child that initialized/listed tools and exited normally therefore reached teardown with no bound context to detach, and final commit reported missingFinalContext(connection none).

## Validation

- ContextBuilderRunLifecycleTests: 12/12 passed
- MCPAgentPolicyAdmissionRaceTests: 32/32 passed
- TabContextRoutingTests: 49/49 passed
- read-file auto-selection suite: 20/20 passed
- strict Swift format/lint: passed
- RepoPrompt coordinated build: passed
- exact-ID ledger verification: passed
- final Codex autoreview: clean

The full pr-ready suite was attempted and stopped on two unrelated tests outside this diff; both passed immediately when rerun in isolation.

Closes #562

### DIFF
--- a/Sources/RepoPrompt/Infrastructure/MCP/MCPConnectionManager.swift
+++ b/Sources/RepoPrompt/Infrastructure/MCP/MCPConnectionManager.swift
@@ -2955,7 +2955,8 @@ actor ServerNetworkManager {
     private func mapConnectionToRunIDForPendingPolicy(
         _ connectionID: UUID,
         runID: UUID,
-        windowID: Int
+        windowID: Int,
+        clientName: String
     ) async -> MCPServerViewModel.PendingPolicyRunIDMappingToken? {
         let token = await MainActor.run { () -> MCPServerViewModel.PendingPolicyRunIDMappingToken? in
             guard let window = WindowStatesManager.shared.window(withID: windowID) else {
@@ -2965,7 +2966,8 @@ actor ServerNetworkManager {
             return window.mcpServer.registerPendingPolicyRunIDMapping(
                 connectionID: connectionID,
                 runID: runID,
-                windowID: windowID
+                windowID: windowID,
+                clientName: clientName
             )
         }
         guard let token else {
@@ -10215,7 +10217,8 @@ actor ServerNetworkManager {
                 pendingPolicyRunIDMappingToken = await mapConnectionToRunIDForPendingPolicy(
                     connectionID,
                     runID: runID,
-                    windowID: policy.windowID
+                    windowID: policy.windowID,
+                    clientName: clientName
                 )
                 routed = pendingPolicyRunIDMappingToken != nil
             }
--- a/Sources/RepoPrompt/Infrastructure/MCP/ViewModels/MCPServerViewModel+TabContext.swift
+++ b/Sources/RepoPrompt/Infrastructure/MCP/ViewModels/MCPServerViewModel+TabContext.swift
@@ -185,6 +185,9 @@ extension MCPServerViewModel {
         let previousRunPrimaryConnectionID: UUID?
         let previousPendingPolicyTokenID: UUID?
         let previousWindowID: Int?
+        var promotedContext: TabContextSnapshot?
+        var promotedContextClientName: String?
+        var promotedContextWindowID: Int?
     }
 
     /// Ordered, immutable coordinator lanes displaced by one pending-policy replacement.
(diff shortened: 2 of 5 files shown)


baron:
Reviewed exact head `25d68027f4c1b3caab5c671c38b96fa0b5d2db76`. The promoted pending context is retained and restored under matching rollback guards; I found no source blocker. The PR conflicts with main, so rebase/resolve and rerun exact-head checks. The unrelated MCP code-structure failure is not attributed to this diff.

baron (approved):
Approved on exact head `c81b700dd29a4c9f0c773752932ea155d6d95491`. The pending Context Builder context is promoted into the exact run mapping, recorded in the mapping token, and restored only by a still-current rollback. The current-main synthetic integration is clean and exact-head checks are green.

PR #588, opened 2026-07-19
Fix Context Builder cancellation after discovery commit

baron:
## Summary
- record cancellation requests that arrive while Context Builder's final context commit is in flight instead of silently ignoring them
- apply deferred cancellation at the next safe boundary while preserving the existing exactly-once terminal authority
- return the immutable committed prompt and selection for direct MCP cancellation after commit, while keeping workspace/tab/window teardown as a cancellation error
- skip typed Oracle generation for cancelled or failed discovery outcomes

## Root cause
Claiming the final context commit made the run reject cancellation before it had terminally settled. If cancellation arrived in that window, the request was ignored. Even when the app retained the committed selection, the typed MCP result selected the immutable pre-run tab for cancelled/failed outcomes, which could expose a blank or stale prompt/selection.

## Correct behavior
- pre-commit cancellation remains immediate
- cancellation during the atomic commit is deferred rather than interrupting the commit
- the first cancellation request owns its waiter/history policy and the run settles exactly once
- direct MCP cancellation after commit returns a cancelled typed result containing the committed context
- workspace switch/removal and tab/window close retain cancellation-error semantics
- only run-owned resources are cancelled; the parent MCP transport remains untouched

## Validation
- `ContextBuilderRunLifecycleTests`: 15 passed
- `ContextBuilderMCPProgressTimelineTests`: 12 passed
- `make dev-lint`: passed
- commit and push contribution preflights: passed
- independent Fable 5 high plan and implementation reviews: no unresolved must-fix findings

## Scope
This intentionally does not add generic deadlines, new settlement coordinators, progress telemetry, selection-projection convergence changes, or transport recovery. The follow-up CB track will address the uninstrumented selection-ingress/startup/finalization waits and caller-visible phase observability separately. General discovery performance remains a distinct measurement-led optimization track.

### DIFF
--- a/Sources/RepoPrompt/Features/ContextBuilder/ViewModels/ContextBuilderAgentViewModel.swift
+++ b/Sources/RepoPrompt/Features/ContextBuilder/ViewModels/ContextBuilderAgentViewModel.swift
@@ -456,6 +456,10 @@ final class ContextBuilderAgentViewModel: ObservableObject {
             let committedTabSnapshotCaptured: (
                 (_ runID: UUID, _ snapshot: MCPServerViewModel.ContextBuilderCommittedTabSnapshot) -> Void
             )?
+            let afterCommittedTabSnapshotCaptured: (@MainActor @Sendable (
+                _ runID: UUID,
+                _ snapshot: MCPServerViewModel.ContextBuilderCommittedTabSnapshot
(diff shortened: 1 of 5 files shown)


provencher:
Baron the goat

PR #590, opened 2026-07-20
Improve CodeMap parsing concurrency

baron:
## Summary

- remove the global Tree-sitter execution lock that serialized independent CodeMap files
- keep cache initialization synchronized while parser, tree, and query-cursor state remains operation-local
- bound default CodeMap admission to the effective bulk permit capacity, preserving one foreground permit
- add real production-permit mixed-language overlap/golden coverage and an N>1 cancellation regression

## Performance

Five warmed production-path samples over 16 mixed Swift, TypeScript, Python, and Go artifacts:

| Metric | Before | After |
| --- | ---: | ---: |
| Median wall time | 8.59 s | 3.32 s |
| Throughput improvement | - | 2.59x |
| Median CPU time | 9.10 s | 9.24 s (+1.6%) |
| Observed parser concurrency | 1 | 3 |
| Process peak RSS | 84.49 MB | 89.15 MB (+5.5%) |

The effective CodeMap limit is max(1, min(4, activeProcessorCount) - 1), matching the existing bulk permit cap and retaining foreground capacity.

## Validation

- CodeMapArtifactBuildCoordinatorTests: passed
- WorkspaceCodemapGitCapabilityServiceTests: 17/17 passed in isolation
- PersistentAgentModeMCPReadFileConnectionTests: 20/20 passed in isolation
- PromptContextPreAssemblyServiceTests: 17/17 passed in isolation
- strict SwiftFormat and SwiftLint: passed
- authoritative test list and ledger verification: passed, 3,439 tests
- commit and push secret/guardrail preflights: passed

The full root-suite PR-ready run reported failures in the three suites above plus unrelated cross-suite state, but each reported suite passed independently afterward. The focused CodeMap suite also passed on the refreshed main base. No implementation scope was added for those unrelated cross-suite failures.

## Review

Claude Fable 5 High reviewed the implementation before and after its initial findings were fixed, then reviewed it again after refreshing to current main. Final review found no must-fix or should-fix findings.

## Notes

- cancellation, currentness, persistence, and publication logic are unchanged
- no parser pool or second worker pool was introduced
- shared cached Query execution was verified against the pinned SwiftTreeSitter implementation: each execution owns a fresh cursor and no query mutators are used


baron:
Closing after a current-main audit: this implementation is superseded. Current main already uses operation-local CodeMap parser/query state and aligns build concurrency with the bulk-permit policy. Keeping this stale rewrite open would duplicate the current architecture rather than add a remaining capability.

PR #606, opened 2026-07-22
Fix bundled Codex release signing

baron:
## Summary
- individually Developer-ID sign every manifest-owned bundled Codex Mach-O before signing the enclosing app
- preserve exact OpenAI package verification before signing and verify signature-normalized content digests afterward
- require complete Mach-O inventory and apply the same contract to Tip, stable, promotion, and public-update validation

## Root cause
PR #605 added pinned Codex runtimes, but the shared staged release signer did not individually sign their eight nested Mach-O files. Apple notarization submission `71b64ead-0d0e-43c3-9b20-effab7d85535` therefore returned `Invalid` in Tip run `29941666336`, despite the enclosing app passing local codesign verification.

## Validation
- `python3 Scripts/test_codex_runtime_artifact.py` (23 passed)
- `python3 Scripts/test_release_tooling.py` (76 passed)
- `python3 Scripts/test_release_promotion.py` (22 passed)
- official cached Codex packages re-verified for both architectures
- `bash Scripts/codex_vendor_guardrails.sh`
- shell/Python syntax checks and `git diff --check`
- `make dev-release-preflight` (ticket `afbd44be-d3db-4bce-96b4-d85395bcbc1a`)
- mandatory commit and push contribution preflights

The protected Developer ID + Apple notarization run remains the required end-to-end confirmation after merge.

### DIFF
--- a/Scripts/codex_runtime_artifact.py
+++ b/Scripts/codex_runtime_artifact.py
@@ -11,6 +11,7 @@ import platform
 import re
 import shutil
 import stat
+import struct
 import subprocess
 import sys
 import tarfile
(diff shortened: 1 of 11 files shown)


Scripts/codex_runtime_artifact.py
```diff
@@ -492,6 +687,49 @@ def acquire(args: argparse.Namespace, manifest: dict[str, Any]) -> None:
...
+    os.replace(temporary, manifest_path)
+    refreshed = load_manifest(manifest_path)
+    for target in targets:
+        verify_package(
+            cache_root / refreshed["version"] / target,
+            target,
+            refreshed,
+            args.lipo,
+            args.codesign,
+        )
```
w-winter: The one recommendable fix I'm seeing: this refresh command replaces the manifest before checking that the updated version is valid.  If validation fails, the manifest would be left partially updated.  The suggestion is to validate the temporary file first, and then replace the manifest only after everything succeeds.

w-winter (changes requested):

Scripts/codex_runtime_artifact.py
baron: Fixed in 97a2489b. The refresh now writes and validates the temporary candidate, re-verifies every selected cached package against it, and only then atomically replaces the tracked manifest. A regression test forces candidate re-verification to fail and asserts the original manifest remains byte-for-byte unchanged and the temp file is removed. Focused artifact/release tests, guardrails, commit/push preflights, and dev-release-preflight all pass.

w-winter: On re-review it LGTM

w-winter (approved):

PR #607, opened 2026-07-22
Stabilize synthetic watcher integration tests

baron:
## Summary
- stop the real FSEvents watcher before synthetic watcher payload scenarios
- keep publisher ingress attached so search coverage and slice rebasing remain exercised
- preserve cached-search qualification through the existing DEBUG-only override

## Root cause
Main CI run 29962478686 rebased the same worktree file mutation twice because the fixture combined a real watched filesystem mutation with an injected synthetic watcher event. This produced the repeated shard-4 range mismatch. Production behavior is unchanged.

## Validation
- `make dev-format`
- `make dev-test FILTER=PersistentAgentModeMCPReadFileConnectionTests` (20/20)
- `make dev-test FILTER=WorkspaceSavePreparationTests` (5/5)
- `make dev-lint`
- contribution commit/push preflights
- read-only review: no must-fix issues


w-winter:
Addressed review feedback with https://github.com/repoprompt/repoprompt-ce/pull/607/changes/134b7cd9a1699ac8caf8df08777614ae5f0bcf1d which updates the helper to fail immediately if synthetic watcher setup doesn’t succeed.  No remaining issues found

baron:
Thanks — exact-head CI is now fully green, including all four app shards and the Sentry-enabled build. Branch protection still reports REVIEW_REQUIRED; please submit a formal approval if you’re satisfied with the amended head so this can merge normally and unblock Tip.

w-winter (approved):

PR #608, opened 2026-07-23
Optimize Swift and TypeScript CodeMap pipelines

provencher:
## Summary

- add deterministic Swift and TypeScript/TSX file-to-CodeMap benchmark/reference coverage with exact query, capture, and artifact parity checks
- optimize Swift signature whitespace normalization, parameter type scanning, property type extraction, and repeated referenced-type inputs
- generalize the proven signature-normalization primitive to JavaScript/TypeScript/TSX with an ASCII no-op/rewrite path and exact Unicode fallback
- remove obsolete model-benchmark/PromptFactory and Neon syntax-highlighting infrastructure while preserving live app diagnostics, CodeMap parsing, Markdown highlighting, and the active SystemPromptService boundary

### Swift deterministic 64-file corpus

| Stage | Baseline | Retained | Delta |
|---|---:|---:|---:|
| Full source-to-CodeMap | 107.107 ms | 97.4485 ms | **-9.02%** |

Retained iterations:
- signature whitespace normalization: 107.107 → 104.660 ms (-2.29%)
- parameter type scanning: 104.660 → 101.2865 ms (-3.22%)
- property type extraction: 101.2865 → 98.874 ms (-2.38%)
- referenced-type input deduplication: 98.874 → 97.4485 ms (-1.44%)

### TypeScript / TSX synthetic pipeline

| Shape | Language | Baseline | Retained | Delta |
|---|---|---:|---:|---:|
| Full file-to-CodeMap | TS | 129.3660 ms | 126.7585 ms | **-2.02%** |
| Full file-to-CodeMap | TSX | 244.8285 ms | 240.6505 ms | **-1.71%** |
| Pre-materialized generator | TS | 66.4170 ms | 63.5775 ms | **-4.28%** |
| Pre-materialized generator | TSX | 121.9060 ms | 116.0640 ms | **-4.79%** |

The shared JSTS normalization leaf improved 15.5–19.7%. Exact TS/TSX captures and artifacts remained unchanged.

## Rejected experiments

- removing a legacy Swift type capture regressed the pipeline and was reverted
- a Swift TypeCleaner ASCII no-op gate measured only 0.10% under stable counterbalanced A/B and was reverted
- TS/TSX raw referenced-type dedup skipped hundreds of cleaner calls but regressed all wall-clock metrics by 6.5–8.9% and was fully reverted

## Validation

[...]

PR #609, opened 2026-07-23
Add per-root Code Map generation controls

provencher:
## Summary
- replace per-root `MAP …` row chrome with a compact, responsive workspace-header mapping tag inspired by the legacy Code Scan preference tag
- show live aggregate percentage in a clickable popover, with scrolling per-root progress and pause/resume controls
- publish the known root catalog candidate total as soon as mapping begins so percentages update during generation
- gate preload, retry, invalidation, setup, and explicit-demand paths while a loaded root is paused

## UI behavior
- the header shows `Mapping` plus a determinate circular indicator when space permits, and falls back to the circle alone at narrow widths
- clicking the indicator opens aggregate progress and per-repository status/controls
- plain root rows return to a compact single line; Git/worktree context remains on its existing second line
- paused roots do not distort the workspace percentage; mixed mapped/paused states are represented explicitly
- active progress is capped at 99%, reserving 100% for authoritative completion

## Pause semantics
Pausing is scoped to the current loaded root lifetime. It immediately revokes publication authority, cancels queued/retry/setup/demand work, and prevents new generation admission. Already-admitted engine transactions may drain to the existing safe currentness boundary but cannot publish as current. Resuming waits for cleanup and schedules one fresh preload; unloading or restarting resets the pause.

## Validation
- `make dev-test FILTER=AgentWorkspaceRootsSidebarStoreTests` — passed, 24 tests
- `make dev-test FILTER=CodemapPreloadTests` — passed, 16 tests
- `make dev-lint` — passed (`0/1391` files require formatting; SwiftLint strict passed)
- contribution commit and push preflights — passed, including redacted staged/outgoing secret scans and guardrails
- PR-ready full root suite from the rebased base — one unrelated failure remains in `CodexNativeSessionControllerGoalConfigTests.testAgentModeDefaultCarriesGoalFeatureConfigToStartAndResume`
- test-ledger verification — feature rows are present; current repository baseline still reports unrelated missing IDs and 0 stale IDs

## Debug note
The header-popover redesign was relaunched successfully with explicit ad-hoc debug signing and is ready for visual testing. Secure settings remain ephemeral for this run.


baron (dismissed):
**Request changes — production design looks solid; two pre-merge items.**

**1. Confirmed flaky test (hosted shard 3).** `Tests/RepoPromptTests/WorkspaceContext/CodemapBindingEngineProjectionTests.swift:117` asserts `observedStatuses.first == .unavailable(.jobNotScheduled)`, but `projectionSnapshotUpdates` intentionally uses `.bufferingNewest(1)` and the collector `Task` is unsynchronized. The initial value is buffered at subscription; if `scheduleProjectionPreload` publishes before the collector’s first `next()`, newest-1 drops that value and the first observation is `.pending(.scheduled)` — exactly what shard 3 reported. The `contains { .pending }` assertion has the same exposure if the collector wakes only after completion. Please make observation deterministic by consuming/acknowledging the initial value before scheduling, or assert only guarantees of a newest-1 stream (eventual `.authoritativeComplete` and no consecutive duplicates). Keeping newest-1 in production is correct; this is a test-only fix.

**2. Retarget to `main` before merge.** The stacked base `perf/swift-pipeline-deep-optimization` served its review purpose, but merging as-is lands on that retained feature branch rather than `main`. PR #608’s head is already an ancestor of `main`, so retargeting should preserve the identical one-commit/11-file content diff. Require green hosted checks on the retargeted merge ref.

A second independent review confirmed the production path is coherent: suspension is installed before detachment, publication authority is revoked synchronously, cleanup is awaited before resume, re-pause cancels an in-flight resume through the transition token, admission is gated across authority/preload/retry/setup/explicit and multi-ticket demand paths, and shutdown/detach/unload clean up the new streams and observer task.

Non-blocking notes only: pause is intentionally in-memory for the loaded-root lifetime; cancelled/superseded idle states can temporarily render `MAP …`; and `rowHasContextLine` is now unconditionally true and could be simplified later.

PR #615, opened 2026-07-23
feat(deep-plan): treat the context_builder export as a preservation baseline

w-winter:
Deep Plan was handing back thinner plans than `rp-build`, even though both route through the same `context_builder` plan pass and the same rich planning meta-prompt.

This has been due to Deep Plan's own "post-processing" telling the orchestrator to extract "architectural bones," make the plan *shorter* as it matured, and let the design critic cut for over-specification.  As a result the implementation details worked out often diligently by the first planner would get compressed back out before hand-off.

I have found better results from reframing the `context_builder` export as a preservation baseline instead of raw material to summarize, and from specifying in the prompt that the final plan must keep every supported implementation-relevant fact, decision, and verification step (while still allowing evidence-backed correction or removal and lossless consolidation -- or else it's not "Deep Plan" anymore).

Discussed in https://discord.com/channels/1262487703744675901/1527322119162040370


baron (approved):
Reviewed exact head a512eed50eebad12f646ef8b061b428500aa12b2 against current main. The Deep Plan v62 preservation/fidelity semantics, catalog metadata, and focused regression coverage are consistent. Independent engineer and pair reviews found no remaining blockers; style is green and the remaining exact-head hosted jobs are in progress.

baron (approved):
Approved after revalidation on refreshed head a40d252a. The branch now contains current main via a clean merge commit; prior RepoPrompt engineer and pair reviews are complete with no outstanding findings, and fresh exact-head hosted CI is running.

provencher (approved):
Looks good! I see a few Claude slopisms in here haha but it's all good

w-winter:
> Looks good! I see a few Claude slopisms in here haha but it's all good

Lol, so it goes - was an iterative and collaborative effort between Sol Max, Fable 5, and me (whose left cerebral cortex is already borked by too much LLM interaction)

PR #620, opened 2026-07-24
Replace codemap projections with root-local incremental graphs

provencher:
## Summary

- replace projection-based codemap publication with one incremental, root-local graph authority per exact root epoch
- keep the last committed graph queryable during ordinary updates, non-preemptive diff/resync installation, and watcher-gap reconciliation; destructive changes use cumulative security fences before completion publication
- move automatic selection to graph-native exact-root traversal with zero source-artifact discovery demand, target-only render demand, deterministic per-root partial aggregation, and receipt revalidation
- make `get_code_structure` a strict breaking five-field request contract: `paths`, `expand`, `depth`, `signatures`, and `max_tokens`; deprecated `scope`, nested/object `expand`, `limits.*`, aliases, and compatibility decoding are removed
- delete the legacy projection/preload/segment/seal/successor/publication-permit machinery and replace its tests with graph foundation, overlay, liveness, reconciliation, MCP, and automatic-selection coverage
- expose graph-native availability, coverage, revision, observed/applied lag, reconciliation deadline/attempt, commit cadence, and privacy-safe aggregate diagnostics

## Breaking MCP change

`get_code_structure` now accepts only the five flat optional fields listed above. Unknown or deprecated fields are rejected before execution. `max_tokens` is the only clamped field; `depth`, `expand`, `signatures`, and `paths` use strict validation.

## Validation

- contribution `commit` preflight: passed staged-index secret scan and repository guardrails
- contribution `push` preflight: passed clean-tree/outgoing-range review, outgoing secret scan, and repository guardrails
- coordinated SwiftFormat/SwiftLint: passed; 0 files require formatting
- coordinated product builds: `RepoPrompt` and `repoprompt-mcp` passed
- authoritative root/provider test lists and curated ledger verification: passed (3,424 live methods)
- focused graph-native/MCP/status suites: 177 tests passed
- non-disruptive existing-app CE smoke: passed
- full root suite: 3,417 tests executed, 11 skipped; the only failure was the independently reproduced, untouched `CodexNativeSessionControllerGoalConfigTests/testAgentModeDefaultCarriesGoalFeatureConfigToStartAndResume` (`thread/start` and `thread/resume` observed `Optional(false)` instead of `Optional(true)`)
- clean-tree `pr-ready`: outgoing safety/secret/guardrail and lint lanes passed, then stopped only on that same known unrelated full-suite failure; the outgoing diff does not touch its test file

## Scope

This change intentionally removes the legacy projection authority rather than retaining a runtime compatibility switch or dual-authority path.


provencher:
Superseded by clean replacement #621, reapplied as a single commit on the current `main` tip.

PR #616, opened 2026-07-23
fix(mcp): preserve caller binding during agent runs

w-winter:
Diagnosed the following downstream of recurring RPCE tool call failures when driven externally via MCP:
```
Error: [-32602] Invalid params: Explicit tab context hint for file_actions targets tab [...] but this connection is already bound to tab external [...] Clear or intentionally rebind the connection before targeting a different tab context
```
External `agent_run.start` calls have been transferring a caller connection’s durable tab binding from the source tab to the newly created child tab, which caused later source-hinted calls on the same connection to fail with a correct but unexpected binding-conflict error.

This patch prevents that: Agent Run now keeps the caller bound to its source context while the child remains independently addressable through its session ID.  Agent Explore continues to apply its existing request-binding policy, and explicit context mismatches remain fail-closed.

## Summary

- Preserve the caller connection’s source-tab binding throughout detached and non-detached external Agent Run starts
- Split the external run starter into explicit caller-preserving and request-binding-policy entry points over one shared implementation
- Remove durable binding capability from the Agent Run service contract and retain Agent Explore’s existing policy at the composition boundary
- Keep child polling, waiting, cancellation, steering, and response control session-ID based without rebinding the caller to the child tab
- Preserve existing explicit-hint mismatch rejection, Agent Manage behavior, and pre-/post-activation cleanup semantics
- Add production-wired regression coverage for detached starts, concurrent source selection, terminal waits, wait cancellation, preparation and dispatch failures, and mismatched hints
- Add seven reviewed caller-binding contracts covering eight scenarios to the authoritative test ledger

## Implementation Notes

- `AgentExternalMCPRunStarter` now exposes `startPreservingCallerBinding` for Agent Run and `startApplyingRequestBindingPolicy` for flows such as Agent Explore, with activation, configuration, dispatch, snapshot resolution, and cleanup remaining single-sourced
- `AgentRunMCPToolService` no longer accepts or stores a connection-binding closure, making the caller-affinity invariant structural rather than dependent on a no-op or restoration path
- Agent Explore captures the existing `bindCurrentRequestToTabIfPossible` policy in `MCPServerViewModel`, preserving configuration → binding-policy → dispatch ordering without expanding the Agent Run contract
- Child lifecycle ownership continues through `AgentMCPControlContext`, `AgentRunSessionStore`, and the returned session ID; no connection lock, snapshot-and-restore flow, compatibility branch, or fallback routing was introduced
[...]

PR #622, opened 2026-07-24
fix(agent-run): improve respond validation diagnostics

w-winter:
I have also noticed over the last ~5 weeks some recurring failures when agents use RPCE MCP to answer a pending Agent Run approval.  These failures cite an error that names a field the tool doesn't accept:

```
Error: [-32602] Invalid params: decision must be one of: accept, decline, cancel.
```

The request had sent `decision`, which `agent_run op=respond` silently ignores.  The canonical field is a top-level scalar `response`.  Because the ignored argument left the approval token empty, the failure surfaced as an invalid `decision` value rather than a missing `response`, and the identical message came back for nested `response` objects.  Stale identity was also opaque: `The pending interaction no longer matches interaction_id` named neither the submitted nor the current interaction, and didn't offer a way forward.

With this patch the respond contract becomes more self-describing at the boundary that owns it, with invalid requests now saying which field is canonical, what values that interaction accepts, that nothing was applied, and how to recover.

## Summary

- Validate approval responses through a canonical top-level scalar `response`, with distinct invalid-params errors for missing, empty, nested, and `decision`-bearing payloads
- Reject `decision` explicitly instead of ignoring it, and state in every rejection that no response was applied
- Name `response` rather than `decision` in allowed-value errors, and advertise only the options the live interaction variant actually supports
- Report a stale `interaction_id` with both the submitted and current identities plus a session-scoped `poll`/`wait` recovery instruction, while leaving the pending interaction untouched
- Emit a copyable `agent_run op=respond …` command carrying the real session and interaction IDs in waiting-approval output, with amendment guidance only where the variant accepts an amendment
- Update the `agent_run` schema and tool description so `response` documents the scalar approval contract instead of a decision token
- Carry the caller's raw response syntax across the service boundary through a discriminated payload so interaction-identity validation still runs before shape validation
- Add five reviewed contracts covering sixteen scenarios to the authoritative test ledger

## Implementation Notes

- `MCPInteractionResponsePayload` now carries `ResponseArgument` (`missing`, `scalar`, `nonScalar`) plus `containsDecisionArgument` in place of a pre-normalized `decisionRaw`, so the parsing layer preserves request shape rather than collapsing absent, empty, and non-scalar arguments into one indistinguishable value
- All approval canonicalization lives in one private `mcpCanonicalApprovalResponse` used only by the approval branch; question, instruction, and MCP elicitation paths keep their existing behavior, and `response` remains globally optional in the schema
- Ordering is preserved deliberately: the live MainActor interaction-identity guard still runs first, so a stale `interaction_id` never returns a shape error and never mutates pending state
- Waiting-approval output selects its lead-in line by interaction kind instead of appending the generic instruction and then retracting it, so the approval branch no longer depends on what preceding shared code emitted; rendered output is unchanged and the absence of the generic line is now asserted
- Diagnostics carry identities and canonical option lists only — approval reason, command, cwd, grant root, amendment, and detail values stay out of error text, asserted with privacy sentinels
- No routing behavior changed, and no `decision` alias, nested `response` compatibility shape, or automatic stale-ID retry was introduced
- The regression suite drives the production `AgentRunMCPToolService` against real controlled sessions across all four live approval variants (worktree merge review, permissions, provider command, provider file change) rather than a production-only test hook

## Review Approach

- The consolidated patch received a whole-patch integration review covering the identity-before-validation invariant, error-surface exactness, formatter and schema contracts, test determinism, and ledger semantics
- Review returned three test-evidence findings — an unproven stale-payload matrix, teardown that was not failure-safe, and a ledger fixture label that did not match the executable fixture — which were fixed and re-reviewed until the 5.6-Sol-Max reviewer and then finally my review approved

## Validation

Passed locally on the committed branch:

- `make dev-test FILTER=RepoPromptTests.AgentRunMCPToolServiceRespondDiagnosticsTests` — 3 tests
- `make dev-test FILTER=ToolOutputFormatter` — 26 tests
- `make dev-test FILTER=RepoPromptTests.ToolCatalogSnapshotTests` — 6 tests
- `make dev-lint` — SwiftFormat: 0/1,392 files require formatting; SwiftLint strict passed
- `make guardrails` — source layout, contributor allowlist, dependency notices, and pinned Codex artifact contracts passed
- Authoritative root test listing completed and recognized all five new ledger entries
- Commit preflight: whitespace, staged-index secret scan, and repository guardrails passed

Exact whole-ledger verification remains blocked by eleven unrelated existing test IDs absent from the ledger; this change introduces no stale IDs, and all five new entries are recognized.

The full root suite was also run on the committed branch with an isolated `HOME`, so the Git fixtures could not inherit the developer's global ignore configuration: 3,576 tests, 11 skipped, one failure. That failure is `CodexAgentModeCoordinatorLivenessTests/testManualTypedFallbackKeepsOptimisticBubbleAndDraft`, which reproduces identically on the parent commit `f56b2c0c` without this patch and passes when the test runs alone — a pre-existing suite-level failure unrelated to this change. No other test in the suite failed.


w-winter:
This PR, btw, is quite low-priority.  Issue is about 15% the severity of https://github.com/repoprompt/repoprompt-ce/pull/616

baron (commented):
Non-blocking review: the implementation looks sound overall, and an independent confirmation review found no blocking defect.

One small formatter edge case: when an approval snapshot has no labeled options, the fallback advertises `accept_with_amendment`, but the amendment guidance is gated on `labels.contains("accept_with_amendment")` and is therefore omitted in that branch. Deriving an effective fallback label list—or omitting that value from the fallback—would keep the guidance complete.

The broader parse-before-identity observation for malformed `skip`/`answers`/`content`/`meta` is pre-existing and outside this PR's scoped `response`/`decision` change; I view it as optional follow-up hardening, not a blocker.

All hosted checks are green. Deferring to you on whether to fold the formatter tweak into this PR or handle it separately.

w-winter:
> Non-blocking review: the implementation looks sound overall, and an independent confirmation review found no blocking defect.
> 
> One small formatter edge case: when an approval snapshot has no labeled options, the fallback advertises `accept_with_amendment`, but the amendment guidance is gated on `labels.contains("accept_with_amendment")` and is therefore omitted in that branch. Deriving an effective fallback label list—or omitting that value from the fallback—would keep the guidance complete.
> 
> The broader parse-before-identity observation for malformed `skip`/`answers`/`content`/`meta` is pre-existing and outside this PR's scoped `response`/`decision` change; I view it as optional follow-up hardening, not a blocker.
> 
> All hosted checks are green. Deferring to you on whether to fold the formatter tweak into this PR or handle it separately.

Done and looks ready to merge

provencher (approved):

PR #631, opened 2026-07-25
Fix Finder launch and install commands when only Xcode beta is installed

dah:
## Summary

Fix the Finder-based Launch and Install commands on Macs where Xcode 26 has been removed and Xcode 27 beta is the only compatible full Xcode installed.

- Add a shared resolver for a full Xcode containing the macOS 26 SDK or newer.
- Preserve a compatible explicit `DEVELOPER_DIR`.
- Otherwise use the selected full Xcode or discover an installed Xcode app, preferring stable releases before beta versions.
- Export the resolved developer directory to the launcher, conductor, and production installer without changing the system-wide `xcode-select` setting.
- Fail before performing lifecycle actions when no compatible Xcode is available, with an actionable error.
- Use the production packager’s known `.build/release` output instead of invoking SwiftPM a second time to locate the app.
- Add launcher and installer coverage for explicit selection, Command Line Tools fallback, beta discovery, stable-version preference, failure handling, environment propagation, and avoiding the redundant Swift invocation.
- Document the full-Xcode requirement and selection behavior.

## Problem

After removing Xcode 26 and installing Xcode 27 beta, the `.command` files could inherit a stale or Command Line Tools developer directory. This caused SwiftPM to fail while initializing its build system even though a compatible full Xcode was installed.

## Validation

- Manually ran `Launch RepoPrompt CE.command`; the debug app built and relaunched successfully.
- Manually ran `Install RepoPrompt CE Local Production.command`; the local self-signed production app built and installed successfully.
- Verified with Xcode 27 beta installed and Xcode 26 removed.

### DIFF
--- a/Launch RepoPrompt CE.command
+++ b/Launch RepoPrompt CE.command
@@ -3,6 +3,7 @@ set -uo pipefail
 
 ROOT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 CONDUCTOR="$ROOT_DIR/conductor"
+FULL_XCODE_RESOLVER="$ROOT_DIR/Scripts/resolve_full_xcode_developer_dir.sh"
 APP_ARGS=("$@")
 
 if ! command -v python3 >/dev/null 2>&1; then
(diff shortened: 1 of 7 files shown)


baron (approved):
Reviewed through the existing RepoPrompt CE engineer/control-plane/pair workflow. The clean rewritten head preserves the original seven-file patch exactly and fresh exact-head/current-main hosted CI is green.

PR #639, opened 2026-07-26
Headless runtime M0: freeze contracts and evidence gates

provencher:
## Summary

PR 1/8 of the headless MCP domain runtime stack.

- Freeze the canonical 27-tool catalog, 86 discriminated actions, 12 actionless schemas, defaults, typed-error behavior, ownership, dependencies, approvals, persistence/save classifications, and MainActor migration denominator.
- Add source- and executable-derived policy/capability/`tools/list` fixtures plus drift guards.
- Record the pinned SDK stdio terminal assessment and the private child endpoint/launch-token contract.
- Keep scope strictly to M0: no runtime target, production launch token, child launch wiring, provider migration, or later-milestone implementation.

## Evidence and validation

- `HeadlessMCPDomainRuntimeM0ContractTests`: 3 tests passed.
- `make dev-lint`: passed.
- Mandatory push preflight: passed, including guardrails and outgoing-range secret scan.
- Path-selected `pr-ready`: lint passed; the full root suite encountered one reproducible unrelated baseline failure in `CodexNativeSessionControllerGoalConfigTests/testAgentModeDefaultCarriesGoalFeatureConfigToStartAndResume` (`goals` false vs expected true). The M0 focused suite remains green and the failing boundary is untouched by this PR.

## Carried-forward empirical gates

- Packaged CLI credential access is not observed and must be measured before M5; until then the contract requires parent-owned secure storage with minimum-scope in-memory handoff.
- Live EditFlowPerf latency is not observed because no app lifecycle action was permitted; it must be measured before M2 no-regression decisions using a separately authorized, already-running CE debug app.


baron (changes requested):
Requesting changes after two independent review passes against head `53c02d3cebd467065d9c3111868c5c499bb62563`.

## Blocking

1. **The frozen manifest cites test IDs that do not exist.**

   In `Scripts/Fixtures/headless_mcp_domain_runtime_m0_contract.json`:

   - `approval.existing_test` names `WorkspaceApprovalCancellationTests/testCancellationSettlesPromptExactlyOnceAndResolvesDenied`, but that suite contains only `testCancellingAwaitedApprovalResolvesDeniedAndClearsOverlay`, `testCancellingQueuedApprovalLeavesActiveRequestPending`, and `testApprovalRequestedFromCancelledTaskResolvesDeniedWithoutPresentingOverlay`.
   - `sdk_stdio.fallback_evidence` names `ProcessLauncherDescriptorInheritanceTests/testProcessLauncherDoesNotInheritActiveUnixSocket`, while the actual method is `testActiveUnixSocketTransportIsNotInheritedAndObservesEOFWhileSpawnedChildRemainsAlive`.

   These are fields in the PR's machine-readable authority, yet the new tests never resolve citation IDs against the contract ledger. Please correct them and add a cheap ledger-existence guard for every `existing_test`, `fallback_evidence`, and baseline `evidence` reference so this class of drift cannot recur.

2. **The claimed resolved `tools/list` freeze tests a parallel reconstruction rather than the production authority.**

   `HeadlessMCPDomainRuntimeM0ContractTests.resolvedAdvertisedTools` recomputes filtering from static restricted/granted/gated policy constants. The real projection in `MCPConnectionManager` additionally depends on the global enabled state, `ToolAvailabilityStore.effectiveDisabledTools`, registered `ServiceRegistry.services`, connection policy hydration, and deduplication. A registration removal, connection-policy wiring regression, or new production filter can therefore leave every frozen profile green.

   Please exercise the actual ListTools/debug list projection under controlled policy/availability state, extract one shared projection authority, or explicitly downgrade the spec/fixture claim to “derived from policy constants; runtime composition not exercised.”

## Should fix or explicitly downgrade

3. **The EditFlowPerf artifact's substantive observations are not guarded.**

   The test checks only the five stage names and that `evidence`/`observations` dictionaries are non-null. It does not validate the recorded capacities, event ordering, workload matrices, counts, or durability names. Those values can become stale while the M0 suite stays green. Either bind them to shared executable authorities or classify them as unguarded informational snapshots.

## Follow-ups

- The custom MainActor and stored-property scanners have silent escape cases (`@Observable @MainActor`, property wrappers, `private(set) var`) that are inconsistent with the spec's “complete denominator” / “cannot silently weaken” wording.
- Ledger `scenario_count` values `4/5/5` appear to group broad domains rather than distinct inputs, boundaries, fixtures, and outcomes as required by `docs/testing.md`.

The direction and fail-closed scoping are strong, and most catalog/policy assertions do bind to real authorities. The issues above matter because later M1–M7 work will treat this record as the baseline; the baseline should not begin with invalid citations or a parallel `tools/list` authority.

baron (approved):
Reviewed and repaired against exact head 1098fcd9. Contract evidence, ledger reconciliation, full local PR-ready validation, current-main merge, and fresh hosted checks are green.

PR #641, opened 2026-07-26
Show build numbers in Sparkle update alerts

baron:
## Summary
- format Sparkle's stock update alert with marketing version and build number for both the available and installed app
- preserve tip appcast comparison metadata while parsing the enriched tip title for display
- add regression coverage for the exact available/installed identities

For the reported example, the popup now reads:

> RepoPrompt CE.app v1.1.0 (31.11.89) is now available—you have 1.1.0 (31.10.88). Would you like to download it now?

## Validation
- make dev-format
- make dev-lint
- make dev-test FILTER=AppPlatformUtilityRecoveryTests (16 tests, 0 failures)
- focused release metadata tests (2 tests, 0 failures)
- make dev-release-preflight
- commit and push safety preflights

## Full-suite note
The PR-ready full root suite was attempted. Unrelated Codemap graph/invalidation tests timed out or failed after roughly 20 minutes, so the run was canceled after those failures were recorded. The focused update suite, app build, lint, and release preflight all pass.

### DIFF
--- a/Scripts/test_release_tooling.py
+++ b/Scripts/test_release_tooling.py
@@ -2994,6 +2994,10 @@ label_generated_tip_appcast""",
         sparkle = "http://www.andymatuschak.org/xml-namespaces/sparkle"
         self.assertEqual(item.findtext(f"{{{sparkle}}}version"), "29.8.52")
         self.assertEqual(item.findtext(f"{{{sparkle}}}shortVersionString"), "9.8.7")
+        self.assertNotEqual(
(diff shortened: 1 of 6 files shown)


baron:
Follow-up for the screenshot regression: removed Tip appcast releaseNotesLink/description content. Sparkle embeds releaseNotesLink targets inside its stock dialog, so pointing it at GitHub loaded the entire release page. Tip update dialogs now remain compact while retaining the marketing version + build number text. Added contract coverage that strips/rejects embedded release content. Focused appcast tests and release preflight pass.

PR #662, opened 2026-07-28
Add Codex Ultrafast public beta support

apollostreetcompany:
## What this adds

RepoPrompt now shows an **Ultrafast — Public Beta** option for Codex models that say they support it.

When someone chooses Ultrafast, RepoPrompt keeps that choice when:

- starting a new session
- reopening an existing session
- sending later messages in the same session

The exact `ultrafast` setting is sent to Codex instead of being dropped or replaced.

## What stays the same

- Ultrafast only appears when the Codex model list says it is supported.
- Models such as `codex-auto-review` do not get an Ultrafast option.
- The existing Fast option still works as before.
- Choosing no speed option still uses the normal Codex default.
- If Codex rejects Ultrafast, RepoPrompt shows that error instead of quietly switching models.
- Ultrafast keeps the model's recommended reasoning level unless the user chooses another one.

This PR does not add a setting for forcing Ultrafast on unsupported models. That is separate follow-up work.

## Testing

The focused model-picker and Codex session tests pass. They cover:

- which models show Ultrafast
- matching choices across RepoPrompt's model pickers
- new, reopened, and continuing sessions
- unchanged Fast and default behavior
- exact error handling without fallback
- keeping the model's recommended reasoning level

Formatting, linting, the Codex schema check, the RepoPrompt build, test-ledger verification, repository guardrails, secret scanning, and contribution checks also pass.

The required GPT-5.6 Sol High review found one issue: the first version could forget the model's recommended reasoning level for Ultrafast. Follow-up commit `2dae4eaf72e083d3825f88909b79f658830611b9` fixes it, the focused tests pass, and the follow-up review is clean.

## Known validation notes

- A full local test run was stopped after more than 23 minutes because unrelated worktree and Context Builder tests were failing or hanging. No Ultrafast-focused test failed.
- The non-disruptive live smoke check could not run because the local debug command-line tool is not installed. The app was not launched or stopped.
- Hosted checks are running on the latest commit. This PR should remain a draft until those checks finish.


baron:
Sorry for the miscommunication, wasted time, and tokens. Still appreciate you making the effort.

PR #686, opened 2026-07-30
Add independent secondary Oracle responses

dsebban:
## Summary

- add an optional Secondary Oracle model alongside the existing Primary Oracle
- run paired Oracle requests concurrently while preserving independent sessions and stable Primary-before-Secondary response shaping
- reuse the established single-Oracle send lifecycle for both lanes instead of introducing a second message-publication protocol
- persist pair identity, lane ownership, and lightweight divergence state across restart
- expose Secondary Oracle configuration in Agent Models settings and `app_settings`
- preserve the merge-base single-Oracle route and error ordering exactly when Secondary Oracle is disabled

## Feature preview

Secondary Oracle is optional and leaves the existing Primary-only path unchanged when disabled.

| Disabled | Enabled |
| --- | --- |
| Secondary Oracle disabled | Secondary Oracle enabled |

When enabled, each Oracle turn is sent concurrently to independent Primary and Secondary sessions. Results preserve Primary-before-Secondary ordering and remain unsynthesized so the caller decides how to use the second opinion.

### Thin runtime layer

When no Secondary model is configured, `tool_chatSend` returns directly through the existing single-Oracle implementation with the original arguments.

When enabled, the pair layer only:

1. resolves two independently owned sessions and models;
2. invokes the established single-send path once per lane;
3. coordinates cancellation and structured terminal outcomes;
4. returns stable Primary and Secondary results without automatic synthesis.

The earlier prepublished-user-turn protocol and pair-specific DEBUG failure hooks were removed.

### Durable pair lifecycle

- `oraclePairID` provides exact sibling identity across restart.
- `OracleLane` records explicit Primary/Secondary ownership.
- pair hydration requires exactly two opposite lanes and reports incomplete pairs as controlled conflicts.
- save, delete, retention, rollback, and bounded history loading operate on logical pairs rather than individual files.
- persistence re-resolves session identity after suspension and never retains array indices or force-unwraps mutable session lookups across `await`.
- divergence is derived from exact user-turn history and persisted symmetrically for lightweight post-restart reporting.

### Settings safety

- Secondary Oracle state uses a content-gated settings schema v5 so enabled state cannot be silently lost after rollback to an older build.
- leaving Secondary disabled does not raise the required schema.
- the settings store owns change notifications.
- `app_settings` accepts custom raw model identifiers consistently with adjacent model settings.
- the picker displays `Disabled` for no selection and preserves unknown custom raw identifiers instead of showing an unrelated fallback model.

## Safety properties

- invalid arguments fail before durable pair creation
- disabled mode performs no pair lookup, rerouting, divergence write, or extra persistence
- real provider/packaging failures and cancellation use structured lane outcomes rather than parsing rendered error text
- concurrent pair entry is rejected by a scoped claim that is released on success, failure, and cancellation
- incomplete pair state fails closed without crashing or mutating unrelated sessions
- atomic lifecycle operations preserve both members or roll back cleanly
- existing single-Oracle UI/MCP presentation remains unchanged

## Validation

Focused coordinated validation:

- `make dev-test FILTER=OraclePairCoordinatorTests` — 10 passed
- `make dev-test FILTER=MCPAskOracleWorktreeTests` — 26 passed
- `make dev-test FILTER=ChatHistoryJSONOnlyTests` — 8 passed
- `make dev-test FILTER=OracleMessageFinalisationHubTests` — 3 passed
- `make dev-test FILTER=AppSettingsMCPServiceAgentModeSettingsTests` — 6 passed
- `make dev-test FILTER=AIModelPreferenceRegressionTests` — 7 passed
- `make dev-test FILTER=SettingsJSONOnlyPersistenceTests` — 83 passed
- `make dev-lint`
- `make dev-test-list`
- contribution commit and push preflights, including guardrails and staged/outgoing secret scans
- coordinated ad-hoc debug package and live app launch from the exact PR head
- background visual QA with both CUA Driver and Peekaboo:
  - disabled picker and `Using: Disabled` agree
  - MCP enable refreshes the already-open Settings view and exposes `Disable`
  - UI `Disable` clears `models.secondary_oracle_model` back to `null`
- live dual-lane Agent Mode smoke:
  - one shared pair ID with distinct Primary and Secondary chat IDs
  - both lanes persisted under the same workspace, tab, Agent session, and run
  - both histories contain exactly one matching user turn and the expected response
  - `oracleHistoryDiverged=false`

Ledger verification reports only six unrelated pre-existing missing rows; all 12 new executable tests are recorded. The post-fix full root run encountered one unrelated `GitWorkspaceStateAuthorityTests` timing failure; the exact failing test passed on immediate coordinated rerun.

## Review follow-up

All 13 review threads have an individual response describing the applied fix or the evidence for an intentionally retained repository pattern. The threads are resolved; the current head is `6688ef59`.

## Ready for review

The requested review fixes, focused tests, live debug-app inspection, MCP setting round-trip, and end-to-end paired runtime smoke are complete. The original Secondary Oracle setting was restored to disabled after validation.


Sources/RepoPrompt/Features/Chat/Services/ChatSession.swift
```diff
@@ -1,5 +1,10 @@
 import Foundation
 
+enum OracleLane: String, Codable {
```
dsebban: Validate that is the idiomatic way to do enums in this repo , also an enum with 2 values it's a bit too much no ?

Sources/RepoPrompt/Features/AgentMode/ViewModels/UI/AgentModelsSettingsViewModel.swift
```diff
@@ -338,6 +358,12 @@ final class AgentModelsSettingsViewModel: ObservableObject {
         }
     }
 
+    func setSecondaryOracleModel(raw: String) {
+        updateSelectedProfile(reason: "agent_models.secondary_oracle_model") { profile in
+            profile.secondaryOracleModelRaw = raw.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty ? nil : raw
```
dsebban: Is this white space check nessecary ? The model should be set from an already validated list of models no ?

Sources/RepoPrompt/Features/AgentMode/ViewModels/UI/AgentModelsSettingsViewModel.swift
```diff
@@ -309,6 +317,18 @@ final class AgentModelsSettingsViewModel: ObservableObject {
         )
     }
 
+    var secondaryOracleModelDestination: ModelDestination {
+        ModelDestination(
+            id: "agentModels.secondaryOracle",
+            getter: { [weak self] in
```
dsebban, line 339: Validate this is idiomatic in this repo this looks a but over engineered

Sources/RepoPrompt/Features/Chat/Services/ChatHistoryManager.swift
```diff
@@ -94,6 +94,9 @@ actor ChatDataService {
         let composeTabID: UUID?
         let agentModeSessionID: UUID?
         let agentModeRunID: UUID?
+        let oraclePairID: UUID?
```
dsebban, line 119: Why do we need this pair id ? Looks overengineered

Sources/RepoPrompt/Features/Chat/ViewModels/Oracle/OracleViewModel.swift
```diff
@@ -355,6 +355,10 @@ class OracleViewModel: ObservableObject {
     @Published private(set) var streamingSessions: Set<UUID> = []
     @Published private(set) var messageStoreRevision: Int = 0
     @Published private(set) var currentQueryId: UUID?
+    var inFlightOraclePairClaimKeys: Set<UUID> = []
```
dsebban: This looks out of place compared to the surrounding code also a debug block seems not idiomatic in this project

Sources/RepoPrompt/Features/Chat/ViewModels/Oracle/OracleViewModel.swift
```diff
@@ -1891,16 +1895,18 @@ class OracleViewModel: ObservableObject {
     @discardableResult
     func startNewChatSession(
         name: String = "New Chat",
+        workspaceID: UUID? = nil,
```
dsebban, line 2010: Why is this needed ? It doesn't related to the oracles

Sources/RepoPrompt/Features/Chat/ViewModels/Oracle/OracleViewModel.swift
```diff
@@ -1891,16 +1895,18 @@ class OracleViewModel: ObservableObject {
...
         tabID: UUID? = nil,
         agentModeSessionID: UUID? = nil,
         agentModeRunID: UUID? = nil,
         activateInUI: Bool = true,
-        setActiveForTab: Bool = true
+        setActiveForTab: Bool = true,
+        reuseBlankSession: Bool = true
```
dsebban, line 2016: What's that ?

Sources/RepoPrompt/Features/Chat/ViewModels/Oracle/OracleViewModel.swift
```diff
@@ -1950,7 +1956,7 @@ class OracleViewModel: ObservableObject {
         let selectedChatPresetId = promptViewModel.selectedChatPresetID
 
         let newSession = ChatSession(
-            workspaceID: workspaceManager.activeWorkspace?.id,
+            workspaceID: workspaceID ?? workspaceManager.activeWorkspace?.id,
```
dsebban, line 2071: Is this really needed ?

Sources/RepoPrompt/Features/Chat/ViewModels/Oracle/OracleViewModel.swift
```diff
@@ -2472,6 +2478,9 @@ class OracleViewModel: ObservableObject {
                 sessionToSave.composeTabID = session.composeTabID
                 sessionToSave.agentModeSessionID = session.agentModeSessionID
                 sessionToSave.agentModeRunID = session.agentModeRunID
+                sessionToSave.oraclePairID = session.oraclePairID
+                sessionToSave.oracleLane = session.oracleLane
+                sessionToSave.oracleHistoryDiverged = session.oracleHistoryDiverged
```
dsebban, line 2595: What is this history diverged thing ? It seems overengineered

Sources/RepoPrompt/Features/Chat/ViewModels/Oracle/OracleViewModel.swift
```diff
@@ -2772,6 +2781,99 @@ class OracleViewModel: ObservableObject {
 
     // MARK: - Main Send/Receive Flow
 
+    struct PrepublishedOracleUserTurn {
```
dsebban: I would like the second oracle to be a thin layer reusing the first oracle existing code , evaluate this bug change if needed

Sources/RepoPrompt/Features/Chat/ViewModels/Oracle/OracleViewModel.swift
```diff
@@ -2811,18 +2914,24 @@ class OracleViewModel: ObservableObject {
...
-            sequenceIndex: nextSequenceIndex(for: targetSessionID)
-        )
-        withSessionMessages(targetSessionID) { msgs in
-            msgs.append(userMessage)
+        if let prepublishedOracleUserTurn {
+            guard prepublishedOracleUserTurn.sessionID == targetSessionID,
+                  messageStore[targetSessionID]?.last(where: { $0.isUser })?.id == prepublishedOracleUserTurn.messageID,
+                  messageStore[targetSessionID]?.first(where: { $0.id == prepublishedOracleUserTurn.messageID })?.content == newUserMessage
+            else {
+                return
```
dsebban: Looks like slop , maybe use switch to make it clearer ?

Sources/RepoPrompt/Features/Settings/Models/GlobalSettingsManager.swift
```diff
@@ -623,9 +625,14 @@ class GlobalSettingsStore: ObservableObject {
         var globalChanged = false
         var modelSelection = scalarPreferences.modelSelection ?? GlobalScalarPreferences.ModelSelectionSettings()
         let planning = normalized(modelSelection.planningModel)
+        let secondaryOracle = normalized(modelSelection.secondaryOracleModel)
         let compose = normalized(modelSelection.preferredComposeModel)
-        if planning != modelSelection.planningModel || compose != modelSelection.preferredComposeModel {
+        if planning != modelSelection.planningModel ||
```
dsebban, line 630: I would like to affect the current one oracle as much as possible , if secondary oracle is dhsabled there should be almost no new code paths as before

Sources/RepoPrompt/Features/Settings/Views/AgentModelsSettingsView.swift
```diff
@@ -388,6 +388,34 @@ struct AgentModelsSettingsView: View {
                     .font(.caption)
                     .foregroundColor(.secondary)
             }
+
+            Divider()
```
dsebban, line 393: Validate this is idiomatic ui for this repo

Sources/RepoPrompt/Features/Chat/Services/ChatSession.swift
dsebban: Addressed after reviewing nearby persistence patterns. I retained the two-case `OracleLane: String, Codable` enum intentionally: the lane is durable role identity, not just a transient Boolean, and it lets hydration require exactly one Primary and one Secondary without inferring authority from order/name. Legacy sessions still decode with `oracleLane == nil`, and the round-trip/corrupt-pair tests cover this contract.

Sources/RepoPrompt/Features/AgentMode/ViewModels/UI/AgentModelsSettingsViewModel.swift
dsebban: Addressed in `e7872535`. The view-model no longer performs its own whitespace/empty normalization; it assigns the raw value and delegates normalization to the durable profile/store authority. MCP input still keeps boundary trimming/length validation, where it belongs.

Sources/RepoPrompt/Features/AgentMode/ViewModels/UI/AgentModelsSettingsViewModel.swift
dsebban: Validated against the adjacent Primary Oracle and Built-in Chat rows. `ModelDestination` is the existing `AIModelDropdown` abstraction used immediately beside this code, so I retained it rather than moving persistence into the SwiftUI view. The Secondary destination now uses the same getter/applier pattern with no parallel settings authority.

Sources/RepoPrompt/Features/Chat/Services/ChatHistoryManager.swift
dsebban: Addressed by validating and hardening the lifecycle around this ID in `af01bdfd`. A durable pair ID is required to recover the exact sibling after restart and to make save/delete/retention operate on one logical conversation; workspace/tab/name matching would be heuristic. Pair hydration now requires exactly two opposite lanes and reports incomplete pairs as a controlled conflict.

Sources/RepoPrompt/Features/Chat/ViewModels/Oracle/OracleViewModel.swift
dsebban: Addressed in `af01bdfd`. The dead file-global debug logger and pair-specific mutable test/failure hooks were removed. Rollback is exercised through the real persistence/lifecycle seams instead of production runtime debug state.

Sources/RepoPrompt/Features/Chat/ViewModels/Oracle/OracleViewModel.swift
dsebban: Validated and retained intentionally. MCP resolves a concrete workspace before asynchronous pair creation; passing that frozen workspace prevents an intervening UI workspace switch from assigning the new lane to the wrong workspace. This is ownership safety for Oracle-created sessions, and the hardened persistence path revalidates identity after suspension.

Sources/RepoPrompt/Features/Chat/ViewModels/Oracle/OracleViewModel.swift
dsebban: This control prevents the ordinary blank-session reuse optimization from returning the same session for both Oracle lanes. Pair creation must produce two distinct session IDs; interactive creation keeps its existing reuse behavior. The pair lifecycle tests now cover distinct membership, rollback, and retry.

Sources/RepoPrompt/Features/Chat/ViewModels/Oracle/OracleViewModel.swift
dsebban: Validated and retained intentionally. The explicit resolved workspace must be applied at `ChatSession` construction time so the session is never transiently owned by whichever workspace happens to be active later. The runtime hardening also removed stale-index/snapshot writes across `await` and added identity-based revalidation.

Sources/RepoPrompt/Features/Chat/ViewModels/Oracle/OracleViewModel.swift
dsebban: Addressed and covered directly. This flag is the persisted lightweight projection that tells MCP/listing code the two user histories are no longer equivalent after a single-lane continuation; without it, every inspection after restart would have to load and compare both full transcripts. Divergence is derived from the exact user-turn sequence, written symmetrically, and tested for persistence/reconciliation.

Sources/RepoPrompt/Features/Chat/ViewModels/Oracle/OracleViewModel.swift
dsebban: Addressed in `af01bdfd`. Secondary Oracle is now orchestration around the existing single-send authority: disabled mode returns directly through the original single-Oracle path with unchanged arguments/error ordering; enabled mode resolves two sessions/models and invokes that same path twice. The alternate prepublication/send protocol was deleted.

Sources/RepoPrompt/Features/Chat/ViewModels/Oracle/OracleViewModel.swift
dsebban: Addressed by deleting the branch rather than restyling it. The `if let` represented an unnecessary second prepublished-message protocol with silent failure/race behavior; converting it to `switch` would only preserve the problem. Both lanes now use the ordinary send lifecycle.

Sources/RepoPrompt/Features/Settings/Models/GlobalSettingsManager.swift
dsebban: Addressed across `af01bdfd` and `e7872535`. With Secondary disabled, `tool_chatSend` now immediately returns through the merge-base single-Oracle path—no pair lookup, rerouting, divergence write, or extra persistence. The cited settings normalization is retained because `nil` is a no-op; schema v5 is content-gated so disabled state does not raise the persisted schema.

Sources/RepoPrompt/Features/Settings/Views/AgentModelsSettingsView.swift
dsebban: Validated against the existing `settingsCard`/`sectionHeader`/`AIModelDropdown` patterns and retained the established layout. The actual UI defect was fixed in `e7872535`: an empty Secondary selection now displays `Disabled`, and an unknown persisted custom model displays its raw identifier instead of falsely showing the first available model. Regression tests cover all fallbacks.

PR #696, opened 2026-07-31
Fix Oracle heartbeat liveness tracking

dsebban:
## Summary

- recognize content-free, non-final OpenAI-compatible chunks as transport activity
- propagate transport activity without mutating content, token, cleanup, or finalization state
- reset Oracle's existing inactivity watchdog while preserving the fixed overall deadline
- add focused classification, sanitization, lifecycle, and deadline regression coverage

## Why

Long-running Oracle requests can receive protocol heartbeat chunks without semantic text. Previously those chunks were discarded, so the local inactivity watchdog could report a stall after 10 minutes even though the provider connection was still alive.

## Validation

- `make dev-format`
- `make dev-lint`
- `make dev-test FILTER=OracleTransportActivityPropagationTests`
- `make dev-test FILTER=ContextBuilderFollowUpFinalizationMonitorTests`
- `make dev-test-list`
- `python3 Scripts/test_suite_optimizer.py verify-ledger --ledger Scripts/Fixtures/test-suite-contract-ledger.tsv`
- `ALLOW_ADHOC_SIGNING=1 make dev-build`
- live custom-provider Oracle review with `gpt-5.6-sol-pro`: completed in 890.5 seconds without the former 10-minute stall

`preflight.sh pr-ready` passed guardrails and lint, then the full root suite encountered unrelated MCP lifecycle timeouts because the intentionally running debug app owned the shared bootstrap socket (`lockUnavailable ... repoprompt-ce-D-7.sock.lock`). The focused changed suites pass.


dsebban:
Second-pass review — test adapter boundary

`OracleTransportActivityPropagationTests` currently exercises the extracted boolean/string predicate, not the `ChatCompletionChunkObject` adapter overload. A regression in `choices?.first` handling or field extraction (including accidentally restoring `choices.count == 1`) could therefore pass the table test. Add decoded-object coverage for zero/one/multiple choices and the represented delta fields, or narrow the ledger claim if constructing the SDK object is not maintainable.

dsebban:
Second-pass review — production stream wiring

The pure classification, sanitization, lifecycle mapping, and deadline-state tests do not drive `transport heartbeat -> semantic delta -> final output` through Oracle’s actual async consumption branch. Add the smallest existing test seam that proves a heartbeat records only `.streamActivity`, does not mutate content/tokens/cleanup/finality, and does not suppress the following semantic/final outputs. This would lock down the load-bearing early `continue` wiring.

dsebban:
Second-pass fixes pushed in `da00cbb4`:

- The real `ChatCompletionChunkObject` adapter is now tested via JSON decoding for zero/one/multiple choices, missing delta, every mapped semantic delta field, finish reason, and usage.
- Existing production-shaped canonical, linked-worktree, and visible-linked Oracle tests now emit a deliberately malicious transport-only frame before the final response and verify that it is ignored while fresh and continuing final responses still arrive.
- Test inventory and ledger were updated and verified (3706 tests).

Validation: focused adapter suite (4 tests), production-shaped transport variants (3 tests), authoritative test list, ledger verification, lint, and `RepoPrompt` product build all pass.

dsebban (commented):
Second-pass review complete. No remaining production correctness findings after `da00cbb4`. The two test-hardening findings raised above are addressed by decoded SDK-object coverage and production-shaped heartbeat-before-final coverage. One environmental limitation remains outside this PR: RPCE can propagate heartbeat frames but cannot prevent a 10-minute inactivity timeout when the upstream custom endpoint emits no frames at all.

baron (changes requested):
## P2 — Make the watchdog grace strictly exceed the verified heartbeat cadence (blocking)

`Sources/RepoPrompt/Features/Chat/ViewModels/Oracle/OracleViewModel.swift:1025-1027, 1199-1213, 1272-1277`

This changes `postContentGrace` from 10 seconds to 30 seconds, but the watchdog sleeps for exactly that grace and cancels when `elapsed >= grace`. If the upstream heartbeat cadence is at or near 30 seconds, the heartbeat and timeout race for MainActor execution: a heartbeat processed first refreshes the timestamp, while the timeout processed first cancels a still-live stream. `Task.sleep` cannot wake early and network/scheduler jitter biases the timeout check toward `elapsed >= 30`, so equality is not a safe liveness boundary.

The PR does not document the motivating endpoint's cadence, maximum observed gap, or jitter. The 30-second change also triples genuine post-content stall detection for providers that do not emit these decoded OpenAI heartbeat chunks, so it is a broader policy change rather than merely propagation of the new liveness signal.

Please either restore the existing grace if the measured heartbeat cadence is safely below it, or document the measured cadence and choose a grace with explicit margin above the maximum expected gap. Add deterministic coverage proving repeated heartbeats at `grace - epsilon` keep the production Oracle watchdog alive, the exact boundary cannot falsely cancel, and silence beyond `grace + epsilon` still cancels. The current `ContextBuilderFollowUpFinalizationState` test uses separate injected values and does not exercise this Oracle watchdog authority.

## P2 — Coalesce transport-activity events before the progress pipeline

`Sources/RepoPrompt/Features/Chat/ViewModels/Oracle/OracleViewModel.swift:1174-1179, 2977-2985`

`Sources/RepoPrompt/Infrastructure/MCP/WindowTools/MCPContextBuilderProgressTimeline.swift:199-200, 262-277, 359-377`

Every transport heartbeat currently performs a provider yield, service yield, MainActor hop, and lifecycle-observer/progress emission. `recordObservedStreamActivity` emits to observers before `armStreamInactivityWatchdogThrottled` applies its one-second throttle, so the watchdog tasks are throttled but the externally visible activity work is not. Unlike semantic output, these wire-level heartbeats bypass `AIQueriesService` batching.

The immediate Context Builder activity stream uses `.bufferingNewest(32)`, but the downstream progress timeline has an unbounded `pendingEvents` array and drains it with `removeFirst()`. If a custom endpoint produces a burst of empty deltas while the MCP progress sink is suspended, the new raw-heartbeat path can accumulate O(H) queued events and incur O(H^2) array shifting while draining, in addition to O(H) MainActor and network-progress work.

Please continue updating `lastAnyStreamActivityAt` for every accepted heartbeat, but coalesce or rate-limit lifecycle/progress emission. Resetting the 10-minute inactivity deadline does not require one user-visible progress event per wire heartbeat.

## P3 — Correct the new test's ledger scenario count

`Scripts/Fixtures/test-suite-contract-ledger.tsv:2238`

`Tests/RepoPromptTests/AI/OracleTransportActivityPropagationTests.swift:93-123`

The ledger records `scenario_count=1` for `testAIQueriesServiceSanitizesTransportActivityOutput`, but the test covers two distinct inputs and boundaries: sanitizing a transport-activity result and rejecting a normal content result. Under the repository's scenario-count rules this should be `2`.

dsebban:
Addressed every blocking item in review 4827008147 in `dbc2839b`:

1. **Watchdog grace and exact boundary**
   - Restored the established 10-second post-content grace because the PR had no measured cadence evidence supporting a broader 30-second policy.
   - The production watchdog now fires only when `elapsed > grace`, so activity at the exact boundary cannot be cancelled by the former `>=` equality check.
   - Added `testOraclePostContentWatchdogUsesStrictGraceBoundary`, using the exact production constant and predicate to cover recurring heartbeats at `grace - epsilon`, the exact boundary, and silence at `grace + epsilon`.

2. **Transport-activity coalescing**
   - Every accepted heartbeat still updates `lastAnyStreamActivityAt`.
   - Lifecycle/progress emission now shares the existing one-second watchdog re-arm throttle, so heartbeat bursts do not enter the Context Builder progress pipeline one-for-one. Semantic stream output remains batched by the existing service path.
   - Added `testOracleCoalescesTransportProgressWhileTrackingEveryHeartbeat`, which proves a 10-heartbeat burst retains the latest timestamp but emits one lifecycle event, then emits again after the throttle interval.

3. **Ledger scenario count**
   - Corrected `testAIQueriesServiceSanitizesTransportActivityOutput` from 1 to 2 scenarios and reviewed its ledger metadata.
   - Added surgical reviewed ledger rows for the two new production-authority tests.

Validation:
- `make dev-test FILTER=OracleTransportActivityPropagationTests` — 6 passed
- `make dev-test FILTER=ContextBuilderFollowUpFinalizationMonitorTests` — 6 passed
- ledger verification — 3,708 exact IDs
- `make dev-format`, `make dev-lint`, `make dev-swift-build PRODUCT=RepoPrompt` — passed
- commit/push contribution preflights — passed

The path-selected full root `pr-ready` lane again encountered the already-documented live CE app bootstrap-socket conflict (`repoprompt-ce-D-7.sock.lock`) and timed out in `ContextBuilderRunLifecycleTests`; I did not stop or relaunch the visible app. The affected focused suites above pass.

baron (approved):
Thanks for the thorough follow-up. The watchdog grace and strict boundary, transport-activity progress coalescing, and ledger scenario count are all addressed with focused production-authority coverage. I reviewed dbc2839b and found no remaining merge blockers. Approving.

PR #705, opened 2026-08-02
Stabilize ledger-free CI routing and codemap contracts

baron:
## Summary
- register authoritative domain workspaces and run routing explicitly only in persistent MCP tests that require them
- establish the filesystem mutation commit boundary after local conflict admission and before uncancellable I/O
- canonicalize app-managed worktree destinations through the mutation path fence so a retargeted symlink cannot escape the managed container
- replace a wall-clock/scheduling-dependent codemap deadline test with deterministic retry-exhaustion ownership cleanup coverage
- make Git-layout conversion readiness honor retryable `busy(retryAfter:)` results under one bounded deadline
- trim the stale non-Git proof regression to its unique local-reclassification-without-Git contract instead of duplicating repository conversion

## Context
Follow-up to #704. Removing the ledger changed suite ordering and process-admission overlap under native discovery/method-count sharding. That exposed missing test-only domain authority setup, a real mutation-boundary ordering gap, and pre-existing codemap tests that asserted incidental background scheduling rather than durable outcomes.

This PR fixes those contracts directly. It does not add assertion retries to the CI runner and does not reintroduce the test ledger or any local `swift test list` flow.

## Validation
- `MCPAskOracleWorktreeTests`: 25/25
- `MCPToolExecutionWatchdogIntegrationTests`: 23/23
- `ContextBuilderWorktreeInheritanceTests`: 9/9
- `CodemapAutomaticSelectionGraphNativeTests`: 17/17
- `WorkspaceCodemapLocalGitClassificationTests`: 11/11
- app-managed worktree symlink-boundary regression: 1/1
- `make dev-format`
- mandatory contribution commit and push preflights

Hosted CI is the final validation for cross-shard process-admission pressure.

### DIFF
--- a/Scripts/ci_app_test_runner.py
+++ b/Scripts/ci_app_test_runner.py
@@ -38,7 +38,6 @@ class OutputSnapshot:
 @dataclass
 class OutputState:
     output_seen: threading.Event = field(default_factory=threading.Event)
-    failure_seen: threading.Event = field(default_factory=threading.Event)
     lock: threading.Lock = field(default_factory=threading.Lock)
     first_failure_line: str | None = None
     last_started_test: str | None = None
@@ -55,7 +54,6 @@ class OutputState:
                 self.last_started_test = started_test
             if failure_line is not None and self.first_failure_line is None:
                 self.first_failure_line = failure_line
-                self.failure_seen.set()
 
     def snapshot(self) -> OutputSnapshot:
         with self.lock:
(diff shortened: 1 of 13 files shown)


baron:
Follow-up for the repeated shard failures:

- Shard 2: the shared AskOracle helper attempted to replace an immutable run-scoped binding when a continuation intentionally reused its endpoint. The fixture now registers domain authority explicitly and permits the immutable result only at that known reuse site.
- Shard 3: the filesystem path declared the domain mutation committed before acquiring the local in-flight mutation reservation. A rejected conflicting operation was therefore misclassified as `partialSuccessAfterCommit`. Commit transition now occurs centrally after local admission and before uncancellable I/O.
- Removed the overly broad fixture-wide domain registration; affected tests opt in explicitly.

Local validation before push: AskOracle 25/25, watchdog 23/23, ContextBuilder inheritance 9/9, and the symlink-boundary regression 1/1. Mandatory commit/push preflights passed.

PR #716, opened 2026-08-03
feat: show sessionless compose tabs in agent chats

w-winter:
MCP tools can create chat contexts containing file selections, prompts, and Context Builder work, but Agent Chats hides them until an agent runs. This makes MCP-created work difficult to revisit from the app and iterate on inside Compose.

This patch adds a default-off setting that exposes those chats in Agent Chats while keeping Agent-session behavior unchanged.

## Summary

- Add **Show chats created by MCP tools** under Settings -> Agent Mode -> Overview
- List open chats before an Agent session exists, including chats created through MCP
- Support selection, pinning, renaming, and deletion without creating or mutating Agent sessions
- Keep Agent-only status, stash, worktree metadata, archived sessions, and auto-archive behavior session-backed
- Preserve row identity when an Agent session later starts

## Implementation Notes

- Agent Chats uses a separate presentation projection and cache, leaving session-only sidebar consumers unchanged
- Session-less rows ignore unbound live-session metadata and expose only chat-owned state
- The preference is app-wide UI state and remains outside the `app_settings` catalog

## Review Approach

- Standard `rp-review` workflows were conducted using Fable 5 High (1x) and Sol Pro (2x)
- One integration review identified rename and unbound worktree-metadata leaks; both were fixed with regression coverage, and Sol Pro's followup review concluded Ship

## Validation

Passed locally:

- Focused sidebar builder, projection, action-boundary, and MCP settings tests
- `make dev-lint`
- `make guardrails`

I confirmed in the live CE debug app: default-off behavior, immediate toggle updates, MCP-created blank and fork chats with zero synthesized Agent sessions, and the expected session-less row actions.

The full root suite was also run; remaining failures were unrelated existing failures in Agent-run, Git-policy, transport, and workspace-materialization tests.

## Screenshots

Screenshot 2026-08-03 at 10 59 21 PM

### DIFF
--- a/Sources/RepoPrompt/Features/AgentMode/ViewModels/AgentModeSidebarSessionBuilder.swift
+++ b/Sources/RepoPrompt/Features/AgentMode/ViewModels/AgentModeSidebarSessionBuilder.swift
@@ -6,7 +6,7 @@ struct AgentModeSidebarSessionBuilder {
     typealias TabSession = AgentModeViewModel.TabSession
 
     let allTabs: [ComposeTabState]
-    let linkedTabs: [ComposeTabState]
+    let rowTabs: [ComposeTabState]
     let sessions: [UUID: TabSession]
     let authoritativeSessionIDByTabID: [UUID: UUID]
     let sessionIndex: [UUID: AgentSessionIndexEntry]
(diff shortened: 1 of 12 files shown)


baron (approved):
Approved at `62d14c07deb0deb6da3da2cdc4622231e9fcee0a`.

I reviewed the sessionless-row projection and action boundaries, Agent-session authority, stable row promotion, metadata isolation, and the focused regression coverage. The hosted CI run for this exact head is green, and I found no blocking correctness, state-safety, security, or performance issue.

Two non-blocking follow-ups remain worth considering: the visible Agent Chats ordering can differ from the existing keyboard/HUD navigation order, and the “created by MCP tools” copy is narrower than the implementation, which exposes all sessionless compose tabs. Neither concern invalidates the feature or its default-off behavior.

PR #722, opened 2026-08-04
fix: require authoritative context builder completion

w-winter:
Context Builder was able to treat ordinary Oracle message finalization as successful provider completion, such that a clean EOF / incomplete provider termination / interactive watchdog could consequently return partial assistant content (including a waiting message) as the final plan, question, or review result.

This issue has been especially noticeable for longer CB runs, and would cause the agent to continuously cancel and redo the oracle call.

This patch gives response-producing Context Builder calls a strict, exact-query completion boundary while preserving ordinary interactive chat behavior.

## Summary

- Require authoritative provider completion before active or headless Context Builder calls succeed
- Distinguish successful completion, incomplete provider termination, clean EOF, watchdog finalization, cancellation, and failure
- Prevent Context Builder queries from being completed by interactive inactivity watchdogs
- Preserve partial or failed Oracle chats for inspection while keeping Context Builder in an error state
- Keep ordinary interactive chat usable by finalizing available partial content after incomplete termination
- Normalize successful and incomplete terminal reasons across OpenAI, OpenRouter, Anthropic, and Azure providers
- Remove post-send query lookup and post-wait message re-read races by returning the exact query ID and completed response directly

## Implementation Notes

- Completion policy is installed before streaming begins, so fast provider events cannot race strict-mode configuration
- Finalization outcomes are recorded first-write-wins per query and consumed by the strict Context Builder waiter
- Provider-specific terminal reasons are normalized at adapter boundaries; Oracle receives provider-neutral completed or incomplete outcomes
- Active and inactive-workspace paths apply the same semantic completion requirement, with persistence occurring only after strict success
- No ordinary MCP wait redesign, compatibility layer, or broader provider architecture migration was introduced

## Review Approach

- The staged patch received iterative maintainability and integration reviews covering provider terminal semantics, exact-query ownership, concurrency safety, interactive behavior, persistence, and test quality
- Review findings identified incomplete terminal gates, unchecked error transfer, interactive partial-content regressions, and unstable test assumptions; each was addressed with focused regression coverage
- Optional waiter refactoring and broader completion taxonomies were excluded to keep the contribution bounded

## Validation

Passed locally on the committed branch:

- `make dev-test FILTER=ProviderStreamCompletionTests`
- `make dev-test FILTER=ContextBuilderStrictFinalizationTests` — 4 tests
- `make dev-test FILTER=OracleMessageFinalisationHubTests` — 3 tests
- `make dev-lint`

## Live App Validation

- Built and launched the patched CE debug app with `make dev-run`
- Completed a real and lengthy (>12 mins duration) Codex-CLI GPT-5.6-backed `context_builder` review request over MCP
- The complete review returned normally without premature finalization, retrying, or restarting the app or MCP connection

### DIFF
--- a/Sources/RepoPrompt/Features/Chat/ViewModels/Oracle/OracleViewModel+MCP.swift
+++ b/Sources/RepoPrompt/Features/Chat/ViewModels/Oracle/OracleViewModel+MCP.swift
@@ -1583,6 +1583,7 @@ extension OracleViewModel {
         finalReviewAuthorization: ContextBuilderFinalReviewAuthorization? = nil,
         agentModeSessionID: UUID? = nil,
         agentModeRunID: UUID? = nil,
+        completionPolicy: OracleResponseCompletionPolicy = .interactive,
         onProgress: ((_ text: String, _ reasoning: String?) -> Void)? = nil
     ) async throws -> ChatSendReply {
         // Check cancellation at entry
@@ -1667,8 +1668,8 @@ extension OracleViewModel {
         // (One Task.sleep for entire stream, not per-chunk - avoids CPU churn)
         let timeout: Duration = .seconds(4 * 60 * 60)
 
-        let (finalText, _, finalTokenInfo, providerCleanupHandle) = try await withThrowingTaskGroup(
-            of: (String, String, ChatTokenInfo, ProviderConversationCleanupHandle?).self
+        let (finalText, _, finalTokenInfo, providerCleanupHandle, terminalOutcome) = try await withThrowingTaskGroup(
+            of: (String, String, ChatTokenInfo, ProviderConversationCleanupHandle?, ChatStreamTerminalOutcome?).self
         ) { group in
             // Timeout task - throws after 4 hours
             group.addTask {
@@ -1682,6 +1683,7 @@ extension OracleViewModel {
                 var accReasoning = ""
                 var tokens = ChatTokenInfo()
                 var cleanupHandle: ProviderConversationCleanupHandle?
+                var terminalOutcome: ChatStreamTerminalOutcome?
                 var iterator = stream.makeAsyncIterator()
 
                 while let chunk = try await iterator.next() {
@@ -1706,8 +1708,12 @@ extension OracleViewModel {
                         let reasoning = accReasoning.isEmpty ? nil : accReasoning
                         await MainActor.run { onProgress(text, reasoning) }
                     }
+                    if let outcome = chunk.terminalOutcome {
+                        terminalOutcome = outcome
+                        break
+                    }
                 }
-                return (accText, accReasoning, tokens, cleanupHandle)
+                return (accText, accReasoning, tokens, cleanupHandle, terminalOutcome)
             }
 
             // Wait for stream to complete or timeout to fire
@@ -1716,8 +1722,22 @@ extension OracleViewModel {
             return result
         }
 
+        if completionPolicy == .contextBuilderStrict {
+            switch terminalOutcome {
+            case .completed:
+                break
+            case let .incomplete(reason):
+                throw OracleContextBuilderCompletionError.providerTerminatedIncomplete(reason: reason)
+            case nil:
+                throw OracleContextBuilderCompletionError.streamEndedWithoutProviderCompletion
+            }
+        }
+
         let trimmedResponse = finalText.trimmingCharacters(in: CharacterSet.whitespacesAndNewlines)
         guard !trimmedResponse.isEmpty else {
+            if completionPolicy == .contextBuilderStrict {
+                throw OracleContextBuilderCompletionError.emptyProcessedContent
+            }
             throw ChatToolError.internalError("Request produced no content.")
         }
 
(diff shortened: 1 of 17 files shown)


baron (changes requested):
Thanks for tackling the premature Context Builder completion path. The exact-query ownership and strict completion boundary are directionally right, but the current head still has one blocking correctness regression and one user-visible regression.

## Blocking finding

### P1 — Agent-facing `ask_oracle` can still return truncated output as success

`OpenAIProvider.swift:609-611` converts `responseIncomplete` into a terminal outcome rather than throwing. `OracleViewModel.swift:3230-3257` then preserves and finalizes the partial assistant message. However, `OracleViewModel+MCP.swift:1211-1227` waits only for message finalization and returns the message content with an empty `errors` array.

That means plan/question/review callers can receive a truncated answer as an authoritative success. This recreates the core failure mode at the agent-facing API boundary: the transcript is preserved correctly, but the caller loses the distinction between authoritative provider completion and incomplete finalization.

Please propagate the query-specific finalization outcome through `tool_chatSend` (and the corresponding headless path) so only an authoritative completed outcome is returned as success. Incomplete, EOF, watchdog, cancellation, and failure outcomes should remain distinguishable to callers while still retaining partial content for inspection.

Focused regression coverage should prove that:

- an incomplete provider termination preserves the partial transcript;
- `ask_oracle` / strict agent-facing calls do not return that partial content as a successful result with an empty error set;
- the completed path still returns the exact completed response;
- active and inactive/headless paths enforce the same outcome contract.

## User-visible regression

### P2 — Interactive chat no longer tells the user that the response was truncated

The same incomplete path finalizes available content without surfacing the incomplete reason in the chat UI (`OracleViewModel.swift:3230-3257`). Before this change, the provider error reached `handleSendMessageError`, which appended a visible error block (`OracleViewModel.swift:3413-3424`).

Keeping partial text is useful, but silently presenting it as a normal completed answer is misleading. Please add an explicit incomplete/truncated indication, or document and justify why losing that signal is an intentional product tradeoff.

## Non-blocking risks to keep explicit

- Strict Context Builder completion remains dependent on provider normalization; compatible providers that still emit a normal stop after truncation may bypass the intended semantic boundary.
- Treating a missing `finish_reason` / `stop_reason` as incomplete may reject otherwise complete replies from loosely compatible gateways.
- Strict queries disable the interactive inactivity watchdog and therefore rely on the Context Builder monitor for hung-stream recovery. That ownership should remain covered by a focused timeout/liveness test.

## Maintainer-guidance check

- **User impact and invariant:** Agent plan/question/review results must never report partial provider output as authoritative success; interactive users must be told when preserved content is incomplete.
- **Root-cause confidence:** Confirmed for the P1 static execution path; provider-specific normalization risks remain plausible and need behavioral coverage.
- **Authority:** The query-specific finalization outcome owned by the Oracle finalization hub, not message-presence or generic finalization alone.
- **State-safety:** Preserve partial transcripts, but do not collapse partial success into successful completion or discard the terminal reason.
- **Scale and observability:** No primary scale concern; incomplete termination must remain visible to both MCP callers and interactive users.
- **Recommended scope:** Fix P1 in this PR. Address P2 here or explicitly record the accepted UX tradeoff. Keep broader provider taxonomy work as a follow-up if it cannot be covered coherently.
- **Validation boundary:** Focused strict-finalization, MCP `ask_oracle`, headless/inactive-workspace, interactive incomplete-display, and monitor timeout tests.

Reviewed at exact PR head `05f3e58b92c433952a3a959e2aea1b29ef40ead3` with Claude Fable 5 High.

**Verdict: request changes for P1.**

w-winter:
> Thanks for tackling the premature Context Builder completion path. The exact-query ownership and strict completion boundary are directionally right, but the current head still has one blocking correctness regression and one user-visible regression.
> 
> ## Blocking finding
> [...]

Addressed these findings in https://github.com/repoprompt/repoprompt-ce/commit/725ac267d2c7c6f9288df1442e52207b2d540fda :
- `ask_oracle` now rejects non-authoritative outcomes while retaining partial transcripts
- interactive chat displays the incomplete provider reason

baron (approved):
Re-reviewed the exact follow-up head `725ac267d2c7c6f9288df1442e52207b2d540fda` against the two blockers from my earlier request-changes review.

Both are resolved:

- `ask_oracle` now captures the exact query ID and awaits the authoritative finalization outcome. Incomplete, failed, cancelled, watchdog, and non-authoritative outcomes propagate as failures; content is returned only for provider completion.
- Interactive incomplete responses retain useful partial content while appending a visible error with the provider termination reason before processing, finalization, and persistence.

The follow-up adds/updates targeted coverage for incomplete `ask_oracle`, successful `ask_oracle`, interactive partial-response disclosure, and retained Context Builder transcript behavior.

Validation performed against the exact commit: `git diff --check` and Swift syntax parsing of the changed production files passed. Focused XCTest could not execute in the isolated local snapshot because dependencies and the Sparkle artifact were unavailable there, so hosted exact-head CI remains the execution gate.

The original P1/P2 change request is cleared. Approving this exact head.

PR #724, opened 2026-08-05
Prevent cancelled CLI requests from stalling or spawning

### LINKED ISSUE #491: Cancelled CLIProcessRunner requests can remain queued and spawn child processes
## Impact

If the CLI process runner is busy and a user cancels the next request, that request can wait forever. If the busy request later releases the permit, the cancelled request can also be resumed and create a child process anyway.

That makes an abort or watchdog timeout unreliable: cancelled work may remain stuck, or it may still launch a command that should never have started.

## Expected behavior

Cancellation before process creation should finish the request as cancelled and prevent `ProcessLauncher.spawn(...)`. Cancellation after a child has been created may follow the caller's child-termination policy, but cancellation must never authorize creating new work.

## Actual behavior

`CLIProcessRunner.run(...)` is admitted through `TaskSemaphore`. On `main`, a caller waiting for a permit is stored in a checked continuation without a cancellation handler, and `withPermit(...)` does not recheck cancellation before entering its body.

The failure sequence is:

1. Request A holds the runner's only permit.
2. Request B waits for that permit.
3. Request B is cancelled.
4. Request A releases the permit.
5. Request B is resumed and continues into environment construction, command resolution, and process creation.

If Request A does not release the permit, Request B can remain suspended indefinitely instead. `cancelAll()` cannot reliably prevent this because a queued request is not yet in the spawned-process registry.

```mermaid
flowchart LR
    A["A holds the only permit"] --> B["B waits in TaskSemaphore"]
    B --> C["User cancels B"]
    C -->|no release| D["B stays parked indefinitely"]
    C -->|release occurs| E["B is resumed as if live"]
    E --> F["B reaches process creation"]
    C -->|required behavior| G["Remove B and resume as not acquired"]
    G --> H["CancellationError; no child process"]
```

## Evidence

Relevant code:

- `Sources/RepoPrompt/Infrastructure/Concurrency/TaskSemaphore.swift`
- `Sources/RepoPrompt/Infrastructure/Process/CLIProcessRunner.swift`

The sibling `runStreaming(...)` path already checks cancellation after admission, after environment construction, after command resolution, and immediately before spawning. That is the intended boundary for the non-streaming path as well.

No focused test covered a cancelled queued `run(...)` request.

## Regression scenario

1. Use a runner with concurrency one.
2. Start a command that holds the permit.
3. Queue a second command that writes a marker immediately after launch.
4. Cancel the second task.
5. Release the first command.
6. Assert that the second request resolves as cancelled, creates no marker or PID, and leaves the gate reusable.

Also cover cancellation during environment construction and command resolution with both child-cancellation policy values.

## Acceptance criteria

- A cancelled waiter completes promptly instead of remaining parked.
- Cancellation before process creation prevents `ProcessLauncher.spawn(...)`.
- `cancelAll()` cannot be followed by a previously cancelled request spawning.
- Admission races do not leak or double-release permits.
- The policy for already-spawned children remains explicit and independently tested.

## Related work

PR #724 implements the semaphore-side cancellation fix and the corresponding CLI call-site handling.

#481 concerns cleanup of an already-running cancelled process. This issue concerns cancelled work creating a process that should never start.

morluto:
Fixes #491.

### What was going wrong

`CLIProcessRunner` uses `TaskSemaphore` as its admission gate. If all permits are busy, a new request waits. Before this change, cancelling that waiting request did not wake it or remove it from the semaphore.

That produced two bad outcomes:

- the cancelled request could remain suspended indefinitely; or
- when a permit was eventually released, the cancelled request could resume, enter `withPermit(...)`, and reach process creation.

```mermaid
flowchart LR
    A["A holds the only permit"] --> B["B waits"]
    B --> C["B is cancelled"]
    C -->|before| D["B remains queued"]
    D --> E["A releases; B may continue"]
    E --> F["Cancelled request can spawn"]
    C -->|after| G["B is removed and resumed false"]
    G --> H["CancellationError; body is skipped"]
```

### Change

Make each queued semaphore waiter cancellation-aware and make admission return an explicit success/failure result.

- A cancelled waiter is removed and resumed with `false`.
- If cancellation races a permit handoff, the handoff is returned to the semaphore rather than consumed by cancelled work.
- `withPermit(...)` throws `CancellationError` before entering its body when admission fails or cancellation is observed.
- Manual `acquire()` callers handle a failed admission result.

This keeps cancellation ownership inside the semaphore and preserves permit accounting for later live requests.

### Regression coverage

`TaskSemaphoreTests.testCancellationCompletesAcquireWithoutWaitingForRelease` holds the only permit, queues a second task, cancels it, and requires that it complete before the held permit is released. It then releases the permit and verifies that a follow-up task can acquire it.

On the affected base behavior, the first completion assertion times out because the queued continuation remains parked; the test also has cleanup to avoid leaving that known-bad waiter suspended during teardown.

### Validation

- `git diff --check` — passed
- Swift/XCTest and repository guardrail execution require the macOS/Xcode toolchain; hosted macOS CI is the remaining build and test proof for this branch.

### DIFF
--- a/Sources/RepoPrompt/Infrastructure/Concurrency/TaskSemaphore.swift
+++ b/Sources/RepoPrompt/Infrastructure/Concurrency/TaskSemaphore.swift
@@ -5,12 +5,14 @@ import Foundation
 ///
 /// Usage:
 ///   let sem = TaskSemaphore(4)
-///   await sem.acquire()
+///   guard await sem.acquire() else { throw CancellationError() }
 ///   defer { await sem.release() }
 public actor TaskSemaphore {
+    private typealias Waiter = (id: UUID, continuation: CheckedContinuation<Bool, Never>)
+
     private let capacity: Int
     private var permits: Int
-    private var waiters: [CheckedContinuation<Void, Never>] = []
+    private var waiters: [Waiter] = []
 
     public init(_ permits: Int) {
         precondition(permits > 0, "Semaphore must have at least one permit")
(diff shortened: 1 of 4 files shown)


baron (approved):
Reviewed exact head 1c468a023b1077eded0c68af28d2af4e6a5b5a8c on the current-main synthetic merge.

The TaskSemaphore cancellation/release race is handled with single-resume actor-owned waiter state, the fast and handoff paths preserve permit accounting, and both production callers now honor failed acquisition. Exact-head hosted CI is fully green. No blocking findings.

Issue #492, opened 2026-07-11
Cancelled AsyncMutex waiters can still enter the critical section


morluto:
## Impact

`AsyncMutex` promises that a waiter cancelled before entering its critical section must not run that section. A cancellation/unlock race can break that promise.

That matters because the mutex protects ACP, Cursor, OpenCode, and Codex session operations. A cancelled operation can still perform the mutation or external work that cancellation was supposed to stop.

## Expected behavior

If a task is cancelled before it enters the critical section, `withLock(...)` must throw `CancellationError` and never execute the body.

Cancellation and `unlock()` must produce one terminal outcome for the waiter: either cancellation removes it, or unlock grants it. They must not both act on the same waiter, and a cancelled waiter must not receive a real lock handoff.

## Actual behavior in plain language

`AsyncMutex` has a queued-waiter path. When a waiter is cancelled, the cancellation handler does not remove it immediately; it schedules removal through a new task:

```swift
Task { await self?.removeCancelledWaiter(waiterID) }
```

That creates a window in which `unlock()` can win the race:

1. Task A owns the mutex.
2. Task B queues behind A.
3. B is cancelled, so removal is scheduled but has not run yet.
4. A calls `unlock()`.
5. `unlock()` removes B from the queue and resumes it with `true`.
6. The scheduled cancellation removal runs too late and finds no waiter.
7. `withLock(...)` trusts `true` and executes B's body.

```mermaid
sequenceDiagram
    participant A as Task A
    participant M as AsyncMutex
    participant B as Task B
    A->>M: acquire; A owns mutex
    B->>M: acquire; B is queued
    B->>B: cancellation requested
    B->>M: schedule removeCancelledWaiter(B)
    A->>M: unlock()
    M-->>B: remove B; resume(true)
    B->>M: withLock() continues
    M-->>B: critical section is entered
    B->>M: cancellation removal runs; B is already gone
```

This contradicts the implementation comment that a cancelled waiter “must NOT enter the critical section.”

## Root cause

The cancellation handler and `unlock()` can both race to decide the waiter's outcome. The resumed value is `true`, but there is no final cancellation check before `withLock(...)` enters the body.

## Evidence

Relevant code:

- `Sources/RepoPrompt/Infrastructure/Concurrency/AsyncMutex.swift`

Production callers include `ACPAgentSessionController`, `CursorACPLaunchResolver`, `OpenCodeACPLaunchResolver`, and `CodexNativeSessionController`.

[...]

PR #725, opened 2026-08-05
Prevent cancelled tasks from entering an unlocked AsyncMutex

morluto:
Related to #492; this PR fixes only the pre-cancelled unlocked fast path.

### What was going wrong

`AsyncMutex.withLock(...)` can acquire the mutex in two ways: it can take an unlocked mutex immediately, or it can wait in the queue. The queued path already had cancellation handling, but the immediate path did not check whether the task was cancelled.

Therefore, a task that was already cancelled could still acquire an unlocked mutex and execute the protected body.

```mermaid
sequenceDiagram
    participant T as Cancelled task
    participant M as AsyncMutex
    T->>T: cancellation requested
    T->>M: withLock()
    M->>M: fast path sees mutex is unlocked
    alt before this change
        M-->>T: grant lock
        T->>T: protected body executes
    else after this change
        M-->>T: reject acquisition
        T-->>T: withLock throws CancellationError
    end
```

### Change

Check cancellation before `AsyncMutex` claims an unlocked state. Normal acquisition remains unchanged, and the existing `withLock(...)` error path turns a failed acquisition into `CancellationError`.

### Regression coverage

`AsyncMutexTests.testCancelledTaskCannotAcquireUnlockedMutex` cancels the task before calling `withLock(...)`, verifies that the protected body is not acquired, and then verifies that a follow-up task can acquire and release the mutex.

### Scope

The queued cancellation/unlock handoff race in #492 remains a separate follow-up. This PR does not change queued waiter ordering or attempt to solve that race.

### Validation

- `git diff --check` — passed
- Swift/XCTest and repository guardrail execution require the macOS/Xcode toolchain; hosted macOS CI is the remaining build and test proof for this branch.

### DIFF
--- a/Sources/RepoPrompt/Infrastructure/AI/Providers/Codex/AppServer/CodexNativeSessionController.swift
+++ b/Sources/RepoPrompt/Infrastructure/AI/Providers/Codex/AppServer/CodexNativeSessionController.swift
@@ -1395,7 +1395,7 @@ final class CodexNativeSessionController {
             try markStartOrResumeSucceeded()
             return sessionRef
         } catch {
-            try? await eventHandlingMutex.withLock {
+            try? await eventHandlingMutex.withLockIgnoringCancellation {
                 cancelBindingSession()
             }
             markStartOrResumeFailed()
@@ -3817,6 +3817,12 @@ final class CodexNativeSessionController {
             }
             try? markStartOrResumeSucceeded()
             return sessionRef
+        }
+
+        func test_bindingBufferState() async throws -> (isBinding: Bool, bufferedCount: Int) {
+            try await eventHandlingMutex.withLock {
+                (isBindingSession, bufferedInbound.count)
+            }
         }
 
         func test_simulateTransportStreamEnded(source: String) async {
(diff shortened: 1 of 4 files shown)


baron (changes requested):
Reviewed exact head e44d0bd04aad923deedf363d4021723da50f534c on the current-main synthetic merge, with a Fable 5 High second opinion.

Blocking: Tests/RepoPromptTests/Infrastructure/Concurrency/AsyncMutexTests.swift:43 passes `try await cleanup.value` into `XCTAssertTrue`. XCTest's assertion autoclosure is not async, so this does not compile; it also triggers the formatter's hoistAwait rule. This deterministically fails Style and every app-test shard.

Please await into a local value first, assert that value, run formatting, and get exact-head CI green. I found no other blocker in the production change.

baron (approved):
Re-reviewed exact head `f1834fc58e5e2ff3e5a253e26e77d712d9d175cd` on current-main synthetic merge `bd24081966763aaac1b22459e1a90039b1b87ce9` (base `cc9ff8f5901c2d511eef2e52d45cb21df780bace`).

The prior blocker is fixed: the async cleanup result is awaited into a local value before the XCTest assertion. The incremental change is scoped to that test, and exact-head CI is green. No remaining blocker found.

PR #760, opened 2026-08-08
Add built-in workflow parity to direct headless mode

### LINKED ISSUE #758: Headless agent_manage advertises list_workflows but rejects it at runtime
## Summary

In RepoPrompt CE v1.2.0, the standalone/headless MCP tool schema advertises `agent_manage` operation `list_workflows`, but invoking that operation on the task-exclusive headless server returns:

```text
unsupported standalone agent_manage op: list_workflows
```

## Reproduction

1. Start the v1.2.0 standalone/headless MCP server.
2. Inspect its `tools/list` schema for `agent_manage`; `list_workflows` is present in the operation enum/help.
3. Invoke `agent_manage` with `{"op":"list_workflows"}`.
4. Observe the unsupported-operation error above.

## Expected behavior

Either the headless implementation should support `list_workflows`, or the standalone/headless schema should omit that operation so clients can capability-detect accurately.

## Impact

Clients cannot rely on the advertised schema to select this read-only discovery operation and must treat the resulting runtime rejection as capability drift.

No app-backed workflow behavior or source changes are required to reproduce the schema/runtime mismatch.

nickdavis:
## Summary

- move the provider-neutral built-in workflow catalog, metadata, and prompt rendering into `RepoPromptShared`
- implement direct-headless `agent_manage list_workflows` for the eight built-ins
- resolve `workflow_id` / `workflow_name`, reject conflicts and unknown values, and apply the canonical built-in template before provider launch
- keep custom-workflow parity and unrelated app/MCP behavior out of scope

## Tests

- `make dev-test FILTER=DirectHeadlessCompositionTests`
- `make dev-test FILTER=WorkflowPromptCatalogTests`
- `make dev-lint`
- `make dev-codex-schema-check`
- `/opt/homebrew/bin/python3 Scripts/test_release_tooling.py`
- `make dev-test`

Fixes #758.


baron (changes requested):
Reviewed the exact PR head `28e6f5c7b5fe1c3c49998beb6b40242e2bddf9db` on a clean local merge with current `origin/main` (`68fda448d1dfabdc28c6e0f8cec7a7c93de998a8`). Two parity issues need to be addressed before approval:

1. **Headless discovery returns noncanonical workflow IDs.** `Sources/RepoPromptMCP/DirectHeadlessCapabilityBackends.swift:793-800` emits raw IDs such as `build`, while app-backed discovery emits `builtin-build` via `AgentWorkflowDefinition.id`. The app resolver only accepts its exact IDs or display names, so a client persisting an ID discovered from headless mode cannot reuse it against the app backend. `DirectHeadlessCompositionTests.swift:91` currently codifies the incompatible raw IDs.

2. **Headless workflow launch ignores the persisted cleanup-guidance setting.** `Sources/RepoPromptMCP/DirectHeadlessProviderCoordinator.swift:421-425` always calls `wrapUserText` with the default `includeSessionCleanupGuidance: true`, even though direct-headless `app_settings` exposes `agent_mode.show_built_in_workflow_cleanup_guidance` (`Sources/RepoPromptDomainRuntime/DomainDirectSettings.swift:72`). Setting that preference to `false` therefore has no effect for selected built-in workflows.

Hosted checks are green on the PR head. A focused local `DirectHeadlessCompositionTests` run in the fresh current-main worktree timed out after one hour during dependency/build work before the tests completed; this is not being treated as a test failure.

nickdavis:
Addressed both review findings in `73c773a5`:

- Direct-headless discovery now emits the same canonical `builtin-*` IDs as app-backed discovery, while resolution continues to accept the existing raw aliases and display names.
- Direct-headless composition now shares one persisted settings store with `app_settings`; selected workflow rendering reads `agent_mode.show_built_in_workflow_cleanup_guidance` immediately before launch and applies it to the canonical template.

Added deterministic coverage for the exact eight-ID catalog, canonical/raw selectors, and an end-to-end captured provider prompt proving persisted cleanup guidance is omitted when disabled. Local validation passed: focused workflow tests, strict lint, full root suite, and both `RepoPrompt` / `repoprompt-mcp` builds.

baron (changes requested):
Re-reviewed exact head `73c773a5b3d0b78bd093cdecb3b4e723f6e6a239` on a clean merge with current `origin/main` (`5277d9765953c35c22683bf6671da45e3ceb3b84`). The two original findings—canonical `builtin-*` IDs and honoring the cleanup-guidance preference in the ordinary launch path—are resolved. One concurrency gap remains:

**[P2] Coalesce the initial settings load before reading it.** `DirectHeadlessProviderCoordinator.startAgent` calls `await settingsStore.bootstrap()` and then reads the preference (`Sources/RepoPromptMCP/DirectHeadlessProviderCoordinator.swift:107-111`). But `DomainDirectSettingsStore.bootstrap()` sets `didBootstrap = true` before awaiting `loadDirectSettingsData()` (`Sources/RepoPromptDomainRuntime/DomainDirectSettings.swift:177-182`). Because Swift actors are reentrant across that await, a second concurrent cold-start launch can see `didBootstrap == true`, skip the unfinished load, and read the default `true` instead of a persisted `false`. A concurrent initial settings write can likewise race the load/digest state.

Please either eagerly complete bootstrap before publishing the shared backends, or retain/await one shared bootstrap task, with deterministic concurrent cold-start coverage. Latest hosted checks are still running; this finding is independent of their result.

nickdavis:
Addressed the cold-start concurrency finding in `fa9c8c47` (with the implementation split across the three new commits since `73c773a5`).

- `DomainDirectSettingsStore` now retains one actor-owned bootstrap task. Concurrent readers and writers join that task, and `didBootstrap` is published only after a missing, valid, or degraded load has established the settings/digest state.
- The empty-profile path now completes bootstrap normally, preserving a nil digest and revision 1 for the first write.
- Deterministic tests gate the real load/join/publication events for concurrent read and write cold starts. All task completion waits are bounded, record results before assertion, and cancel without awaiting again after a timeout.
- Review-only instrumentation is module-internal, and the earlier CAS-adjacent async hook was removed so it cannot introduce a production reentrancy point.

Local validation on the exact pushed tree:

- `DomainDirectSettingsStoreTests`: 3 passed
- `DirectHeadlessCompositionTests`: 10 passed
- strict SwiftFormat + SwiftLint: passed
- `swift build --product repoprompt-mcp`: passed
- contribution commit/push guardrails and outgoing-range secret scan: passed

I also attempted the full local suite. It reached the same unrelated existing `WorktreeAPISmokeHarnessTests.testManageWorktreeAndAgentRunAPISmokeFlow` deadlock seen in the prior hosted shard rather than producing a failure in this change, so I am not claiming a full-suite local pass. The focused tests above were rerun in an isolated clean worktree against the exact pushed file hashes.

baron:
Re-reviewed exact head `fa9c8c47d86403ed41fbe29bbafcf41dc7479915`. The persisted-settings bootstrap race is addressed by retaining one shared bootstrap task and marking completion only after persisted state is loaded and published. I see no remaining source blocker from the prior findings. This head conflicts with current main; resolve it and rerun exact-head checks. The remaining managed-auth test failure is outside this diff.

nickdavis:
Resolved the current-main conflict in `f0decc98531597d82f8ef664e9428bc45f0d83d8` with a normal merge of `origin/main` (`3b732742ea14262f0ec4f06493a1e1e88945eccb`). The sole content conflict was in `DirectHeadlessCompositionTests.swift`; the resolution retains both main’s workspace-write/no-`--full-auto` assertion and all workflow-parity tests.

Exact merged-head local checks passed:
- `DomainDirectSettingsStoreTests`: 3 passed
- `DirectHeadlessCompositionTests`: 11 passed
- `WorkflowPromptCatalogTests`: 6 passed
- strict SwiftFormat + SwiftLint
- `repoprompt-mcp` product build
- contribution commit/push guardrails and outgoing-range secret scan

The PR diff against current main remains scoped to 34 files (+866/-164). Hosted exact-head checks are running. As noted in the review, I am not treating the unrelated managed-auth failure as part of this diff.

baron (approved):
Approved on exact head `f0decc98531597d82f8ef664e9428bc45f0d83d8`. The prior blockers are resolved: bootstrap is coalesced and only publishes completion after persisted settings load, and built-in workflow discovery/resolution now uses canonical `builtin-*` IDs while retaining compatible aliases. Exact-head checks are green and the current-main synthetic integration is clean.

baron (approved):
Approved on exact head `f0decc98531597d82f8ef664e9428bc45f0d83d8`. The prior blockers are resolved: bootstrap is coalesced and only publishes completion after persisted settings load, and built-in workflow discovery/resolution now uses canonical `builtin-*` IDs while retaining compatible aliases. Exact-head checks are green and the current-main synthetic integration is clean.

PR #815, opened 2026-08-13
Fix: decode legacy mutation journal status

w-winter:
An earlier RepoPrompt CE development build could persist protected mutation journal records with the status `failed_before_write`.  Because journal state survives build changes, current stable and Tip builds fail to decode those records, blocking unrelated MCP mutations such as `file_actions` and `apply_edits`.  Example shown in https://discord.com/channels/1262487703744675901/1365796959616630794/1537557255136739471 

This patch restores that schema-v1 read boundary while preserving the current canonical journal state.

## Summary

- Decode `failed_before_write` as the current retryable `failedBeforeCommit` state
- Continue writing only the canonical `failed_before_commit` value
- Keep unknown journal statuses fail-closed
- Cover restart, retry, and on-disk canonical rewrite behavior

## Review Approach

- The patch received focused maintainability reviews and final integration reviews from 5.6-Sol-Pro
- Historical status transitions and adjacent persistence paths were checked to confirm the mapping and scope

## Validation

Passed locally:

- `make dev-test FILTER=DomainProtectedMutationJournalTests` — 13 tests
- `make dev-lint`
- `make guardrails`
- Commit and exact outgoing-range secret scans

The full root suite was also run. Remaining Agent-run and Git-policy failures reproduced on unchanged `main`; all journal-focused tests passed.

### DIFF
--- a/Sources/RepoPromptDomainRuntime/DomainMutationJournal.swift
+++ b/Sources/RepoPromptDomainRuntime/DomainMutationJournal.swift
@@ -8,6 +8,29 @@ package enum DomainMutationJournalStatus: String, Codable, Sendable {
     case cancelledBeforeCommit = "cancelled_before_commit"
     case failedBeforeCommit = "failed_before_commit"
     case indeterminateAfterCommit = "indeterminate_after_commit"
+
+    package init(from decoder: Decoder) throws {
+        let container = try decoder.singleValueContainer()
+        let rawValue = try container.decode(String.self)
+        if rawValue == "failed_before_write" {
+            // Schema-v1 builds emitted this spelling. Delete this alias when v1 documents are no longer accepted;
+            // encoding below rewrites the record with the canonical current spelling.
+            self = .failedBeforeCommit
+            return
+        }
+        guard let status = Self(rawValue: rawValue) else {
+            throw DecodingError.dataCorruptedError(
+                in: container,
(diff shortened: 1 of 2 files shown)


w-winter:
I investigated the failing shard 2 and am confident that it's unrelated to this journal fix.  Looks like a preexisting shard-sensitive CI issue that got exposed by test rebalancing.

Adding two tests changed CI’s shard assignment, moving this existing worktree smoke test from shard 4 to shard 2.  It passed on the exact base commit and passes locally on both the base and PR branch, but timed out twice in shard 2 at an unrelated worktree-startup step.

baron (approved):
Reviewed exact current head `0f1d89abb5dada781c52b0928a462427ab9dd8a8` integrated with current `main` at synthetic merge `5f9052caa74db71cfae9d5f3f82c7ce0427e2786`.

No blocking code findings. The legacy `failed_before_write` alias maps narrowly to the retry-safe pre-commit state, canonical encoding remains `failed_before_commit`, and unknown statuses continue to fail closed. The focused `DomainProtectedMutationJournalTests` suite passed 13/13.

Non-blocking follow-up: consider exercising an unknown status through the complete persisted-journal denial path and asserting the mutation backend is not called.

The app-shard-2 Worktree API smoke timeout is outside this two-file patch; please require the usual exact-head CI disposition/rerun before merge.

PR #822, opened 2026-08-16
Fix: bound MCP structure settlement recovery

### LINKED ISSUE #818: Bug: a detached MCP read can indefinitely fence all structure tools for a window
## Summary

When `read_file`, `get_file_tree`, or `get_code_structure` exceeds its 30-second execution deadline and does not settle during the 5-second cancellation grace period, RepoPrompt CE detaches the provider and fences all three structure tools for that window.

The fence is removed only if the original provider eventually returns. If it never returns, every later structure call remains rejected for the lifetime of the app:

```text
tool_execution_structure_settlement_busy: A prior timed-out MCP operation for window <id> is still settling. Retry after it drains.
```

A new MCP client, rebinding, reconnecting, and waiting do not recover the affected window. Non-structure MCP operations remain usable.

## Impact

One stalled read-only MCP operation can make repository exploration unusable for every MCP client targeting that window. This affects unrelated sessions because settlement state is window-scoped rather than connection-scoped.

The response is also misleading:

- It reports the condition as retryable.
- It advises retrying after drain even though no maximum drain time or guaranteed recovery exists.
- It does not identify the original tool, invocation, connection, provider phase, or age.
- If the provider never settles, restarting RepoPrompt CE is the only guaranteed recovery.

## Environment

- RepoPrompt CE 1.3.0, build 35.14.70
- macOS 26.5.2
- Apple silicon
- MCP client over the bundled stdio bridge

## Observed behavior

A fresh MCP client connected to an existing RepoPrompt CE app and successfully:

1. listed windows;
2. bound to the target window and tab.

Its first `read_file` and `get_file_tree` calls then returned `tool_execution_structure_settlement_busy`. No structure operation in that client had previously timed out.

The same error persisted across repeated calls and after a 45-second wait. Routing and binding operations continued to work normally.

This proves that the new client inherited application-owned settlement state created by an earlier MCP invocation.

The exact originating invocation could not be recovered because the busy response omits its identity and no durable diagnostic for active detached settlements is exposed.

## Confirmed source behavior

The relevant implementation is in:

- `Sources/RepoPromptDomainRuntime/MCPDomainToolExecutionContract.swift`
- `Sources/RepoPromptDomainRuntime/MCPDomainToolExecutionWatchdog.swift`
- `Sources/RepoPromptDomainRuntime/MCPDomainCodeStructureSettlementRegistry.swift`
- `Sources/RepoPrompt/Infrastructure/MCP/MCPConnectionManager.swift`

`read_file`, `get_file_tree`, and `get_code_structure` use the `detachAndSettle` cleanup disposition.

After deadline and cancellation-grace expiry, the settlement slot becomes detached. `MCPCodeStructureSettlementRegistry.admit` then rejects every later detach-disposition structure call for the same window.

The registry removes a detached entry only when the original operation reports completion. It has no:

- maximum detached lifetime;
- terminal recovery transition;
- window-lifecycle cleanup;
- connection-lifecycle cleanup;
- administrative recovery operation.

PR #644 intentionally extended this policy to `read_file` and `get_file_tree` and notes that “a never-settling detached read can fence same-window detach-disposition reads until drain.” The observed incident is that documented unbounded case occurring in normal use.

## Deterministic maintainer reproduction

The existing watchdog integration fixtures already provide an uncooperative provider gate:

1. Start `read_file`, `get_file_tree`, or `get_code_structure` against a provider that does not return after cancellation.
2. Advance through the 30-second execution deadline and 5-second cancellation grace.
3. Observe the original call return `tool_execution_timeout` with `settlement: detached`.
4. Do not manually release the provider gate.
5. Call any of the other two structure tools in the same window.
6. Observe `tool_execution_structure_settlement_busy`.
7. Reconnect with a different MCP client and observe the same rejection.
8. Wait beyond the suggested retry interval and observe that the window remains fenced.

The current tests manually release the provider to prove eventual drain. The missing case is a provider that never returns and therefore never releases the window.

## Expected behavior

A timed-out read-only provider must not leave a window indefinitely unable to execute unrelated structure reads.

Recovery should be bounded and observable. Later clients should not inherit an immortal retryable-busy state whose only guaranteed recovery is restarting the app.

## Related issues

- #457 describes one possible provider-level trigger: an uncancellable selection-mirror chain that can make `read_file` time out. It is related but not a duplicate. This issue concerns the downstream watchdog and settlement policy regardless of which provider stage failed to settle.
- PR #644 introduced the current detach-and-fence behavior for `read_file` and `get_file_tree`.

No open or closed issue was found for `tool_execution_structure_settlement_busy`, `detached_settlement_in_progress`, or the indefinite window fence itself.

## Acceptance criteria

- [ ] A structure provider that ignores cancellation cannot leave `read_file`, `get_file_tree`, and `get_code_structure` unavailable indefinitely.
- [ ] The affected window becomes usable again through a bounded recovery path without restarting the entire app.
- [ ] Recovery does not permit unbounded accumulation of detached provider tasks.
- [ ] A busy or recovery response identifies the originating tool, invocation, connection, detached age, and last known handler phase.
- [ ] `retryable: true` is returned only when CE has a bounded recovery condition; otherwise the response gives an explicit terminal recovery action.
- [ ] Connection and window teardown cannot leave orphaned settlement entries that affect later clients.
- [ ] Regression coverage uses an operation that never returns after cancellation and proves both bounded recovery and successful subsequent structure calls.
- [ ] Regression coverage confirms that an affected window does not poison sibling windows.

w-winter:
This fixes https://github.com/repoprompt/repoprompt-ce/issues/818 (also [mentioned](https://discord.com/channels/1262487703744675901/1365796959616630794/1537839672762441950) on Discord), which is a bug where, whenever `read_file`, `get_file_tree`, or `get_code_structure` would time out and ignore cancellation, RepoPrompt CE could then leave all three tools blocked for that window until the app restarted (even for a new MCP client).  This patch gives that settlement fence a bounded recovery path while continuing to track the original provider.

## Summary

- Recover same-window structure tools 30 seconds after a provider detaches
- Keep late completion tied to the exact invocation so it cannot publish stale output or clear newer work
- Limit each window to one still-running provider after recovery
- Return non-retryable restart guidance if another provider becomes stuck before the first settles
- Include the originating tool, invocation, connection, age, remaining wait, and handler phase in busy responses
- Keep sibling windows usable throughout recovery

## Implementation Notes

- The settlement registry remains the single owner of admission and recovery state
- Providers remain tracked after they stop fencing new calls
- Recovery and escaped-provider accounting remain window-scoped

## Review Approach

- The patch received one focused maintainability review and complete-patch correctness reviews against the scope of #818
- Findings around terminal retry behavior, explicit registry inputs, and duplicate diagnostics were fixed with focused regressions

## Validation

Passed locally:

- `make dev-test FILTER=MCPToolExecutionWatchdog` — 41 tests
- `make dev-test FILTER=MCPCodeStructureSettlementRegistryTests` — 7 tests
- `make dev-lint`
- Commit and push safety preflights

The full root suite was also run. One Codex timing failure passed on focused rerun; remaining local failures were in unrelated Agent-run and Git-policy tests. All structure-settlement-focused tests passed.

### DIFF
--- a/Sources/RepoPrompt/Features/Diagnostics/MCP/MCPToolExecutionDiagnostics.swift
+++ b/Sources/RepoPrompt/Features/Diagnostics/MCP/MCPToolExecutionDiagnostics.swift
@@ -13,6 +13,12 @@ enum MCPToolExecutionHandlerPhase: String, Equatable {
     case fileActionsPostMutationCatalog = "file_actions.post_mutation_catalog"
     case fileActionsPostMutationSelection = "file_actions.post_mutation_selection"
     case fileActionsReplyConstruction = "file_actions.reply_construction"
+    case readFileRequestResolution = "read_file.request_resolution"
+    case readFileContentRead = "read_file.content_read"
+    case readFileAutoSelection = "read_file.auto_selection"
+    case getFileTreeRequestResolution = "get_file_tree.request_resolution"
+    case getFileTreeIngressWait = "get_file_tree.ingress_wait"
+    case getFileTreeConstruction = "get_file_tree.construction"
     // Graph-first get_code_structure execution stages.
     case getCodeStructureSeedResolution = "get_code_structure.seed_resolution"
     case getCodeStructureGraphSnapshot = "get_code_structure.graph_snapshot"
--- a/Sources/RepoPrompt/Infrastructure/MCP/MCPConnectionManager.swift
+++ b/Sources/RepoPrompt/Infrastructure/MCP/MCPConnectionManager.swift
@@ -11830,14 +11830,19 @@ actor ServerNetworkManager {
                                         switch self.codeStructureSettlementRegistry.admit(
                                             windowID: windowID,
                                             connectionID: connectionID,
-                                            invocationID: invocationID
+                                            invocationID: invocationID,
+                                            toolName: toolName,
+                                            now: executionWatchdogEnvironment.now(),
+                                            handlerPhase: {
+                                                handlerPhaseRecorder.snapshot()?.phase.rawValue
+                                            }
                                         ) {
                                         case let .admitted(slot):
                                             settlementAdmission = (.detachAndSettle, slot)
-                                        case let .busy(reason):
+                                        case let .busy(context):
                                             throw MCPToolExecutionDispatchError.structureSettlementBusy(
                                                 windowID: windowID,
-                                                reason: reason
+                                                context: context
                                             )
                                         }
                                     } else {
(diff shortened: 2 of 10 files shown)


baron (changes requested):
Reviewed exact PR head `734aeab1c31f7b752d751dd726151c219d5268d0` integrated with current `origin/main` `157f3464728ba8885d2d9021c2b2841aef54762a` at hosted merge `4a7ab71b00fa588080f5fc6533a713ea02510bc5`.

Two changes are required:

1. **Window teardown must retire settlement state.** `clearWindowSelectionIfClosed` clears routing and active-tool scopes but never clears `codeStructureSettlementRegistry` (`MCPConnectionManager.swift:2259-2268`). A cancellation-ignoring provider can therefore leave its registry entry, drain waiters, and captured diagnostic closure retained after the window closes (`MCPDomainCodeStructureSettlementRegistry.swift:198-200, 294-304`). Please add authoritative window retirement that removes the bucket, resumes waiters, safely ignores late callbacks, and covers detached/released/terminal teardown.

2. **Provide meaningful handler phases for all three affected tools.** The busy payload includes `last_handler_phase`, but the phase vocabulary and production reporting only cover `get_code_structure` (`MCPToolExecutionDiagnostics.swift:16-24`). `read_file` and `get_file_tree` have no phase reports in `MCPFileToolProvider.swift:305-490`, so the incident class originating from either tool still reports `unreported`. Please instrument their blocking stages and add deterministic origin-stall coverage for both tools.

Static review only; `git diff --check` passed. Hosted app shard 2 is currently red on an apparently unrelated `ContextBuilderRunLifecycleTests` assertion, and shard 1 remains pending.

w-winter:
> Reviewed exact PR head `734aeab1c31f7b752d751dd726151c219d5268d0` integrated with current `origin/main` `157f3464728ba8885d2d9021c2b2841aef54762a` at hosted merge `4a7ab71b00fa588080f5fc6533a713ea02510bc5`.
> 
> Two changes are required:
> 
> 1. **Window teardown must retire settlement state.** `clearWindowSelectionIfClosed` clears routing and active-tool scopes but never clears `codeStructureSettlementRegistry` (`MCPConnectionManager.swift:2259-2268`). A cancellation-ignoring provider can therefore leave its registry entry, drain waiters, and captured diagnostic closure retained after the window closes (`MCPDomainCodeStructureSettlementRegistry.swift:198-200, 294-304`). Please add authoritative window retirement that removes the bucket, resumes waiters, safely ignores late callbacks, and covers detached/released/terminal teardown.
> 2. **Provide meaningful handler phases for all three affected tools.** The busy payload includes `last_handler_phase`, but the phase vocabulary and production reporting only cover `get_code_structure` (`MCPToolExecutionDiagnostics.swift:16-24`). `read_file` and `get_file_tree` have no phase reports in `MCPFileToolProvider.swift:305-490`, so the incident class originating from either tool still reports `unreported`. Please instrument their blocking stages and add deterministic origin-stall coverage for both tools.
> 
> Static review only; `git diff --check` passed. Hosted app shard 2 is currently red on an apparently unrelated `ContextBuilderRunLifecycleTests` assertion, and shard 1 remains pending.

Thanks @baron

Agree that `read_file` and `get_file_tree` currently report `unreported`.  I'll scope the smallest useful phase coverage for those two providers.  I want to avoid introducing broad instrumentation for this issue

OTOH I don't believe window retirement is required for this bug.

Sol Max:
```
Window IDs are monotonic and never reused, so a closed-window bucket cannot affect later clients. The drain waiter is DEBUG-only, and removing the entry cannot terminate the escaped provider. I’d prefer to record registry retirement as non-blocking cleanup unless there is a concrete later-client failure path I’m missing.
```

baron:
Agreed. On re-check, the window-retirement point should not be a blocker for this PR. Window IDs are monotonic and not reused, the app drain-waiter surface is DEBUG-only, and removing the registry entry would not terminate the escaped provider. I’m happy to treat that as non-blocking cleanup rather than part of #818.

The narrowly scoped phase coverage for `read_file` and `get_file_tree` sounds right. Once that follow-up commit is pushed, I’ll re-review the new exact head and current CI.

baron (approved):
Re-reviewed exact head `569aa464d78b0c0a8973ded2d012dd92e621bf68` integrated with current `origin/main` `157f3464728ba8885d2d9021c2b2841aef54762a` at hosted merge `b681d8f8a396730e28c6fe28da6ba44360914556`.

The focused remediation fully addresses the remaining request: `read_file` and `get_file_tree` now report narrowly scoped production handler phases, with deterministic manual-clock/gate coverage for the blocking content-read and ingress-wait paths. I found no blocking correctness or scope issues. As discussed, registry retirement on window close is non-blocking cleanup rather than part of #818.

`git diff --check` passed. Seven hosted checks are green. App shard 3 remains red after reruns on the same `ContextBuilderRunLifecycleTests.testRealConnectionCleanupCannotEraseContextBeforeCommit` prompt-projection assertion; neither this PR nor the remediation touches that Context Builder lifecycle path, so I do not consider it a PR blocker. The red exact-head check still needs normal CI disposition before merge.

PR #825, opened 2026-08-16
Add Grok Build reasoning-effort selection

datawisebets:
## Summary

Follow-up to #817: exposes Grok Build's per-model reasoning efforts in Agent Mode.

Live-verified wire contract (grok 1.0.4): `availableModels[]._meta` carries `supportsReasoningEffort`, `reasoningEffort` (current/default), and `reasoningEfforts[{id, value, label, default}]` — grok-4.6 offers **xhigh/high/medium/low**, grok-4.5 **high/medium/low**. Effort applies via `session/set_model` with camelCase `_meta.reasoningEffort` (same-model and cross-model both verified); unknown efforts are **silently ignored** by the server, so validation is client-side against the advertised list.

- The SessionModelState parser reads the effort metadata, marks base options with supported/default efforts, and synthesizes effort-variant options with compound raws (`grok-4.6-low`) — the shared catalog, menus, and snapshot validation treat them as ordinary options.
- The controller decomposes compounds into base + effort, resolves bare picks to the model's advertised default (so bare-after-variant resets effort), skips the RPC when base+effort already match, and sends `modelId` + `_meta.reasoningEffort`.
- Fail-closed model authority from #817 is preserved: the Ok check compares the provider-declared base id (`expectedConfirmationModelRaw`), never the compound.
- Effort changes force a fresh controller via the existing requested-model reuse key; unadvertised compounds (`grok-4.5-xhigh`) throw before any RPC.
- `ACPDynamicProviderRecord` gains `currentEffortRaw` (optional; older persisted records decode unchanged).

## Verification

- 78 GrokBuild suite tests green, incl. 5 new effort cases: variant synthesis, wire meta, identical-set short-circuit, bare-after-variant reset, unadvertised-effort rejection, effort-keyed reuse
- Full local suite: no new failures (5 pre-existing git/codemap env failures unchanged)
- `make dev-lint` clean
- Live dogfood on a locally built beta: `grokBuild:grok-4.6-low` run applies effort via `session/set_model` meta (below)


baron (changes requested):
Requesting changes for the confirmed reasoning-effort parsing bug in `GrokBuildModelSpecifier`.

The snapshot parser accepts every `CodexReasoningEffort` and synthesizes selectable compounds for them, but the specifier only decomposes `-xhigh`, `-medium`, `-high`, and `-low`. If Grok advertises `none`, `minimal`, `max`, or `ultra`, a selection such as `grok-4.6-max` is therefore sent as `modelId: "grok-4.6-max"` without `_meta.reasoningEffort`, rather than base `grok-4.6` plus effort `max`.

Please derive decomposition from the advertised snapshot/base-effort association (or otherwise cover the full accepted effort set), and use the known base-model set so a literal advertised model ID ending in an effort token is not misparsed. Add focused wire-parameter coverage for the omitted efforts and for a base ID ending in a recognized suffix.

Reviewed exact head: `86436dcb07e59b731d4131072a5a008f354a33fb`.

baron:
Thanks for the remediation — the production path now uses structured effort provenance and the original suffix-parsing defect appears fixed. One narrow regression-coverage request remains before approval.

Please add a table-driven positive wire test for every previously omitted accepted effort: `none`, `minimal`, `max`, and `ultra`. For each case, have the fake Grok snapshot advertise the effort, select the synthesized compound option, and assert:

- `modelId` is the literal base ID (for example, `grok-4.6`);
- `_meta.reasoningEffort` equals the advertised effort; and
- the compound raw (for example, `grok-4.6-max`) is never sent as `modelId`.

The current fixture uses `"none"` as the sentinel for missing session configuration, so that sentinel will need a distinct value to cover the real `.none` effort. No further production redesign is requested.

Rechecked latest head: `03d998d5a9153d8dad4a82821e9a42344b78b37b`.

datawisebets:
@baron Done in 952db33e — `testEveryAcceptedEffortSendsBaseModelIdAndEffortMeta` is table-driven over `none`/`minimal`/`max`/`ultra`: the fake snapshot advertises each via a new fixture knob, the synthesized compound is selected, and the test asserts modelId is the literal base id, `_meta.reasoningEffort` equals the advertised effort, and the compound never goes out as modelId. The missing-session-config sentinel moved from `none` to `absent` so the real `none` effort is covered. Suite: 100 Grok tests green.

datawisebets:
Final review-loop disposition (9 rounds, GPT-5.6 Pro via codex-chatgpt-web). Every actionable finding through round 8 is fixed with regression coverage (100 Grok tests green, full matrix green on the last push, new CI running on 952db33e).

The one round-9 finding is rejected as out of contract: it posits an effort mutation desynchronizing the modern `sessionModeSnapshot`. For Grok that state is unreachable — (1) grok 1.0.3/1.0.4 never advertise `configOptions`, and `sessionModeSnapshot` is only ever populated from a modern configOptions parse; (2) grok runs carry `acpSessionModeID = nil`, so `setSessionMode` is never invoked for this provider; (3) a session advertising a modern model selector disables the direct path entirely (`sessionModelConfigOptionID != nil` gates it), so the two representations cannot both be authoritative for one session. If a future Grok version advertises a modern mode selector alongside SessionModelState, this would deserve a revisit.

Net state: fail-closed model + effort authority (exclusive-Ok confirms, exclusive-Err preserves, everything else invalidates), live-only effort mutations, structured effort provenance end-to-end, and positive wire coverage for every accepted effort.

datawisebets:
@baron Follow-up pushed in `cfb07f69`.

This hardens the remaining effort-metadata edges:

- preserves the exact provider-advertised wire value (for example, `maximum` stays `maximum` rather than being rewritten to `max`);
- fails closed on semantic alias conflicts and on unknown/incomplete defaults or mode matches;
- keeps value-only entries in current-effort ambiguity analysis; and
- adds focused coverage for exact aliases, conflicting aliases, unknown defaults, missing IDs, a genuine future effort token, and missing session mappings.

CI is green across all four app shards, provider tests, style/guardrails, secret scanning, and the Sentry/privacy lane.

baron (approved):
Approved after re-reviewing the complete remediation on exact head `cfb07f696c8a9d9c9e787cc59a867cb0e7cf2778`.

The original reasoning-effort suffix defect is resolved through structured variant provenance; literal base-ID collisions are protected; current/default effort authority, warmed-state handling, and ambiguous acknowledgements fail closed; the positive wire matrix covers `none`, `minimal`, `max`, and `ultra`; and the latest hardening preserves exact provider-advertised wire aliases while rejecting ambiguous mappings. All hosted checks are green.

PR #796, opened 2026-08-11
Add configurable multi-Oracle groups

dsebban:
This adds configurable multi-Oracle groups across the macOS app, Context Builder, and direct headless MCP. A permanent Primary Oracle can be joined by up to four ordered additional Oracles; every lane receives the same frozen context and returns an independent, unsynthesized response.

## Summary

- Adds an ordered Oracle roster: Primary plus zero to four additional Oracles, including intentional duplicate model choices.
- Runs Oracle lanes independently against the same request snapshot and preserves deterministic lane order in results.
- Keeps successful lane output when another lane fails or is cancelled, with group status derived from the complete lane set.
- Exposes the roster in Agent Models settings and the debug MCP `app_settings` surface.
- Integrates group execution with Context Builder, Oracle history, tool cards, and direct headless MCP.
- Preserves the existing single-Oracle path when no additional Oracle is configured.

## Implementation Notes

- Shared domain contracts own roster validation, lane identity, ordered results, coordinator behavior, claims, and persistence.
- App and direct-headless adapters use those contracts rather than implementing separate group semantics.
- Context Builder commits one immutable roster per run and maps every lane back to the originating tool card/history entry.
- Responses remain independent by design; this PR does not add synthesis, voting, or cross-lane prompting.
- Compatibility behavior is limited to the existing one-Oracle contract and persisted Oracle history migration.

## Scope

This PR is intentionally limited to multiple-Oracle configuration, execution, persistence, presentation, and the tests required for those boundaries. It does not redesign unrelated Agent Mode workflows, provider selection, general chat, file selection, or MCP tool behavior.

## Review Approach

- Review the complete branch against `main`, not only the last commit.
- Treat unrelated refactors, speculative abstractions, compatibility layers without an observed contract, and changes outside the multi-Oracle boundary as must-fix scope issues.
- Verify the one-Oracle fallback, ordered duplicate lanes, partial failure/cancellation, durable history, Context Builder mapping, and direct-headless parity.

## Validation

- Conductor built, ad-hoc signed, and validated the exact `af733abb` debug app bundle; its embedded MCP helper smoke passed. CUA Driver then launched that exact executable background-only and preserved the user's foreground app.
- A fresh live Context Builder call at `af733abb` ran four concurrent Oracle lanes. Primary Oracle, Oracle 2, Oracle 3, and Oracle 4 each returned `PR796_AF733ABB_FOUR_ORACLE_OK`; group `46F0B699-9604-457F-AFAC-2343571824E0` completed.
[...]

PR #832, opened 2026-08-17
test: stabilize worktree API smoke and agent-run startup

dsebban:
Split the two CI test-stability commits out of PR #796 (Oracle groups) so the maturing Oracle feature ships only its scoped changes.

- test(agent-run): observe durable startup state (AgentRunWorktreeStartTests)
- test(worktree): avoid seeded startup in API smoke (WorktreeAPISmokeHarnessTests)

Both address pre-existing host-runner flakes, neither is Oracle-scope. Cherry-picked onto upstream/main. See docs/investigations/oracle-group-branch-scope-audit.md.


baron (approved):
Reviewed against current main. The test-only changes correctly use durable run lifecycle state rather than relying solely on the transient agentTask handle, and the DEBUG startup override is scoped and restored. No blocking issues found.

PR #816, opened 2026-08-13
Support exact secondary-root Agent worktrees

nickdavis:
## Summary

- authorize Agent worktree execution against the complete canonical saved-workspace root set, including exact worktrees for secondary repository roots
- keep saved-workspace identity separate from session worktree projection while failing closed for stale, unmatched, duplicate, or ambiguous root bindings
- preserve explicit secondary-repository selection through Agent start/explore paths and add direct create-session, bind, and builtin-review steer coverage

## Validation

- `make dev-test FILTER=AgentRunWorktreeStartTests` — 55 tests passed
- `make dev-lint`
- `make dev-swift-build PRODUCT=RepoPrompt`
- `.agents/skills/rpce-contribution-check/scripts/preflight.sh commit`
- `.agents/skills/rpce-contribution-check/scripts/preflight.sh push`
- independent read-only review of `origin/main..HEAD` — no blockers

## Additional validation context

The full local PR-ready root suite was attempted. It progressed through the affected coverage but did not complete: an unrelated durable-artifact concurrency test reported a busy publication result, and the run later stopped producing output in workspace-context tests. The focused Agent worktree suite, strict lint, and affected product build all passed.

PR #835, opened 2026-08-17
Reject rootless working_dirs tab bindings

nickdavis:
## Summary

- validate a `working_dirs` bind against the prospective active tab's authoritative file-tool root projection before changing the connection binding
- fail closed with workspace, tab, context, window, and missing-root guidance when that tab is rootless
- preserve the previous explicit binding; do not switch or create tabs, retry, or use tab-order heuristics

## Reproduction and root cause

In RepoPrompt CE 1.3.0, `working_dirs` could select the correct saved workspace/window while retaining its active explicit Agent tab. That tab's authoritative file-tool lookup context could be fail-closed and empty, so `get_file_tree(type="roots")` returned `Roots: 0` / `No workspace loaded` while the diagnostic footer still listed the workspace's loaded root.

The binding path resolved the workspace/window and captured `activeComposeTabID`, but did not validate the prospective tab's actual `WorkspaceLookupContext`. Workspace-visible roots are diagnostic and can diverge from the bound tab's file-tool projection. Tab index/order is not involved.

## Validation

- `WorkingDirsBindRootProjectionTests`: 1 passed
- `BindContextRoutingRecoveryTests`: 6 passed
- strict SwiftFormat/SwiftLint: passed
- commit and push safety preflights: passed
- independent read-only review: no blockers
- full PR-ready root suite: timed out after 3600 seconds with no assertion failure while `GitWorktreeCreationReceiptTests.testConcurrentSameRepositoryCreationsKeepReceiptsSessionIsolated` was still running; the change-specific and neighboring suites above passed independently

## Relationship to open PRs

This PR is standalone on current `main`. #812 restricts `context_id` candidate windows but does not validate `working_dirs` active-tab roots. #816 handles downstream secondary-root Agent worktrees and does not touch `bind_context`. Neither PR is required.

### DIFF
--- a/Sources/RepoPrompt/Infrastructure/MCP/WindowRoutingService.swift
+++ b/Sources/RepoPrompt/Infrastructure/MCP/WindowRoutingService.swift
@@ -1805,19 +1805,43 @@ final class WindowRoutingService: Service {
     private func bindTarget(
         _ target: ResolvedBindTarget,
         connectionID: UUID,
-        clientName: String?
+        clientName: String?,
+        expectedWorkingDirsSourceIdentity: AgentWorkspaceLookupContextIdentity? = nil,
+        bindingAlreadyMatches: Bool = false
     ) async throws {
         guard let targetWindow = windowStates.allWindows.first(where: { $0.windowID == target.windowID }) else {
             throw MCPError.invalidParams("Window \(target.windowID) not found")
         }
-        clearNonRunScopedBindingsAcrossWindows(for: connectionID)
-        try targetWindow.mcpServer.bindTabForConnection(
-            connectionID: connectionID,
-            clientName: clientName,
-            tabID: target.tabID,
-            workspaceID: target.workspaceID,
-            windowID: target.windowID
-        )
+        if let expectedWorkingDirsSourceIdentity {
+            let currentTarget = try? resolveActiveTabBindTarget(
+                windowID: target.windowID,
+                expectedWorkspaceID: target.workspaceID,
+                matchedBy: target.matchedBy,
+                normalizedWorkingDirs: target.normalizedWorkingDirs
+            )
+            guard currentTarget?.tabID == target.tabID,
+                  targetWindow.mcpServer.prospectiveFileToolLookupSourceIsCurrent(
+                      tabID: target.tabID,
+                      workspaceID: target.workspaceID,
+                      expectedSourceIdentity: expectedWorkingDirsSourceIdentity
+                  )
+            else {
+                throw MCPError.invalidRequest(
+                    "The working_dirs target changed while its root projection was being resolved. " +
+                        "The existing MCP binding was not changed. Bind again with the same working_dirs."
+                )
+            }
+        }
+        if !bindingAlreadyMatches {
+            clearNonRunScopedBindingsAcrossWindows(for: connectionID)
+            try targetWindow.mcpServer.bindTabForConnection(
+                connectionID: connectionID,
+                clientName: clientName,
+                tabID: target.tabID,
+                workspaceID: target.workspaceID,
+                windowID: target.windowID
+            )
+        }
         try await networkMgr.setActiveWindowForCurrentConnection(target.windowID)
     }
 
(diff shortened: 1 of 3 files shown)


Sources/RepoPrompt/Infrastructure/MCP/WindowRoutingService.swift
```diff
@@ -1850,6 +1850,32 @@ final class WindowRoutingService: Service {
         )
     }
 
+    private func ensureWorkingDirsRootProjectionIsLoaded(
+        _ target: ResolvedBindTarget,
+        requestedRoots: [String]
+    ) async throws {
+        let window = try resolveWindowForBinding(windowID: target.windowID)
+        let lookupContext = try await window.mcpServer.resolveFileToolLookupContext(
```
dsebban: **P1 — This can reject a valid Agent tab before its bindings hydrate.** `resolveFileToolLookupContext(tabID:workspaceID:)` snapshots the current provider state and immediately calls `lookupContext(for:)`; unlike the real request-based file-tool resolver, it never invokes `agentWorktreeBindingStateResolver` when the state is `.unhydrated`. That recoverable state therefore becomes an empty fail-closed projection and `working_dirs` is rejected even though the next file-tool request would hydrate and succeed. Please resolve the prospective tab through the same hydration and stale-route checks, without changing the connection binding, and add a test where `.unhydrated` resolves to valid hydrated bindings and the bind succeeds.

Sources/RepoPrompt/Infrastructure/MCP/WindowRoutingService.swift
```diff
@@ -2028,6 +2054,10 @@ final class WindowRoutingService: Service {
                                 normalizedWorkingDirs: target.normalizedWorkingDirs
                             )
                         }
+                        try await ensureWorkingDirsRootProjectionIsLoaded(
```
dsebban: **P1 — Revalidate after this await before clearing the old binding.** This preflight yields on `@MainActor`; while it is resolving/materializing roots, the selected tab can close or its Agent session/projection can change. `bindTarget` then clears every non-run-scoped binding before `bindTabForConnection` validates the captured tab, so a stale target can both fail the bind and erase `previousBinding`—the opposite of the guarantee in the new error path. Please make the final target/projection validation and clear+bind mutation one no-await transaction, or at minimum validate target existence/current session immediately before clearing, and cover the race with a barrier-controlled test.

dsebban (changes requested):
Summary: This change adds a fail-closed preflight for `working_dirs` so a connection is not rebound to an active tab whose authoritative file-tool projection lacks the requested roots, with a focused regression test for an unavailable/rootless Agent session. The core direction is sound and the scope is appropriately limited to the routing implementation and its test, but the preflight currently bypasses the established hydration/staleness path and introduces an async gap before the old binding is cleared, so I’m requesting changes.

### P1

- `Sources/RepoPrompt/Infrastructure/MCP/WindowRoutingService.swift:1858` — The shortcut resolver treats recoverable `.unhydrated` Agent bindings as an empty fail-closed projection. Use the same hydration and stale-route checks as real file-tool resolution and add a success regression test.
- `Sources/RepoPrompt/Infrastructure/MCP/WindowRoutingService.swift:2057` — The new await can leave `tabTarget` stale; `bindTarget` clears the previous binding before validating the tab. Revalidate and make validation plus clear/bind atomic, with a race regression test.

Sources/RepoPrompt/Infrastructure/MCP/WindowRoutingService.swift
dsebban: **P1. This guards the symptom rather than fixing the reported defect.** The affected workspace remains unusable through `working_dirs`; the behavior changes from a successful bind with `Roots: 0` to an explicit rejection telling the user to change tabs. Before treating this as the fix, trace why the active Agent tab's authoritative projection is unavailable in the reproduced workspace. If it is stale or recoverable persisted Agent binding state, repair or invalidate that state at its ownership boundary. Keep this preflight as a safety net, but add an end-to-end regression where the original affected workspace binds successfully and the subsequent file-tool roots contain the requested root. A rejection-only test proves safer failure, not restored usability.

dsebban (dismissed):
Re-review using Cursor pstack Poteto Mode. The earlier two P1 findings still stand. A root-cause pass adds one broader concern. This PR makes the failure safer, but it does not yet fix the reported workspace-usability defect. The regression proves rejection, not recovery. Please keep the fail-closed guard as defense, but trace and repair the state that makes the active Agent tab's authoritative projection unavailable, then prove the original `working_dirs -> file tool roots` path succeeds end to end.

dsebban (commented):
Fresh Poteto Mode rerun against unchanged head `b1f1f9c`.

The diff remains focused. It changes the routing implementation and adds one targeted regression test. I found no unrelated files or scope creep.

I am withdrawing my broader request for automatic Agent-state repair. Fail-closed rejection with recovery guidance is a valid scoped contract for this PR. I resolved that thread.

Two P1 blockers remain. The prospective lookup bypasses recoverable `.unhydrated` binding hydration. The awaited preflight also permits a stale target to erase the previous binding during clear-before-validation.

No additional findings.

dsebban:
Poteto rerun finding on unchanged head `b1f1f9c`.

Two P1 blockers remain.

1. The prospective `working_dirs` preflight uses a lookup path that does not hydrate recoverable `.unhydrated` Agent bindings. It can reject a tab that the normal file-tool path would hydrate successfully. See the inline finding at `WindowRoutingService.swift:1858`.

2. The preflight awaits before rebinding. The captured tab or Agent projection can become stale, after which `bindTarget` clears the previous binding before validating the destination. A failed stale bind can therefore erase the valid prior binding. See the inline finding at `WindowRoutingService.swift:2057`.

The diff remains focused to the routing implementation and its targeted test. No unrelated files or additional scope findings.

nickdavis:
Addressed both remaining P1 findings in `917095c5`:

1. Prospective `working_dirs` resolution now hydrates recoverable `.unhydrated` Agent bindings through the established hydration/staleness path, with a success regression test.
2. The destination tab and Agent projection are revalidated immediately before a no-await clear/bind transaction, with a barrier-controlled race test confirming the previous binding is preserved.

The focused routing tests, shared context/freshness suites, strict lint, and the RepoPrompt product build pass. Ready for re-review.

baron (approved):
Reviewed the exact current head after the working_dirs binding authority fixes. The replacement is generation-fenced, revalidates the active target at commit time, preserves the prior usable binding on validation failure, and compares projected roots case-insensitively. Focused regressions, lint/build validation, and exact-head hosted CI are green.

PR #846, opened 2026-08-19
Add session ID copying to Agent Mode sidebar menus

### LINKED ISSUE #842: Add Copy session ID to Agent Mode sidebar context menus
## Problem statement

When someone wants an agent or the history tool to inspect a specific session, they currently have to reference its title and hope it resolves to the right chat. Sometimes they must rename the chat first. The sidebar already knows the persistent session ID, but gives the user no way to copy it.

## Solution

Add `Copy session ID` to the context menu for active, archived, and sub-agent chat rows in the Agent Mode sidebar.

The command copies the raw persistent session UUID. No prefix, title, or other text.

Keep the command visible when the row does not have a persistent session ID yet, but disable it. This makes the feature easy to find and explains why it is temporarily unavailable.

Limit this change to sidebar context menus.

## User stories

1. As a user, I want to copy a chat's session ID, so that I can point an agent to the exact session.
2. As a user with several similarly named chats, I want to reference sessions by ID, so that title ambiguity cannot select the wrong one.
3. As a user, I want the raw UUID on my clipboard, so that I can reuse it in prompts, commands, and tool arguments without removing a prefix.
4. As a user reviewing old work, I want the command on archived chats, so that older sessions remain easy to reference.
5. As a user reviewing delegated work, I want the command on sub-agent chats, so that I can reference the child session instead of its parent.
6. As a user on a new compose tab, I want to see the disabled command, so that I know where it will become available once the session exists.
7. As a user, I want the command in the same part of each context menu, so that I can find it without scanning the whole menu.
8. As a user copying several IDs, I do not want a toast after each copy, so that the action stays quiet.
9. As a user who renames chats, I want copied references to remain valid, so that chat organization does not affect session identity.
10. As a user of the existing Handoff command, I want this action to copy only the ID, so that the two commands keep separate jobs.

## Implementation decisions

- Use the persistent Agent Mode session UUID as the authority.
- Do not copy the compose-tab ID, workspace ID, provider ID, or chat title.
- Copy the UUID string with no prefix or surrounding text.
- Pass the optional persistent session ID to the sidebar row's copy action.
- Show `Copy session ID` on every active and sub-agent row. Place it immediately after `Rename chat`.
- Show `Copy session ID` on every archived row. Place it after `Restore tab` and before the divider above the delete command.
- Disable the command when the row cannot resolve a persistent session ID.
- Use the same copy behavior for active, archived, and sub-agent rows.
- Write to the clipboard once when the user selects an enabled command.
- Do not show a toast or copied-state indicator.
- Leave the session-title menu and its Handoff command unchanged.
- Do not change session persistence, history lookup, MCP contracts, or stored data.

The active menu order should be:

1. Select chat
2. Divider
3. Pin or Unpin chat
4. Rename chat
5. Copy session ID
6. Stash chat for later, when available
7. Dismiss status badge, when available
8. Divider
9. Delete chat

The archived menu order should be:

1. Select chat
2. Divider
3. Restore tab
4. Copy session ID
5. Divider
6. Delete stashed tab

## Testing decisions

Keep one small testable boundary for the copy action. It should accept an optional session UUID and an injected clipboard writer.

Automated tests should prove:

- A persistent session ID enables copying.
- Copying writes the exact raw UUID once.
- The copied value contains no prefix or title.
- A missing session ID disables copying and causes no clipboard write.
- Active, archived, and sub-agent rows use the same copy action.

Use the existing sidebar-session builder tests as the authority for persisted and sessionless rows. Do not repeat those persistence tests here.

The existing titlebar menu tests provide useful prior art for injected clipboard writes and exact write-count assertions. This feature should not modify or extend the titlebar menu.

Do not add broad SwiftUI snapshots, tests that inspect source code, or tests that write to the real clipboard.

Run the focused root test for the copy action, then the relevant root test suite. Run the coordinated formatter and lint checks for the Swift changes.

After receiving approval to launch the debug app, check these menus by hand:

- an active persisted chat;
- a new compose tab with no session ID;
- an archived chat;
- a persisted sub-agent chat.

The manual check should confirm presence, order, enabled state, and the exact clipboard value.

## Acceptance criteria

- Active chat context menus show `Copy session ID` after `Rename chat`.
- Archived chat context menus show it after `Restore tab` and before the delete divider.
- Persisted sub-agent chats show the same command.
- The command copies the persistent Agent Mode session UUID and nothing else.
- The command remains visible but disabled when no persistent session ID exists.
- Copying does not show a toast.
- The session-title menu does not gain this command.
- No hover button or keyboard shortcut is added.
- Existing selection, pin, rename, stash, restore, Handoff, and delete behavior still works.

## Out of scope

- The session-title menu
- Hover buttons
- Keyboard shortcuts
- Labels such as `session_id:`
- Chat titles in the clipboard value
- Handoff or supervisor prompts
- Copy confirmation UI
- History tool or MCP changes
- Changes to session creation or persistence

## Further notes

Issue #531 is related but not a duplicate. It proposed a full CLI supervisor prompt in the session-title menu. This issue adds a smaller sidebar command that copies only the session UUID.

### Maintainer-guidance check

| Check | Finding |
| --- | --- |
| User impact and invariant | Every sidebar chat row exposes the same command. When enabled, it copies only the persistent session UUID. |
| Root-cause confidence | Confirmed. The sidebar has access to the ID but does not expose a copy action. |
| Authority | The persistent Agent Mode session ID. |
| State risk | Low. The command changes only the clipboard. |
| Scale and observability | The action does constant work and starts no background task. It needs no progress UI. |
| Recommended scope | One Agent Mode sidebar change. |
| Validation | Focused copy-action tests plus a short check of the assembled menus. |

aAtila:
## Summary

Agent Mode sidebar context menus can now copy the raw persistent session UUID for active, archived, and sub-agent chats. The command stays visible but disabled when a row has no persistent session ID.

## Changes

- Add one shared clipboard action that owns enablement and writes the exact UUID once.
- Place `Copy session ID` after rename for active and sub-agent chats, and after restore for archived chats.
- Cover raw clipboard output, missing IDs, and reuse across all three row types.

## Testing

- `./conductor test --filter AgentSidebarSessionIDCopyActionTests`
- `./conductor lint`
- `./conductor guardrails`
- `./conductor swift-build --product RepoPrompt`

The full `./conductor test` run did not complete cleanly. It reported failures outside this feature in Codex managed-auth and configuration tests and Context Builder tests, then stopped producing output in `WorkspaceSwitchRecoveryTests`. The job was canceled after 18 minutes so the focused test and product build could run.

Closes #842

## Screenshot

image

### DIFF
--- a/Sources/RepoPrompt/Features/AgentMode/ViewModels/AgentModeViewModel+SidebarSessions.swift
+++ b/Sources/RepoPrompt/Features/AgentMode/ViewModels/AgentModeViewModel+SidebarSessions.swift
@@ -25,6 +25,7 @@ extension AgentModeViewModel {
         let filteredTabs: [StashedTab]
         let sortedTabs: [StashedTab]
         let dateInfoByStashedTabID: [UUID: SidebarSessionDateInfo]
+        let sessionIDByStashedTabID: [UUID: UUID]
     }
 
     struct ArchivedHUDSessionDescriptor {
@@ -32,6 +33,11 @@ extension AgentModeViewModel {
         let entry: AgentSessionIndexEntry?
         let dateInfo: SidebarSessionDateInfo
         let searchFields: AgentSessionSearchFields
+    }
+
+    private struct ArchivedSidebarSessionMetadata {
+        let dateInfoByStashedTabID: [UUID: SidebarSessionDateInfo]
+        let sessionIDByStashedTabID: [UUID: UUID]
     }
 
     private struct ArchivedSidebarSessionLookup {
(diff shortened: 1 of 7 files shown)


baron (approved):
Reviewed the exact current integration at PR head 923ada10133749c80b10f6281b541a5d10a6d127 and synthetic merge 21ef8bfff647d2b920484d2c8ed1b2a129bd7fb6 against base 86cbb7e1611d6ae5cbb4fffb1c0bf535c32b2ac1. No actionable findings. The persistent-session identity projection, menu ordering and enablement, archived fallback, and focused test coverage are coherent. This was a static source review; I did not run local tests or exercise the live clipboard/UI. All exact-head hosted checks are green.

PR #844, opened 2026-08-19
Keep Grok Oracle assistant text complete through ACP drain

dsebban:
## Summary
Grok CLI Oracle answers were getting truncated or mixed with status/system events because ACP shutdown finished the stream before remaining assistant chunks were drained.

This change:
- Filters status/system events out of assistant content (no length heuristic)
- Drains ACP stdout/stderr after the prompt completes and emits `message_stop` last
- Converts unsettled shutdowns to an incomplete terminal while keeping partial content

## Test plan
- [x] `ProviderStreamCompletionTests` semantic-text filtering
- [x] `GrokBuildACPHeadlessAgentProviderTests` late-output and unsettled-transport cases
- [ ] Live Grok Oracle call after a visible CE debug relaunch (not run; prior approval timed out)

PR #847, opened 2026-08-20
Drain leftover Grok ACP stdout after session/prompt, Grok-only

dsebban:
## Why

Grok CLI can still write `session/update` chunks after `session/prompt` returns `end_turn`. The old one-shot path shut down immediately and dropped those bytes, so short Oracle answers could be truncated. OpenCode and Cursor use the same bridge but do not opt into this drain path.

## What changed

- Grok remains the only provider that opts into transport draining.
- Cooperative settlement now waits for the root process and both existing stdout and stderr consumer tasks under one bounded deadline. This proves that pipe EOF was observed and queued chunks reached the framers.
- Successful settlement flushes partial lines after the consumers finish, then publishes the terminal event.
- Timeout or cancellation removes handlers, closes the channels, terminates the saved process group, and publishes `message_incomplete` with `transport_did_not_settle_before_shutdown`.
- OpenCode and Cursor retain the existing prompt-then-shutdown path.

```mermaid
sequenceDiagram
    participant Oracle
    participant Bridge
    participant ACP
    participant CLI as Grok stdio

    Oracle->>Bridge: execute prompt
    Bridge->>ACP: session/prompt
    CLI-->>ACP: end_turn
    CLI-->>ACP: late session/update chunks
    ACP->>ACP: await root exit and both pipe consumers
    alt transport settles before deadline
        ACP->>ACP: flush framers
        ACP-->>Bridge: content then message_stop
        Bridge-->>Oracle: complete response
    else deadline or cancellation
        ACP->>ACP: cancel consumers and terminate process group
        ACP-->>Bridge: message_incomplete
        Bridge-->>Oracle: bounded incomplete response
    end
```

The regression test forks a descendant that retains inherited stdout and stderr after the root exits. It verifies that late content remains ordered, the response does not hang, and exactly one incomplete terminal is emitted when transport settlement misses the deadline.

This supersedes the extra work from #844.

## Verification

- `make dev-test FILTER=GrokBuildACPHeadlessAgentProviderTests` passed 9 tests.
- `make dev-test FILTER=ProviderStreamCompletionTests` passed 6 tests.
- `make dev-lint` passed.
- `make dev-swift-build PRODUCT=RepoPrompt` passed.
- `make dev-format` reported 0 of 1517 files formatted.

## Files

- `Sources/RepoPrompt/Infrastructure/AI/ACP/ACPAgentSessionController.swift` changed.
- `Sources/RepoPrompt/Infrastructure/AI/AIQueriesService.swift` changed.
- `Sources/RepoPrompt/Infrastructure/AI/Providers/ACPHeadlessAgentProviderBridge.swift` changed.
- `Sources/RepoPrompt/Infrastructure/AI/Providers/GrokBuild/GrokBuildACPHeadlessAgentProvider.swift` changed.
- `Tests/RepoPromptTests/AI/ProviderStreamCompletionTests.swift` changed.
- `Tests/RepoPromptTests/AgentMode/GrokBuildACPHeadlessAgentProviderTests.swift` changed.

### DIFF
--- a/Sources/RepoPrompt/Infrastructure/AI/ACP/ACPAgentSessionController.swift
+++ b/Sources/RepoPrompt/Infrastructure/AI/ACP/ACPAgentSessionController.swift
@@ -239,6 +239,7 @@ actor ACPAgentSessionController {
     private var stdoutConsumerTask: Task<Void, Never>?
     private var stderrConsumerTask: Task<Void, Never>?
     private var processWaitTask: Task<Void, Never>?
+    private var isDrainingTransport = false
     private var stdoutFramer = LineFramer()
     private var stderrFramer = LineFramer()
     private var launchDescription: String?
@@ -1172,31 +1173,80 @@ actor ACPAgentSessionController {
         }
     }
 
+    func prepareForPromptTransportDrain() {
+        isDrainingTransport = true
+    }
+
     func shutdown() async {
+        _ = await shutdown(drainTransport: false)
+    }
+
+    /// Shuts down a completed one-shot prompt without discarding bytes that become
+    /// readable while the ACP process responds to stdin closure.
+    ///
+    /// Returns `true` only when the process and its transport settle cooperatively
+    /// before the bounded timeout. Forced termination discards bytes still buffered
+    /// in the pipes and cannot prove that the provider finished writing the response.
+    func shutdownAfterPromptTransportDrain() async -> Bool {
+        await shutdown(drainTransport: true)
+    }
+
+    private func shutdown(drainTransport: Bool) async -> Bool {
         #if DEBUG
             debugResumeConfigurationMutationPostcheck()
         #endif
-        guard state != .closed, state != .closing else { return }
+        guard state != .closed, state != .closing else { return false }
         state = .closing
+        isDrainingTransport = drainTransport
         log("Shutting down ACP controller")
 
         await cancelPrompt()
         failAllPromptSettlementWaiters(with: ControllerError.transportClosed)
         failPendingRequests(with: ControllerError.transportClosed)
 
-        stdoutChannel?.finish()
-        stderrChannel?.finish()
-        stdoutConsumerTask?.cancel()
-        stderrConsumerTask?.cancel()
-        processWaitTask?.cancel()
-
-        if let process {
-            process.stdout.readabilityHandler = nil
-            process.stderr.readabilityHandler = nil
+        var transportSettled = !drainTransport
+        if drainTransport, let process {
             process.stdin?.closeFile()
-            _ = await ProcessTermination.terminateAndReap(pid: process.pid, processGroupID: process.processGroupID)
+            transportSettled = await waitForTransportSettlementDuringDrain()
+            if transportSettled {
+                stdoutFramer.flush { lineData in
+                    handleJSONLine(lineData)
+                }
+                stderrFramer.flush { lineData in
+                    handleStderrLine(lineData)
+                }
+            } else {
+                process.stdout.readabilityHandler = nil
+                process.stderr.readabilityHandler = nil
+                stdoutChannel?.finish()
+                stderrChannel?.finish()
+                stdoutConsumerTask?.cancel()
+                stderrConsumerTask?.cancel()
+                processWaitTask?.cancel()
+                _ = await ProcessTermination.terminateAndReap(
+                    pid: process.pid,
+                    processGroupID: process.processGroupID
+                )
+            }
+        } else {
+            stdoutChannel?.finish()
+            stderrChannel?.finish()
+            stdoutConsumerTask?.cancel()
+            stderrConsumerTask?.cancel()
+            processWaitTask?.cancel()
+
+            if let process {
+                process.stdout.readabilityHandler = nil
+                process.stderr.readabilityHandler = nil
+                process.stdin?.closeFile()
+                _ = await ProcessTermination.terminateAndReap(
+                    pid: process.pid,
+                    processGroupID: process.processGroupID
+                )
+            }
         }
 
+        isDrainingTransport = false
         await clearExpectedAgentPIDIfNeeded()
         await cleanupLaunchArtifacts()
 
@@ -1219,6 +1269,39 @@ actor ACPAgentSessionController {
         suppressSessionLoadReplayUpdates = false
         state = .closed
         finishEventsIfNeeded()
+        return transportSettled
+    }
+
+    private func waitForTransportSettlementDuringDrain() async -> Bool {
+        guard let processWaitTask,
+              let stdoutConsumerTask,
+              let stderrConsumerTask
+        else { return false }
+        let timeout = ProcessTermination.cooperativeCancellationWaitTimeout()
+        return await withTaskGroup(of: Bool.self) { group in
+            group.addTask {
+                await processWaitTask.value
+                await stdoutConsumerTask.value
+                await stderrConsumerTask.value
+                return true
+            }
+            group.addTask {
+                do {
+                    try await Task.sleep(nanoseconds: UInt64(timeout * 1_000_000_000))
+                } catch {
+                    return false
+                }
+                return false
+            }
+            let settled = await group.next() ?? false
+            if !settled {
+                processWaitTask.cancel()
+                stdoutConsumerTask.cancel()
+                stderrConsumerTask.cancel()
+            }
+            group.cancelAll()
+            return settled
+        }
     }
 
     // MARK: - Transport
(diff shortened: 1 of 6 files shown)


Sources/RepoPrompt/Infrastructure/AI/AIQueriesService.swift
```diff
@@ -330,6 +330,16 @@ public class AIQueriesService {
         return trimmed.hasPrefix("**") || trimmed.contains("****")
     }
 
+    static func assistantText(for result: AIStreamResult) -> String? {
```
dsebban, line 333: **P0 — revert this allowlist from this PR.** This drops valid Anthropic streaming text: `AnthropicProvider` builds `AIStreamResult(type: result.type, text: result.contentBlock?.text ?? result.delta?.text)`, and the pinned SwiftAnthropic SDK reports those text-bearing events as `content_block_start` / `content_block_delta`, not `content` / `final_content`. Exact change: delete this helper, restore the previous nonempty-`result.text` check, and move semantic isolation to a separately audited provider-normalized contract.

Sources/RepoPrompt/Infrastructure/AI/ACP/ACPAgentSessionController.swift
```diff
@@ -1172,31 +1173,80 @@ actor ACPAgentSessionController {
...
+                    handleStderrLine(lineData)
+                }
+            } else {
+                process.stdout.readabilityHandler = nil
+                process.stderr.readabilityHandler = nil
+                stdoutChannel?.finish()
+                stderrChannel?.finish()
+                stdoutConsumerTask?.cancel()
+                stderrConsumerTask?.cancel()
+                processWaitTask?.cancel()
```
dsebban, line 1225: **P1 — this creates two termination/reap owners.** `waitForTransportSettlementDuringDrain()` already cancels `processWaitTask`; its `waitForTermination` cancellation path performs TERM→KILL and reaps. The following direct `terminateAndReap` repeats that work after the task-group wait. Exact change: remove the second direct termination call; cancel and await the existing process wait task as the sole owner, while independently finishing/canceling the pipe consumers.

Sources/RepoPrompt/Infrastructure/AI/Providers/ACPHeadlessAgentProviderBridge.swift
```diff
@@ -122,24 +130,37 @@ final class ACPHeadlessAgentProviderBridge: HeadlessAgentProvider {
 
             await controller.setExpectedMCPRunID(runID)
             let events = await controller.currentEventsStream()
+            let transportSettlement = ACPHeadlessTransportSettlement()
```
dsebban, line 133: **P1 — terminal ownership is at the wrong layer.** The description says the controller flushes and then publishes terminal, but the controller actually publishes terminal before drain and this bridge actor hides/reconstructs it. Exact change: have the controller retain the prompt terminal while drain mode is armed and emit it after flush (or as incomplete on timeout), then delete this settlement actor plus `terminalResult` buffering/synthesis. Keep only the completion policy and direct forwarding here.

Tests/RepoPromptTests/AgentMode/GrokBuildACPHeadlessAgentProviderTests.swift
```diff
@@ -53,13 +53,88 @@ final class GrokBuildACPHeadlessAgentProviderTests: XCTestCase {
...
+
+        let elapsed = ProcessInfo.processInfo.systemUptime - startedAt
+        XCTAssertEqual(results.filter { $0.type == "content" }.compactMap(\.text).joined(), "Full response.")
+        let terminals = results.filter { $0.type == "message_stop" || $0.type == AIStreamResult.incompleteType }
+        XCTAssertEqual(terminals.count, 1)
+        let terminal = try XCTUnwrap(terminals.first)
+        XCTAssertEqual(terminal.type, AIStreamResult.incompleteType)
+        XCTAssertEqual(terminal.stopReason, "transport_did_not_settle_before_shutdown")
+        XCTAssertFalse(results.contains { $0.type == "message_stop" })
+        XCTAssertLessThan(elapsed, 15)
```
dsebban, line 105: This bound is too loose to prove the configured settlement deadline; it only proves the test did not wait for the fake child's full 30-second sleep. Inject a short drain timeout into the harness and assert against that timeout plus the explicit TERM/KILL cleanup allowance. Also add an unterminated-final-line case so the new framer flush is directly covered.

dsebban (commented):
## Review result: changes required

The underlying truncation bug is real, and most of the added lines are useful fake-ACP regression coverage rather than gratuitous production code. However, the current implementation is not safe or minimal enough to merge.

The invariant this PR needs to own is:

> For Grok drain mode, publish exactly one terminal result, and publish it last—only after the root process and both pipe consumers have settled and both framers have been flushed. If settlement misses the bound, preserve already-delivered content and publish exactly one `message_incomplete` terminal.

The current code splits that invariant between the controller and a second state machine in the bridge. It also introduces a provider-wide transcript regression and gives forced process cleanup two owners.

### Required code changes

1. **Revert the global `AIQueriesService` type allowlist from this PR.**
   - Delete `AIQueriesService.assistantText(for:)`.
   - Restore `if let text = result.text, !text.isEmpty` in `sendPrompt`.
   - Delete/move `testAIQueriesBuffersOnlySemanticAssistantText` from `ProviderStreamCompletionTests`.
   - Delete/move `testStatusRemainsDistinctFromValidShortAssistantContent` unless this PR separately introduces a safe semantic-classification contract.

   This is not merely undocumented scope creep: it breaks existing Anthropic streaming. `AnthropicProvider` carries assistant text in `result.contentBlock?.text ?? result.delta?.text` while forwarding SwiftAnthropic's raw event type. The pinned SDK uses `content_block_start` / `content_block_delta`, so a global `content|final_content` allowlist drops those text chunks. Semantic isolation may be worth doing, but it needs an explicit provider-normalized contract and provider-wide regression coverage—not a raw string allowlist inside this Grok-only transport PR.

2. **Make `ACPAgentSessionController` own terminal ordering.** The PR description says ACP flushes and then publishes the terminal, but the implementation currently emits `message_stop` and `.terminal` inside `prompt()`, then asks the bridge to hide and reconstruct that ordering.
   - Keep the Grok-only completion policy/opt-in.
   - When drain mode is armed, have `prompt()` retain the values produced by `messageStopResult(...)`, `terminalState(...)`, and `terminalErrorText(...)` instead of emitting them immediately.
   - In `shutdownAfterPromptTransportDrain()`, close stdin; await the root wait task plus stdout/stderr consumers; flush both framers; then emit the retained terminal result and `.terminal` event.
   - On timeout/cancellation, emit the retained result as `message_incomplete` with `transport_did_not_settle_before_shutdown`, then emit the corresponding terminal event and finish the stream.
   - Preserve the existing immediate emit/shutdown behavior for OpenCode and Cursor.

3. **Simplify `ACPHeadlessAgentProviderBridge` after the controller publishes terminal-last.**
   - Keep `PromptCompletion` (renaming it to `TransportShutdownPolicy` would describe it more accurately) and the Grok opt-in.
   - Restore direct event forwarding.
   - Delete `ACPHeadlessTransportSettlement`.
   - Delete `terminalResult` buffering, duplicate-terminal arbitration, and bridge-owned terminal synthesis.
   - Move any required successful-to-incomplete conversion beside the controller's pending terminal state. The event producer should enforce ordering; the bridge should not maintain a second terminal lifecycle.

4. **Give process termination/reaping exactly one owner.** `waitForTransportSettlementDuringDrain()` cancels `processWaitTask`; `ProcessTermination.waitForTermination` responds to that cancellation by terminating and reaping the process family. `shutdown` then calls `ProcessTermination.terminateAndReap` again.
   - Preferred patch: on timeout, remove readability handlers, finish channels, cancel the consumer tasks, cancel and await the existing `processWaitTask`, and remove the second direct `terminateAndReap` call.
   - Do not retain both paths. Besides duplicate TERM/KILL grace work, a second PID/PGID cleanup after the first owner has reaped the root creates an avoidable reuse/signaling window.

5. **Tighten the regression tests without deleting the useful transport scenarios.**
   - Keep the pre-terminal fragments, post-response content, inherited-pipe descendant, and forced-incomplete cases.
   - Add a fake response whose final JSON line has no trailing newline, proving the explicit framer flush is necessary and works.
   - Add a cancellation case and verify the same one-terminal, bounded incomplete contract.
   - Assert exactly one terminal and that it is the final result in every drain scenario.
   - Inject a shorter drain timeout for tests. `XCTAssertLessThan(elapsed, 15)` only proves “not the fake child's 30 seconds”; it does not verify the intended settlement bound.
   - Define and test failure precedence. The current bridge preserves an already-incomplete provider terminal when transport also fails to settle, although the PR body says transport timeout publishes `transport_did_not_settle_before_shutdown`.

### File-by-file verdict

- `ACPAgentSessionController.swift`: **core behavior justified**, but terminal ownership and forced-cleanup ownership must be consolidated.
- `AIQueriesService.swift`: **unsafe provider-wide regression; revert/split**.
- `ACPHeadlessAgentProviderBridge.swift`: **policy switch justified; settlement actor and terminal reconstruction overengineered**.
- `GrokBuildACPHeadlessAgentProvider.swift`: **justified; keep the Grok-only opt-in**.
- `ProviderStreamCompletionTests.swift`: **belongs with a separate, correctly designed semantic-classification change**.
- `GrokBuildACPHeadlessAgentProviderTests.swift`: **mostly justified and valuable**; keep the transport cases, remove/split the unrelated semantic test, and add the missing terminal/flush/cancellation assertions above.

CI is also currently red in app-test shards 3 and 4. I could not retrieve the failed test names from the available job logs, so I am not treating that as evidence for the architectural findings above, but the revised head should receive a green rerun.

dsebban:
This drain path does not fix the live Oracle bug. Grok was still a coding agent with tools, so "I'll keep reading..." became the answer.

Closing in favor of https://github.com/repoprompt/repoprompt-ce/pull/856, which runs Oracle Grok as a one-shot CLI instead.

PR #851, opened 2026-08-21
Enable Codex apps with runtime 0.149.0

nickdavis:
## Summary

- rotate the bundled Codex runtime and external compatibility floor from 0.147.0 to 0.149.0 using verified official macOS packages
- stop forcing Codex apps, plugins, MCP elicitation, and tool suggestions off; these are stable and enabled by default in 0.149.0
- reconcile the one bounded app-server schema change in `hooks/list` and refresh runtime, legal, CI, rollback, and candidate fixtures

## Why

Codex 0.149.0 removes the ChatGPT workspace-settings gate from app and plugin availability (openai/codex#38994). RepoPrompt still bundles 0.147.0 and passes explicit `features.apps=false` / `features.plugins=false` overrides to every Codex process, so updating a user's PATH CLI cannot make connected apps callable in Agent Mode.

A live probe ran the official 0.149.0 app server against RepoPrompt's existing isolated `CODEX_HOME`. The existing connected apps were installed, enabled, and callable without sharing the Codex Desktop home or adding a RepoPrompt-specific auth path. A clean 0.149.0 home also reports all four removed feature flags as stable and enabled by default.

## Validation

Passed:

- guarded `codex-update-candidate` verification for official 0.149.0 arm64 and x86_64 packages
- `make codex-acquire` and `make codex-status` for both macOS architectures
- `make dev-codex-schema-check`
- `make dev-test FILTER=CodexIntegrationConfigurationTests`
- `make dev-test FILTER=CodexRuntimeAuthorityTests`
- offline runtime artifact, update-candidate, and schema-gate unit tests
- `make dev-lint`
- `make guardrails`
- `make dev-release-preflight`
- contribution commit/push secret scans

The full `pr-ready` root suite was attempted. It exposed unrelated existing timing failures in codemap, MCP timeout, and Git cancellation tests, then deadlocked in an unrelated `GitWorktreeInitializationAPITests` `git init` child while writing to the test-output pipe. The exact conductor ticket was cancelled after 33 minutes. The changed Codex suites had already passed; hosted CI should supply the full-suite result.

No visible app was launched, stopped, or replaced.


baron:
Enabling these additional features by default without giving the user any control over them would be too disruptive and could break third party integration. Therefore I've taken over the PR and made each additional feature strictly opt in.

In the future please do not make a disruptive change to the core product without community discussion or a GitHub issue first. However, I have no issues giving users the option to test these features even if it degrades the general experience so long as they are aware of the risks.

PR #838, opened 2026-08-18
Add Grok Build CLI to Oracle and built-in chat models

dsebban:
## Summary
- add a thin non-agent `GrokBuildCLIProvider` over the existing Grok Build ACP headless runtime
- add provider-qualified model identity, catalog, factory, and availability routing for Oracle and built-in chat
- keep Oracle UI, profiles, recommendations, and Grok Build process ownership unchanged

## Validation
- `make dev-test FILTER=GrokBuildModelRoutingTests`
- `make dev-test FILTER=GrokBuildACPHeadlessAgentProviderTests`
- `make dev-lint`
- `make dev-swift-build PRODUCT=RepoPrompt`
- contribution commit and push preflights

Live app smoke was not run.

PR #856, opened 2026-08-23
Fix Grok Oracle returning progress text instead of an answer

dsebban:
## What this fixes

Grok Oracle could start a coding-agent turn instead of answering the supplied prompt. On a large review it might read files, spend its turn on tools, and return progress text such as "I'll keep reading..." as the Oracle answer.

This PR gives Grok chat and Oracle a fresh, text-only, one-shot CLI path. Grok Agent Mode stays on ACP and keeps its sessions, tools, and interactive behavior.

## The `/bro` version

Oracle already gives Grok the prompt and selected code. Grok does not need to explore the repository again.

Before this change, Grok could behave like a coding agent: look for files, announce tool work, and run out of turns before answering. Now Oracle starts a separate Grok process with no usable tools, sends the prompt through a private file, accepts one completed JSON answer, and exits.

Agent Mode is unchanged. It still uses `grok agent stdio` because that is where tools and long-lived sessions belong.

## Flow (`/show-me`)

```mermaid
flowchart LR
    A[RepoPrompt request] --> B{Request type}
    B -->|Agent Mode| C[grok agent stdio]
    C --> D[ACP session with tools]
    B -->|Chat or Oracle| E[Private prompt file]
    E --> F[grok one-shot JSON]
    F --> G[Prepared tool count: 0]
    G --> H[One completed text answer]
    H --> I[Delete prompt directory]
```

## Live Oracle proof

Completed Grok 4.6 XHigh Oracle response over 31 selected files

The debug app completed this Grok 4.6 XHigh Oracle review over 31 selected files. It returned a direct review answer with about 100k input tokens and 19k output tokens instead of stopping at tool progress text.

The code change is deliberately narrow:

```diff
 GrokBuildCLIProvider
-  GrokBuildACPHeadlessAgentProvider
+  GrokBuildOneShotHeadlessAgentProvider

 Agent Mode
   GrokBuildACPAgentProvider        # unchanged
```

## Why the tool flags look strange

The final command includes both:

```text
--tools read_file
--disallowed-tools read_file,search_tool,use_tool
```

That is intentional. The installed Grok CLI has no reliable empty-tool flag.

`read_file` is a recognized allowlist anchor. The denylist then removes it. `search_tool` and `use_tool` remove Grok's MCP meta-tools. In the validated Grok 1.0.5 build, this produced:

```text
tool_count: 0
tools: []
```

`--disallowed-tools` removes matching tools from the model schema. It is different from `--deny`, which leaves a tool visible and only blocks execution.

This policy depends on `read_file` remaining a recognized Grok tool. If Grok removes or renames it, its unresolved-allowlist behavior may fail open. A live `tool_count: 0` probe is therefore the compatibility check for future Grok releases.

## What we tried first

The final policy came from live CLI probes, not flag names alone.

### Empty allowlist

```text
--tools ''
```

This did not disable tools. Grok still exposed `read_file`, attempted a read, consumed the single tool cycle, and exited with `Error: max turns reached`.

### Unknown sentinel allowlist

```text
--tools __repoprompt_no_tools__
```

Grok treated the unknown entry as unresolved and retained the normal toolset. This also reached the max-turn failure.

### Large denylist

The first implementation denied 36 known tool names. It still prepared five dynamically injected tools, including `OpenCode:write` and four media tools. A model selected `write`; the permission rule blocked execution, but the blocked call consumed `--max-turns 1` and ended as `cancelled/max_turns_reached`.

This proved that permission denial is a safety backstop, not schema removal.

### Empty schema without a text-only prompt

The recognized allowlist and denylist intersection produced `tool_count: 0`, but Grok could still narrate a fake `<tool_call>` in plain text because it retained its coding-agent system prompt.

The final path also uses a static text-only `--system-prompt-override`. With both controls in place, the negative file-read probe returned `NO_READ_TOOL` with `end_turn` and made no tool call.

## One-shot contract

For each chat or Oracle request, RepoPrompt:

1. Resolves Grok through the trusted one-shot launch preflight and revalidates the executable identity immediately before execution.
2. Validates the selected model against the discovered Grok model snapshot.
3. Creates a `0700` temporary directory and writes the prompt to a `0600` file. Prompt text never appears on argv.
4. Sets the child working directory to that temporary directory, not the workspace.
5. Runs Grok with JSON output, verbatim input, one maximum tool cycle, no subagents, no plan mode, no memory, no web search, an empty prepared toolset, and a text-only system prompt.
6. Keeps the existing permission denials as an execution backstop.
7. Parses one final JSON object and accepts only successful terminal stop reasons.
8. Terminates the child on stream cancellation and removes the temporary directory after every outcome.

A representative invocation is:

```text
grok \
  --prompt-file <private-file> \
  --output-format json \
  --verbatim \
  --max-turns 1 \
  --tools read_file \
  --disallowed-tools read_file,search_tool,use_tool \
  --system-prompt-override <text-only-prompt> \
  --no-subagents --no-plan --no-memory --disable-web-search
```

## Model and effort handling

The one-shot path validates the selected compound model against the discovered catalog before splitting it into base model and effort.

Headless `--reasoning-effort` and ACP `_meta.reasoningEffort` are different interfaces:

- ACP must preserve the exact advertised wire value. For example, the semantic `max` option may advertise `maximum`.
- The headless CLI documents canonical effort tokens, including `max`, and resolves them after loading the model menu.

The live one-shot runs exercised `high`. They did not compare `max` and `maximum` side by side, so this PR does not claim that equivalence has been measured on the installed CLI.

## Failure behavior

The adapter rejects:

- an unknown or stale selected model;
- a failed stored-key lookup, instead of silently falling back to a different native Grok credential;
- unsafe or changed executable identity;
- resume attempts on the one-shot path;
- process timeout or nonzero exit;
- malformed, empty, or multiple JSON payloads;
- empty successful completions; and
- non-success stop reasons such as cancellation or max-token termination.

Structured Grok errors on stdout remain available when the process exits unsuccessfully. Legacy stop-reason spellings are normalized before terminal validation.

## Validation

Local checks completed against commit `6429fd41`:

- `make dev-format`
- `make dev-lint`
- `GrokBuildCLIProviderProcessTests`: 8 of 8 passed
- key-provider failure propagation test confirmed a keychain read failure prevents process launch
- `GrokBuildModelRoutingTests`: 8 of 8 passed
- Debug app packaged and relaunched through Conductor with explicit ad-hoc signing

Live Grok CLI checks confirmed:

- `tool_count: 0`
- `tools: []`
- no tool call
- four direct prompts returned completed text answers
- the file-read probe returned `NO_READ_TOOL/end_turn` with the final system prompt
- the rebuilt CE debug CLI routed `oracle_send` to Grok 4.6 High and returned exactly `GROK_ORACLE_CLI_SMOKE_OK` through the one-shot `--prompt-file` path
- temporary Oracle model settings were restored and no one-shot prompt directory remained

Not completed:

- The Oracle-sized prompt and its negative control both exceeded the bounded 90-second probe. That validation remains open.
- Hosted CI completed with all 8 checks passing. App shard 2 initially hit the unrelated `ContextBuilderRunLifecycleTests.testRealConnectionCleanupCannotEraseContextBeforeCommit` race, then passed on the failed-job rerun without code changes.

## Scope

This PR changes three files:

```text
Sources/RepoPrompt/Infrastructure/AI/Providers/GrokBuild/
├── GrokBuildCLIProvider.swift
└── GrokBuildOneShotHeadlessAgentProvider.swift

Tests/RepoPromptTests/AI/
└── GrokBuildCLIProviderProcessTests.swift
```

It does not change shared ACP draining, Agent Mode's Grok transport, or other providers. The ACP drain work remains separate in #847.

### DIFF
--- a/Sources/RepoPrompt/Infrastructure/AI/Providers/GrokBuild/GrokBuildCLIProvider.swift
+++ b/Sources/RepoPrompt/Infrastructure/AI/Providers/GrokBuild/GrokBuildCLIProvider.swift
@@ -1,7 +1,7 @@
 import Foundation
 
-/// Grok Build CLI provider for non-agent use (chat, Oracle, AI queries) backed by ACP.
-/// Runs a fresh prompt-only ACP session per request without injecting RepoPrompt MCP/tools.
+/// Grok Build CLI provider for non-agent use (chat, Oracle, AI queries).
+/// Runs a fresh prompt-only one-shot process per request without injecting RepoPrompt MCP/tools.
 final class GrokBuildCLIProvider: AIProvider {
     private let activeProviders = ActiveGrokBuildCLIProviderStore()
 
@@ -31,9 +31,8 @@ final class GrokBuildCLIProvider: AIProvider {
         model: AIModel,
         maxTokens _: Int? = nil
     ) async throws -> AsyncThrowingStream<AIStreamResult, Error> {
-        let provider = GrokBuildACPHeadlessAgentProvider(
-            config: Self.makeHeadlessConfig(modelName: grokBuildModelName(for: model)),
-            workspacePath: nil
+        let provider = GrokBuildOneShotHeadlessAgentProvider(
+            config: Self.makeHeadlessConfig(modelName: grokBuildModelName(for: model))
         )
         activeProviders.insert(provider)
 
@@ -116,7 +115,7 @@ final class GrokBuildCLIProvider: AIProvider {
                 if let value = result.cost { cost = value }
             case "error":
                 throw AIProviderError.invalidConfiguration(
-                    detail: result.text ?? "Grok Build ACP reported an error"
+                    detail: result.text ?? "Grok Build reported an error"
                 )
             default:
                 continue
@@ -264,21 +263,21 @@ private extension AIStreamResult {
 
 private final class ActiveGrokBuildCLIProviderStore: @unchecked Sendable {
     private let lock = NSLock()
-    private var providers: [ObjectIdentifier: GrokBuildACPHeadlessAgentProvider] = [:]
+    private var providers: [ObjectIdentifier: GrokBuildOneShotHeadlessAgentProvider] = [:]
 
-    func insert(_ provider: GrokBuildACPHeadlessAgentProvider) {
+    func insert(_ provider: GrokBuildOneShotHeadlessAgentProvider) {
         lock.lock()
         providers[ObjectIdentifier(provider)] = provider
         lock.unlock()
     }
 
-    func remove(_ provider: GrokBuildACPHeadlessAgentProvider) {
+    func remove(_ provider: GrokBuildOneShotHeadlessAgentProvider) {
         lock.lock()
         providers.removeValue(forKey: ObjectIdentifier(provider))
         lock.unlock()
     }
 
-    func removeAll() -> [GrokBuildACPHeadlessAgentProvider] {
+    func removeAll() -> [GrokBuildOneShotHeadlessAgentProvider] {
         lock.lock()
         let current = Array(providers.values)
         providers.removeAll()
--- a/Sources/RepoPrompt/Infrastructure/AI/Providers/GrokBuild/GrokBuildOneShotHeadlessAgentProvider.swift
+++ b/Sources/RepoPrompt/Infrastructure/AI/Providers/GrokBuild/GrokBuildOneShotHeadlessAgentProvider.swift
@@ -4,7 +4,7 @@ import Foundation
 /// Agent Mode continues to use `grok agent stdio`; this adapter uses the documented
 /// one-shot JSON CLI and preserves the existing trusted Grok executable preflight.
 final class GrokBuildOneShotHeadlessAgentProvider: HeadlessAgentProvider {
-    typealias APIKeyProvider = @Sendable () async -> String?
+    typealias APIKeyProvider = @Sendable () async throws -> String?
 
     private let config: GrokBuildAgentConfig
     private let launchResolver: GrokBuildACPLaunchResolver
@@ -17,7 +17,7 @@ final class GrokBuildOneShotHeadlessAgentProvider: HeadlessAgentProvider {
         launchResolver: GrokBuildACPLaunchResolver = GrokBuildACPLaunchResolver(),
         requestTimeout: TimeInterval = 6000,
         apiKeyProvider: @escaping APIKeyProvider = {
-            try? await KeyManager().getAPIKey(for: .grok)
+            try await KeyManager().getAPIKey(for: .grok)
         }
     ) {
         self.config = config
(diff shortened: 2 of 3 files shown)


najibninaba (approved):
Thanks, Daniel. This is a strong, focused correction.

I like that you separated non-Agent chat and Oracle from the Grok ACP coding-agent runtime while keeping Agent Mode unchanged. The private prompt-file boundary, executable identity checks, strict terminal handling, and process cleanup keep the new one-shot path contained.

I reviewed exact head `6429fd41e6a01ec6bbb1d3f8401fc06550e044eb`.

Review method:

- Exact-head code, GitHub, and CI review with Codex
- Full RepoPrompt Context Builder and Oracle review
- Independent RP Design review using Claude Fable 5 High
- Independent adversarial review using Grok 4.6
- Independent test and release review using Kimi K3
- Focused correctness, architecture, and release-testing lanes
- Local reproduction of the earlier hosted Context Builder failure against both the PR head and exact base

My call is **GO-WITH-FOLLOW-UPS for merge and inclusion in the next release**.

I did not find a code defect that I would block this PR on.

The latest commit also closes a real failure-handling gap. A stored-key lookup error now stops before launching Grok instead of being treated as an absent key and silently falling through to another authentication path. A genuinely missing stored key still permits Grok CLI-managed authentication.

The independent review lanes converged on the same conclusion:

- Agent Mode remains on the existing ACP path.
- The one-shot path preserves the trusted executable and model authorities.
- Prompt text stays out of argv and is written under a private temporary directory.
- Cancellation terminates the child process and cleanup remains reliable.
- Successful output requires a valid JSON completion and a successful terminal stop reason.
- The live Grok 1.0.5 evidence demonstrates an empty prepared toolset, no file-read capability, direct text responses, and an authenticated Oracle request through the new path.

Validation at this head:

- `GrokBuildCLIProviderProcessTests`: 8/8 passed in the PR validation record
- The key-provider failure test confirms lookup errors propagate before process launch
- All eight GitHub checks are green
- The original app shard 2 failure was reproduced identically on the exact base and was unrelated to this PR; its rerun has passed

I recommend keeping these as explicit follow-ups:

- Harden the bounded startup case where a persisted Grok model can be requested before the model registry finishes warming.
- Give the one-shot path its own capability probe and consolidate Grok credential ownership.
- Qualify `max` versus `maximum`, large-prompt timing, cancellation visibility, and progress behavior before or during the release cut as appropriate.
- Update the settings guidance that still describes all headless Grok tasks as ACP-based.

I do not consider these merge blockers at this head. The authenticated Grok smoke and zero-tool evidence satisfy the same release-qualification standard we applied in #838. Broader runtime compatibility remains an ongoing qualification responsibility when the supported Grok CLI changes.

I am approving the code at this head.

@baron, I am comfortable with the direction. I will defer to you on whether the remaining qualification follow-ups should be completed before the release cut or tracked separately.

Nice work keeping the change contained and adding the live evidence needed to verify the new path.

PR #857, opened 2026-08-23
Harden Apple identity migration bridge

baron:
## Summary

- move migration authority from mutable UserDefaults into a dual-identity Keychain journal with resumable bridge records
- authenticate the journal and bridge-manifest decrypt ACLs against exactly the legacy and successor designated requirements before accepting or propagating them
- scope each migration attempt with a unique generic-password service name, avoiding orphan collisions on the actual Keychain primary key
- freeze the version-2 migration catalog while retaining an authenticated committed bridge as the authority for accounts added later
- validate the embedded successor anchor at runtime and build a minimal universal signed anchor instead of copying the main executable
- repair credentials into the selected runtime backend, fail closed on invalid migration state, and surface blocked-update guidance in Settings
- keep Tip builds migration-disabled and tighten release signing and post-copy anchor validation

## Validation

Review-fix pass:

- `swift test --filter KeychainServiceTests` — 15 passed
- `swift test --skip-build --filter SecureStorageIdentityMigrationTests` — 21 passed
- `swift test --filter SecureStorageAccountCatalogTests` — 6 passed
- targeted release-tooling source-contract test — passed
- `make dev-lint`
- `make guardrails`
- mandatory contribution commit and push preflights, including staged/outgoing secret scans

Earlier implementation pass:

- focused Sparkle startup-gating and secure-storage repair tests
- targeted release-tooling Python tests
- bash syntax check for `Scripts/sign_staged_release.sh`
- universal anchor compile and architecture verification

No live Keychain mutation probe, visible app launch, release build, or redundant full test suite was run for this follow-up.

### DIFF
--- a/.github/workflows/main-tip.yml
+++ b/.github/workflows/main-tip.yml
@@ -11,6 +11,14 @@ on:
         description: Commit SHA on protected main to publish as tip
         required: false
         type: string
+      identity_migration_phase:
+        description: Apple identity migration mode for a manual Tip publish
+        required: true
+        default: disabled
+        type: choice
+        options:
+          - disabled
+          - legacy-preparer
 
 concurrency:
   group: >-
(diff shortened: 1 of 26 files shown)


baron:
Addressed the requested review changes in `3d555b5b`:

1. Journal and bridge-manifest items now have their decrypt ACL authenticated before their JSON is accepted, and an existing manifest ACL is revalidated before it is copied to newly created credentials. The accepted principal set is exactly the legacy and successor designated requirements; wildcard, missing, duplicate, malformed, or additional principals fail closed. Added forged-journal, forged-manifest, and ACL-set regressions.
2. Each migration attempt now uses an attempt-specific Keychain service name, so the attempt UUID participates in the generic-password primary key. Added lost-journal/orphaned-attempt recovery coverage.
3. The version-2 account catalog is frozen for the migration window, while an authenticated committed bridge remains the canonical backend for accounts introduced later. Added drift-gate and later-account coverage.
4. Refreshed the stale Sparkle startup source contract that was failing Style.

Focused validation: 15 Keychain tests, 21 identity-migration tests, 6 catalog tests, the targeted release-tooling test, `make dev-lint`, `make guardrails`, and both mandatory contribution preflights. No live Keychain mutation probe or redundant full rebuild was run.

PR #878, opened 2026-08-26
Separate sidebar command progress from bulk selection UI

### LINKED ISSUE #875: Agent Mode sidebar briefly shows "0 selected" bulk UI during direct actions
## Summary

Using a direct action on one Agent Mode sidebar row briefly puts the whole sidebar into multi-select presentation.

During a stash, the sidebar shows:

- a header reading "0 selected"
- selection circles beside every session
- hidden normal row actions

The UI returns to normal when the action finishes. Stash and delete make the problem easiest to see because they can remain pending while cascade and cleanup work runs.

## Steps to reproduce

1. Open RepoPrompt CE with an Agent Mode session that contains a conversation.
2. Hover over the session in the sidebar.
3. Click the stash action.
4. Observe the sidebar while the stash is in progress.

## Actual behavior

The sidebar briefly displays bulk-selection controls even though the user did not select any sessions. The header reports "0 selected."

## Expected behavior

A direct row command should not display multi-select controls.

Long-running direct actions should still show meaningful progress, but without a selection count or selection circles.

## Affected actions

The same state path is used by:

- active-session stash
- active-session delete
- archived-session delete
- Clear Archived
- direct pin and unpin

Pin and unpin may finish before SwiftUI draws the intermediate state, but they publish the same state internally.

## Investigation

This behavior dates to PR #802, which added sidebar multi-select and bulk actions:

https://github.com/repoprompt/repoprompt-ce/pull/802

Direct row commands intentionally use the shared bulk mutation coordinator. That preserves workspace fencing, cascade handling, cleanup, and result notices.

The presentation bug comes from `AgentSidebarSelectionState.isSelectionMode`:

```swift
var isSelectionMode: Bool {
    !selectedIdentities.isEmpty || inFlightAction != nil
}
```

A direct command creates an `inFlightAction` without adding any selected identities. The view then enters selection mode, displays the live selection count as zero, and enables selection presentation for every row.

Relevant files:

- `Sources/RepoPrompt/Features/AgentMode/ViewModels/UI/AgentSidebarSelectionState.swift`
- `Sources/RepoPrompt/Features/AgentMode/ViewModels/UI/AgentSessionSidebarUIStore.swift`
- `Sources/RepoPrompt/Features/AgentMode/Views/Components/AgentSessionsSidebarView.swift`
- `Sources/RepoPrompt/Features/AgentMode/Views/Components/AgentSessionRows.swift`
- `Sources/RepoPrompt/Features/AgentMode/ViewModels/AgentModeViewModel+SidebarUI.swift`

Current `main` still contains this behavior. No later sidebar change altered the state contract.

## Constraints

A fix should preserve the shared mutation coordinator and its workspace and operation-token checks.

It should also preserve progress and interaction locking for a real bulk action after selected rows disappear from the sidebar during stash or delete. Changing selection mode to depend only on a nonempty live selection would not handle that case.

## Acceptance criteria

- Direct row commands never show "0 selected" or row selection circles.
- Long-running direct stash and delete operations have a non-selection progress indicator.
- Real bulk operations keep their progress visible if selected rows disappear before completion.
- Row mutations, archived restore, rename, and Clear Archived remain unavailable while a sidebar mutation is running.
- Workspace changes still invalidate stale operations and notices.
- Tests cover the in-progress state for direct and selection-driven actions, not only the final result.

aAtila:
## Summary

A direct action on one sidebar row put the whole sidebar into multi-select presentation for the length of the operation. Selection circles appeared beside every session, normal row actions hid, and the header read "0 selected" when nothing was selected. Stash and delete showed it most because they stay pending through cascade and cleanup work, but archived delete, Clear Archived, and direct pin and unpin all publish the same state. Direct actions now show plain progress with no selection count. Bulk-selection actions keep their real count.

## Changes

- Each sidebar mutation records whether it came from bulk selection or from a direct command. The shared mutation coordinator and its workspace/token fencing stay as they were.
- The selection UI no longer keys off the global mutation lock, so a direct action never draws selection circles or a selected-count header.
- Bulk actions keep showing progress after reconciliation removes the selected rows. The active mutation now carries its own copy of the operation targets instead of reading rows that may be gone.
- Row, archived-session, rename, and confirmation callbacks bail out when the workspace changed or another mutation is already running.
- New tests cover both origins, in the store and against a suspended coordinator: reentry rejection, reconciliation, workspace invalidation, and notice wording.

## Validation

CI is green on all eight checks, including the four app test shards, Provider Tests, Style, and the Sentry-enabled build.

Locally:

- `make dev-test FILTER=AgentSessionSidebarUIStoreTests`, 11 tests passed.
- `make dev-test FILTER=BackgroundComposeTabAdmissionTests`, 39 tests passed.
- `make dev-lint`, passed.
- `ALLOW_ADHOC_SIGNING=1 make dev-build`, passed.

Closes #875

### DIFF
--- a/Sources/RepoPrompt/Features/AgentMode/ViewModels/AgentModeViewModel+SidebarSessions.swift
+++ b/Sources/RepoPrompt/Features/AgentMode/ViewModels/AgentModeViewModel+SidebarSessions.swift
@@ -590,6 +590,12 @@ extension AgentModeViewModel {
             return cached.projection
         }
 
+        let existingSelectionIdentities = Set(
+            composeTabs.map { AgentSidebarSelectionIdentity.active(tabID: $0.id) }
+                + stashedTabs.map {
+                    AgentSidebarSelectionIdentity.archived(stashedTabID: $0.id, tabID: $0.tab.id)
+                }
+        )
         let canonicalRows = buildSidebarSessions(
             for: composeTabs,
             workspaceID: workspaceID,
@@ -640,6 +646,7 @@ extension AgentModeViewModel {
                 in: canonicalRows,
                 searchText: sidebarSnapshot.searchText
             ),
+            existingSelectionIdentities: existingSelectionIdentities,
             renderedSelectionOrder: renderedSelectionOrder
         )
         sidebarListProjectionCache = (key, projection)
--- a/Sources/RepoPrompt/Features/AgentMode/ViewModels/AgentModeViewModel+SidebarUI.swift
+++ b/Sources/RepoPrompt/Features/AgentMode/ViewModels/AgentModeViewModel+SidebarUI.swift
@@ -168,6 +168,13 @@ extension AgentModeViewModel {
             return workspaceManager?.activeWorkspaceID == workspaceID
                 && ui.sessionSidebar.isCurrentBulkAction(token: token, workspaceID: targets.workspaceID)
         }
+        let onProjectionRemovalCommitted: PromptViewModel.ComposeTabsProjectionRemovalCallback = { [weak self] tabIDs in
+            self?.ui.sessionSidebar.retireCommandRowProgress(
+                token: token,
+                forRemovedTabIDs: tabIDs,
+                workspaceID: workspaceID
+            )
+        }
         var notice: AgentSidebarBulkActionNotice?
 
         switch action {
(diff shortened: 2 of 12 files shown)


w-winter (changes requested):
Thanks for this fix.  The UI change looks good to me in the debug app, but there's one lifecycle race to resolve:

`retireCommandRowProgress` is called from delayed `handleComposeTabsDidRemove` callbacks but matches the current operation only by workspace and tab ID.  So if command A stalls after projection removal, a workspace roundtrip can invalidate A, restore the same tab, and start command B.  When A resumes, its callback can retire B’s progress because no operation token is checked.  Checking only current row absence or removal kind is insufficient because both can also match command B.

Please correlate retirement with the originating mutation, require it to match the current sidebar token, and a regression test for stale callbacks after a workspace round trip & same-ID command turnover

aAtila:
Fix pushed.

Row progress now carries the token from the sidebar command that started the removal. PromptViewModel calls the retirement callback right after it publishes the new projection and before the awaited cleanup work. The store checks the token before changing the current operation. I also removed the old retirement call from the generic did-remove handler.

w-winter (approved):
clean fix and e2e looks good to me

PR #884, opened 2026-08-26
Add support for multiple Oracles

dsebban:
## Summary

This PR turns **Oracle groups** into a supported, durable product path across both the macOS app and `repoprompt-mcp --backend headless`.

A reviewer can think of it as one request sent to an **ordered roster of 2–5 independent Oracle models**. Every model receives the same input, runs independently, and keeps its own answer or error. The app and headless MCP now share one grouped-turn lifecycle, while model routing and UI/transport concerns remain surface-owned.

> [!IMPORTANT]
> This is parallel independent review, **not** answer synthesis. The runtime does not rank, vote on, merge, retry, or choose a winning answer.

## Third-party review follow-up

The latest review comments were treated as hypotheses and checked against the owning code, predecessor behavior, and focused tests. The final patch keeps the real fixes and removes the workaround-shaped changes.

| Review finding | Decision | Final shape |
|---|---|---|
| A named single-chat continuation could be promoted after enabling more models | **Accepted** | Explicit named continuations stay on the existing single-chat path; a real headless regression test covers the settings change |
| Child bridge writes used a blocking syscall, so the timeout was not reliable | **Accepted** | Reuses the existing non-blocking writer; a stalled-destination test proves bounded failure |
| Grouped child MCP needed a strict read-only capability set | **Accepted, narrowed** | Child-visible tools map only to structural exploration, file reads/search, and read-only Git operations |
| Cancelled lane output was inferred from error strings | **Accepted, corrected at the owner** | Typed `.cancelled` status now flows through `ToolOutputFormatter`; no string heuristic remains |
| Context Builder prelaunch failures could leave generation state stuck | **Accepted, corrected at the owner** | The generation-owning view model performs cleanup using its exact generation; the external property-wrapper/`generation + 2` guess was removed |
| A prepared plan could leak across provider suspension | **Not substantiated** | No change: plans are consumed before suspension, revocation discards them, and shutdown clears them |

The follow-up is commit [`11a0c289`](https://github.com/repoprompt/repoprompt-ce/commit/11a0c289).

## Reviewer quick start

1. Configure a primary Oracle and up to four additional models.
2. Start or continue an Oracle request as usual.
3. Open the `Oracles · N` result to inspect each lane in roster order.
4. For implementation review, start with `OracleGroupRuntime.swift`, then persistence, then the two surface adapters.

### 1. Configure an ordered Oracle roster

The existing primary Oracle remains first. Reviewers can add up to four additional models. Order is significant, and duplicate model identifiers are intentionally preserved.

<p align="center">
  Oracle Models settings with a primary Oracle and Oracle 2
</p>

The same semantic settings are available to direct headless MCP through:

- `models.planning_model`
- `models.additional_oracle_models`

The app and headless process keep separate settings stores, but both use the same validation rules.

### 2. Grouped results stay compact in the UI

A grouped result is represented by one `Oracles · N` pill rather than pretending the lanes are one synthesized answer.

<p align="center">
  Compact Oracles 2 result pill
</p>

### 3. Each lane remains independently reviewable

The result view and exports keep lane output separate and ordered. Partial failures and cancellation remain visible as structural lane outcomes; successful answers are not discarded because another lane failed.

<p align="center">
  Grouped Oracle result showing separate Oracle and Oracle 2 lanes
</p>

## Behavior at a glance

| Request | Route | Durable group record | Result behavior |
|---|---|---:|---|
| One model (`N=1`) | Existing app/headless path | No | Existing single-Oracle behavior is preserved |
| 2–5 models (`N>1`) | Shared `OracleGroupRuntime` | Yes | One ordered result per lane |
| Group continuation | Stored group roster | Existing record is extended | Configured-roster mismatch is rejected |
| Named single-chat continuation | Existing single-chat path | No promotion | It is never silently converted into a group |

Group status is derived from lane outcomes:

- **`completed`** — every lane completed without warnings.
- **`partial_failure`** — the primary completed, but an additional lane failed, was cancelled, or warned.
- **`failed`** — the primary lane did not complete.

## Runtime shape

```mermaid
flowchart LR
    App[macOS app] --> Route{Roster size}
    MCP[Headless MCP] --> Route
    Route -->|N = 1| Existing[Existing single-Oracle path]
    Route -->|N = 2–5| Runtime[OracleGroupRuntime]
    Runtime --> Claim[Acquire durable claim]
    Claim --> Reload[Reload + recover interrupted turn]
    Reload --> Prepare[Persist prepared turn]
    Prepare --> Lanes[Run ordered independent lanes]
    Lanes --> Drain[Drain cancellation/callbacks]
    Drain --> Publish[Persist + publish exact terminal result]
    Publish --> Release[Release claim]
```

```text
RepoPromptDomainRuntime/
├── OracleGroupContracts.swift          # identities, roster, results, statuses, MCP encoding
├── OracleGroupRuntime.swift            # only mutating N>1 lifecycle entry point
├── OracleGroupCoordinator.swift        # structured parallel lane execution
├── OracleGroupPersistence.swift        # durable documents, indexes, CAS, journal recovery
├── OracleGroupClaims.swift             # cross-process lifecycle serialization
└── OracleGroupTerminalPublisher.swift  # retry exact canonical terminal outcome

Surface adapters
├── OracleViewModel+Groups.swift         # macOS app routing and chat projections
└── DirectHeadlessOracleAdapter.swift    # headless routing, carrier validation, child launches
```

### Shared grouped-turn runtime

- Adds `OracleGroupRuntime` as the sole mutating entry point for `N>1` turns.
- Uses structured parallel execution while preserving roster order regardless of completion order.
- Preserves duplicate model identifiers and assigns stable group, member, lane, turn, run, invocation, and claim identities.
- Publishes one structural outcome for every lane.

### Durable lifecycle and recovery

- Holds one durable claim across preparation, lane execution, cancellation drain, terminal publication, and release.
- Reloads after claim acquisition to avoid acting on stale state.
- Recovers interrupted prepared turns lazily.
- Uses revision compare-and-swap, filesystem mutation locks, a transaction journal, and a member-to-group continuation index.
- Retries the same canonical terminal result if publication fails; it does not invent replacement output.
- Keeps cancellation under claim until callbacks drain and cancelled lane outcomes are durable.

### App integration

- Routes configured grouped app requests through the shared runtime.
- Rebuilds lane chat projections by canonical member identity and rejects conflicting group metadata.
- Keeps group continuations on their stored roster.
- Adds the compact `Oracles · N` presentation and ordered lane export formatting.

### Context Builder integration

- Uses the grouped app path for multi-Oracle plan/review follow-ups.
- Shows primary-lane progress while retaining every final lane result.
- Fences callbacks by generation, group, turn, lane, and sequence so stale work cannot overwrite a newer run.

### Direct headless MCP parity

- Adds durable grouped start and continuation support to `repoprompt-mcp --backend headless`.
- Allows continuation from any member chat ID while returning the primary member as the root chat ID.
- Resolves immutable child launch plans before dispatch.
- Binds group, lane, and claim identity through credential scope, reservation, environment, token redemption, and child handshake.
- Fails closed before claim acquisition or history mutation when carrier authorization is incomplete or mismatched.

### Settings and schema authority

- Adds an ordered `models.additional_oracle_models` setting with a maximum of four entries.
- Uses shared roster descriptors for app and headless validation.
- Makes the Swift canonical MCP definitions authoritative and updates the generated review snapshot.
- Rejects incomplete indexes, mismatched group filenames, partial launch identities, and unknown future schemas.

### Schema compatibility

- Reads existing schema 1 group documents as schema 2 by deriving their missing terminal status.
- Writes schema 2 on the next actual mutation; there is no eager history rewrite.
- Unknown future schema versions fail closed.

## Explicit limitations and non-goals

- **No synthesis or arbitration:** answers are not merged, ranked, voted on, summarized, or reduced to a winner.
- **No hidden retries/advisers:** each configured lane is the work that runs.
- **Maximum roster size is five:** one primary plus at most four additional models.
- **`N=1` stays outside the group runtime:** it creates no durable group document.
- **Continuations cannot change models:** group continuations use the stored roster; headless continuation model overrides are rejected.
- **Direct grouped Context Builder requires a frozen pack:** `context_pack_ref` must be a verified canonical `oracle-pack:sha256:<digest>` reference. Raw `instructions` remain available for `N=1` only.
- **Routing remains surface-owned:** provider selection, ACP/Grok transport behavior, app projections, and headless child launching are not moved into the domain runtime.
- **Settings storage is not unified:** app and direct headless settings remain separate stores with shared semantics.
- **No migration aliases for unshipped experimental pair contracts.**

## Failure and safety boundaries

| Boundary | Contract |
|---|---|
| Additional lane fails | Primary success is retained; group becomes `partial_failure` |
| Primary lane fails | Group becomes `failed` |
| User cancels | Lane callbacks drain; structural cancelled outcomes persist before claim release |
| Publisher fails | Prepared turn remains recoverable; exact terminal outcome is retried |
| Roster mismatch | Rejected before provider dispatch |
| Invalid headless carrier/frozen pack | Rejected before claim acquisition or group history mutation |
| Missing canonical group document for a grouped projection | Fails closed; does not fall back to `N=1` |
| Unknown future persisted schema | Fails closed |

## Compatibility and scope

- Existing single-Oracle app and headless behavior is intentionally preserved.
- Existing schema 1 Oracle group history remains readable.
- Exact lane response/error text is preserved; only metadata is trimmed where required.
- The PR centralizes grouped lifecycle behavior without centralizing surface-specific routing or presentation.
- The architectural contract is documented in [`docs/architecture/oracle-groups-rewrite.md`](https://github.com/repoprompt/repoprompt-ce/blob/feature/oracle-group-runtime/docs/architecture/oracle-groups-rewrite.md).

## Validation

The Checks tab is authoritative for hosted validation of the latest pushed head. The review-follow-up commit was locally validated with:

- `make dev-test FILTER=ContextBuilderOracleGroupStateTests` — 6 tests passed
- Combined focused suites for `ContextBuilderOracleResultTests`, `DirectHeadlessOracleGroupTests`, `DirectHeadlessCompositionTests`, and `OracleLaneMarkdownFormatterTests` — all passed
- `make dev-lint` — SwiftFormat and strict SwiftLint passed
- `ALLOW_ADHOC_SIGNING=1 make dev-build` — packaged and codesign-verified the real debug app plus embedded MCP helper
- Mandatory commit and push contribution preflights — guardrails and redacted secret scans passed
- Live CE debug-app Context Builder smoke with the configured two-model roster:
  - Oracle: Codex `gpt-5.6-sol`, high effort — **Completed**
  - Oracle 2: Custom `gpt-5.6-sol-pro` — **Completed**
  - both returned the exact marker and remained separate, ordered lane results

The broader branch validation performed during development also includes:

- Full root test suite and four hosted root shards
- Provider tests and Sentry-enabled build
- `RepoPrompt` and `repoprompt-mcp` product builds
- Focused runtime, persistence, coordinator, contract, app routing, Context Builder, authorization, formatting, and MCP catalog suites
- Live MCP grouped start/continuation across mixed schema 1/schema 2 history

## Suggested review order

1. **Contract and lifecycle**
   - `Sources/RepoPromptDomainRuntime/OracleGroupContracts.swift`
   - `Sources/RepoPromptDomainRuntime/OracleGroupRuntime.swift`
   - `Sources/RepoPromptDomainRuntime/OracleGroupCoordinator.swift`
2. **Durability and failure handling**
   - `Sources/RepoPromptDomainRuntime/OracleGroupPersistence.swift`
   - `Sources/RepoPromptDomainRuntime/OracleGroupClaims.swift`
   - `Sources/RepoPromptDomainRuntime/OracleGroupTerminalPublisher.swift`
3. **Surface adapters**
   - `Sources/RepoPrompt/Features/Chat/ViewModels/Oracle/OracleViewModel+Groups.swift`
   - `Sources/RepoPromptMCP/DirectHeadlessOracleAdapter.swift`
4. **UI and Context Builder**
   - `AgentOraclePill.swift`
   - `ContextBuilderOracleGroupState.swift`
   - `ContextBuilderAgentViewModel.swift`
5. **Boundary tests**
   - `OracleGroupRuntimeTests.swift`
   - `OracleGroupPersistenceTests.swift`
   - `DirectHeadlessOracleGroupTests.swift`
   - `OracleGroupBoundaryTests.swift`
   - `OracleLaneLaunchAuthorizationTests.swift`


docs/architecture/oracle-groups-rewrite.md
```diff
@@ -0,0 +1,176 @@
+# Oracle groups architecture
```
dsebban, line 1: lan should be excluded from commit , you can take a summary of it and re work the PR description

Sources/RepoPrompt/App/AppDomainRuntimeComposition.swift
```diff
@@ -71,7 +74,7 @@ final class AppDomainRuntimeComposition: Sendable {
         let workspaceStorageDirectory = customStoragePath.map {
             URL(fileURLWithPath: $0, isDirectory: true)
         } ?? root.appendingPathComponent("Workspaces", isDirectory: true)
-        runtime = MCPDomainRuntime(
+        let runtime = MCPDomainRuntime(
```
dsebban, line 76: why add this let ?

Sources/RepoPrompt/App/AppDomainRuntimeComposition.swift
```diff
@@ -84,6 +87,19 @@ final class AppDomainRuntimeComposition: Sendable {
                 metrics: AppDomainRuntimeMetrics.editFlowSink
             )
         )
+        self.runtime = runtime
```
dsebban, line 89: this is a new change , is it idiomatic in this project ?

dsebban (commented):
`docs/architecture/oracle-groups-clean-runtime-plan.md` is local-only and is not part of this PR, so this cannot be attached inline: plan should be excluded from commit , you can take a summary of it and re work the PR description

Sources/RepoPrompt/App/AppDomainRuntimeComposition.swift
```diff
@@ -84,6 +87,19 @@ final class AppDomainRuntimeComposition: Sendable {
...
+        oracleConversationStore = DomainOracleConversationStore(
```
dsebban, line 90: is it aifiomatic to add these variable here ? should they be somwhere else ?

Sources/RepoPrompt/Features/AgentMode/ViewModels/UI/AgentModelsSettingsViewModel.swift
```diff
@@ -180,6 +181,33 @@ final class AgentModelsSettingsViewModel: ObservableObject {
         displayName(forChatModelRaw: profileSnapshot.planningModelRaw, fallback: "Select an Oracle model")
     }
 
+    var additionalOracleModelRaws: [String] {
+        profileSnapshot.additionalOracleModelRaws
+    }
+
+    var oracleCount: Int {
+        1 + additionalOracleModelRaws.count
```
dsebban, line 189: what's this weird 1 = ?

Sources/RepoPrompt/Features/AgentMode/ViewModels/UI/AgentModelsSettingsViewModel.swift
```diff
@@ -180,6 +181,33 @@ final class AgentModelsSettingsViewModel: ObservableObject {
...
+    }
+
+    var canAddOracle: Bool {
```
dsebban, line 192: is it idiomatic to declare bool like that in this file ?

Sources/RepoPrompt/Features/AgentMode/Views/Components/AgentOraclePill.swift
```diff
@@ -110,8 +111,29 @@ enum AgentOraclePillLogic {
         in sessions: [ChatSession],
         streamingSessionIDs: Set<UUID>
     ) -> ChatSession? {
-        latestStreamingSession(in: sessions, streamingSessionIDs: streamingSessionIDs)
+        guard let latest = latestStreamingSession(in: sessions, streamingSessionIDs: streamingSessionIDs)
```
dsebban, line 114: is it idiomatic to use guard in this file ? is the added code out of place ?

Sources/RepoPrompt/Features/AgentMode/Views/Components/AgentOraclePill.swift
```diff
@@ -110,8 +111,29 @@ enum AgentOraclePillLogic {
...
             ?? sessions.max(by: { $0.savedAt < $1.savedAt })
+        else { return nil }
```
dsebban, line 116: seems odd

Sources/RepoPrompt/Features/AgentMode/Views/Components/AgentOraclePill.swift
```diff
@@ -110,8 +111,29 @@ enum AgentOraclePillLogic {
...
+            $0.oracleGroupID == groupID && $0.oracleLaneIndex == 0
+        }) ?? latest
+    }
+
+    static func canCopyAll(_ session: ChatSession) -> Bool {
+        session.oracleGroupID != nil
+    }
+
+    static func aggregateOracleCount(configuredAdditionalCount: Int, sessions: [ChatSession]) -> Int {
+        let configured = min(max(1 + configuredAdditionalCount, 1), 5)
```
dsebban: min max . why so compliated is it overengineered ?

Sources/RepoPrompt/Features/AgentMode/Views/Components/AgentOraclePill.swift
```diff
@@ -110,8 +111,29 @@ enum AgentOraclePillLogic {
...
+        guard let latest = sessions.max(by: { $0.savedAt < $1.savedAt }),
+              let groupID = latest.oracleGroupID
+        else { return configured }
+        let projected = sessions
+            .filter { $0.oracleGroupID == groupID }
+            .compactMap(\.oracleGroupSize)
+            .max() ?? 1
```
dsebban: seems over complicated is it idiomatic ?

Sources/RepoPrompt/Features/AgentMode/Views/Components/AgentOraclePill.swift
```diff
@@ -176,6 +198,35 @@ enum AgentOraclePillLogic {
...
+    ) -> String? {
+        if let last = liveMessages.last(where: { !$0.isUser }) {
+            return last.content
+        }
+        return storedMessages.last(where: { !$0.isUser })?.rawText
+    }
+
+    static func assistantContentIndicatesFailure(_ content: String?) -> Bool {
+        guard let content else { return false }
+        if content.contains("\n--\nError:\n") { return true }
```
dsebban, line 217: parsing strings is idiomatic in reporpmpt , is it brittle ?

Sources/RepoPrompt/Features/AgentMode/Views/Components/AgentOraclePill.swift
```diff
@@ -251,43 +345,41 @@ struct AgentOraclePill: View {
         #if DEBUG
             let _ = AgentModePerfDiagnostics.increment("ui.body.statusPills.oracle")
         #endif
-        Group {
+        let cornerRadius = AgentPillMetrics.cornerRadius()
+        Button {
             if hasAnySessions {
-                let cornerRadius = AgentPillMetrics.cornerRadius()
```
dsebban, line 256: make sure all the deletions are justified

Sources/RepoPrompt/Features/AgentMode/Views/ToolCards/ContextBuilderToolCards.swift
```diff
@@ -43,7 +45,7 @@ struct ContextBuilderCallCard: View {
     }
 
     private var detailLine: String? {
-        contextBuilderCardDetailLine(contextBuilderAgentVM: contextBuilderAgentVM)
+        contextBuilderCardDetailLine(contextBuilderAgentVM: contextBuilderAgentVM, dto: nil)
```
dsebban: dto: nil , weird signature , it seems like this param doesnt belong here , evalaute why it's here is it idiomatic ?

Sources/RepoPrompt/Features/Chat/Services/ChatHistoryManager.swift
```diff
@@ -300,19 +309,69 @@ actor ChatDataService {
...
+        // Apply the limit to logical conversations, not projection files. If any
+        // header is unreadable, retain everything: an unreadable projection may
+        // belong to a canonical group and must never be deleted independently.
         let limit = chatHistoryLimit
-        if limit != .unlimited, sortedFiles.count > limit.rawValue {
-            let filesToDelete = sortedFiles.dropFirst(limit.rawValue)
-            for url in filesToDelete {
-                // Best-effort delete; ignore individual failures
-                try? FileManager.default.removeItem(at: url)
+        guard limit != .unlimited else { return sortedFiles }
```
dsebban, line 475: guard seems not idiomatic in reporpompt code base evalaute

Sources/RepoPrompt/Features/Chat/Services/ChatHistoryManager.swift
```diff
@@ -300,19 +309,69 @@ actor ChatDataService {
...
+        var sessionsByURL: [URL: ChatSession] = [:]
```
dsebban, line 476: are vars used in repoprompt it seems non idiomatic evaluate

Sources/RepoPrompt/Features/Chat/Services/OracleGroupLanePayloadLoader.swift
```diff
@@ -0,0 +1,154 @@
+import Foundation
+import RepoPromptDomainRuntime
+
+enum OracleGroupLanePayloadLoader {
```
dsebban, line 4: evalaute the naming is it accurate , is this idiomatic in the codebase

Sources/RepoPrompt/Features/Chat/Services/OracleGroupLanePayloadLoader.swift
```diff
@@ -0,0 +1,154 @@
...
+            case .missingGroup:
+                "Canonical Oracle group was not found."
+            case .mismatchedProjection:
+                "The selected chat is not a member of its canonical Oracle group."
+            }
+        }
+    }
+
+    @MainActor
+    static func load(
```
dsebban, line 20: make sure thsi respect the concurrency guidlies of the codebased and does not incur performance issues

Sources/RepoPrompt/Features/Chat/Services/OracleHeadlessRuntime.swift
```diff
@@ -41,7 +41,13 @@ final class OracleHeadlessRuntime {
...
     private let cleanupConversation: CleanupConversation
     private let timeout: Duration
-    private var streamIDsByTabID: [UUID: ChatStreamID] = [:]
+    private var streamIDsByTabID: [UUID: Set<ChatStreamID>] = [:]
+    private var streamIDsByInvocationID: [UUID: Set<ChatStreamID>] = [:]
+    private var invocationIDsByTabID: [UUID: Set<UUID>] = [:]
+    private var tabIDByInvocationID: [UUID: UUID] = [:]
+    private var cancellationGenerationByTabID: [UUID: UInt64] = [:]
+    private var cancellationGenerationByInvocationID: [UUID: UInt64] = [:]
+    private var pendingStreamCountsByTabID: [UUID: Int] = [:]
```
dsebban: that's a lot of new added vars here is it jsutified ?

Sources/RepoPrompt/Features/Chat/Services/OracleLaneMarkdownFormatter.swift
```diff
@@ -0,0 +1,84 @@
...
+        let partialResponse: String?
+        let errorCode: String?
+        let errorMessage: String?
+    }
+
+    let lanes: [Lane]
+}
+
+enum OracleLaneMarkdownFormatter {
+    static let unspecifiedMetadata = "Provider default / not specified"
```
dsebban, line 30: not sure about the formatting maybe there is already code in the codebase that does that and this is not needed

Sources/RepoPrompt/Features/Chat/ViewModels/Oracle/OracleReviewPackagingDiagnostics.swift
```diff
@@ -4,6 +4,7 @@ import Foundation
 enum OracleSendOrigin: String {
     case askOracle
     case oracleSend
+    case contextBuilder
```
dsebban: weird this a PR about added oracles , why is contexxtbuilder added it  could it be removed safely ?

Sources/RepoPrompt/Features/Chat/ViewModels/Oracle/OracleViewModel+Groups.swift
```diff
@@ -0,0 +1,806 @@
...
+    }
+
+    private static func providerID(for model: AIModel) -> String {
+        if case let .customProvider(_, provider, _) = model {
+            return provider
+        }
+        return providerID(for: model.providerType)
+    }
+
+    private static func providerID(for provider: AIProviderType) -> String {
```
dsebban, line 64: what is going on here , why provider names , this is about adding extra oracles , tehre should not be any depdency to provider names etc , reevaluate this class if needed really

Sources/RepoPrompt/Features/Chat/ViewModels/Oracle/OracleViewModel+MCP.swift
```diff
@@ -920,7 +1001,8 @@ extension OracleViewModel {
         tabID: UUID? = nil,
         activateInUI: Bool = true,
         agentModeSessionID: UUID? = nil,
-        agentModeRunID: UUID? = nil
+        agentModeRunID: UUID? = nil,
+        implicitSessionID: UUID? = nil
```
dsebban, line 1018: what implicit session id ? did you make it overcomplicated ?

Sources/RepoPrompt/Features/Chat/ViewModels/Oracle/OracleViewModel+MCP.swift
```diff
@@ -971,58 +1053,22 @@ extension OracleViewModel {
             return existing.id
         }
 
-        func eligible(_ session: ChatSession, allowUnownedLegacy: Bool = true) -> Bool {
```
dsebban, line 974: that's a lot of deletion is it justiifed ?

Sources/RepoPrompt/Features/Chat/ViewModels/Oracle/OracleViewModel+MCP.swift
```diff
@@ -1055,7 +1101,10 @@ extension OracleViewModel {
     func tool_chatSend(
         args: [String: Value],
         promptVM: PromptViewModel,
-        tabContext: OracleSendTabContext? = nil
+        tabContext: OracleSendTabContext? = nil,
+        resolvedModel: AIModel? = nil,
+        implicitSessionID: UUID? = nil,
+        onProgress: ((_ text: String, _ reasoning: String?) -> Void)? = nil
```
dsebban, line 1120: that doesnt seem idiomatic in reporpmpt reevaluate if you can remove

Sources/RepoPrompt/Features/Chat/ViewModels/Oracle/OracleViewModel+MCP.swift
```diff
@@ -1128,26 +1186,16 @@ extension OracleViewModel {
...
-            let activeSessionID = workspaceManager.activeChatSessionID(forTabID: tabContext.tabID)
-                ?? currentSessionID.flatMap { currentID in
-                    sessions.first(where: { $0.id == currentID && $0.composeTabID == tabContext.tabID })?.id
-                }
-            let isUserStreaming = isSessionStreaming(activeSessionID)
-            shouldActivate = isFocusedTab && !isUserStreaming
-        } else {
-            shouldActivate = true
-        }
+        let shouldActivate = shouldActivateOracleSendSession(tabContext: tabContext, promptVM: promptVM)
```
dsebban, line 1202: why did oyu inline , prefer keeping original instead of inilining

Sources/RepoPrompt/Features/Chat/ViewModels/Oracle/OracleViewModel.swift
```diff
@@ -3627,9 +3665,7 @@ class OracleViewModel: ObservableObject {
     private func handleComposeTabsWillClose(_ tabIDs: Set<UUID>) async {
         for tabID in tabIDs {
             // 1. Cancel headless stream (plan/question generation) for this tab
-            if headlessRuntime.hasActiveStream(for: tabID) {
```
dsebban, line 3663: weird deletion , are you sure ?

Sources/RepoPrompt/Features/ContextBuilder/ViewModels/ContextBuilderAgentViewModel.swift
```diff
@@ -470,6 +486,9 @@ final class ContextBuilderAgentViewModel: ObservableObject {
                 afterCommittedTabSnapshotCaptured: (@MainActor @Sendable (
                     _ runID: UUID,
                     _ snapshot: MCPServerViewModel.ContextBuilderCommittedTabSnapshot
+                ) async -> Void)? = nil,
```
dsebban, line 487: what is this signature , that doesnt look idiomatic

Sources/RepoPrompt/Features/ContextBuilder/ViewModels/ContextBuilderAgentViewModel.swift
```diff
@@ -4404,6 +4463,329 @@ final class ContextBuilderAgentViewModel: ObservableObject {
         )
     }
 
+    private func cancelAndDrainOracleGroup(
```
dsebban, line 4569: that's a huge function si it needed to eb so complicated reevaluate it

Sources/RepoPrompt/Features/ContextBuilder/ViewModels/ContextBuilderRunLifecycle.swift
```diff
@@ -188,6 +188,7 @@ struct ContextBuilderMCPRunConfiguration {
     let runBehavior: ContextBuilderRunBehavior
     let responseType: String?
     let planningModelRaw: String?
+    let agentModelsProfile: AgentModelsSettingsProfile
```
dsebban, line 191: what's taht , is this related to the multiple oracle it talsk about profiles wierd

Sources/RepoPrompt/Infrastructure/AI/ACP/ACPAgentSessionController.swift
```diff
@@ -239,6 +239,7 @@ actor ACPAgentSessionController {
     private var stdoutConsumerTask: Task<Void, Never>?
     private var stderrConsumerTask: Task<Void, Never>?
     private var processWaitTask: Task<Void, Never>?
+    private var isDrainingTransport = false
```
dsebban: boolean flag liek this is suspsicious why is this needed for the multiple oracle feature

Sources/RepoPrompt/Infrastructure/AI/ACP/ACPAgentSessionController.swift
```diff
@@ -1219,6 +1269,39 @@ actor ACPAgentSessionController {
...
+                await processWaitTask.value
+                await stdoutConsumerTask.value
+                await stderrConsumerTask.value
+                return true
+            }
+            group.addTask {
+                do {
+                    try await Task.sleep(nanoseconds: UInt64(timeout * 1_000_000_000))
+                } catch {
+                    return false
```
dsebban: ???

Sources/RepoPrompt/Infrastructure/AI/ACP/ACPAgentSessionController.swift
```diff
@@ -1219,6 +1269,39 @@ actor ACPAgentSessionController {
...
+                }
+                return false
```
dsebban: ???

Sources/RepoPrompt/Infrastructure/AI/ACP/ACPAgentSessionController.swift
```diff
@@ -1284,22 +1367,26 @@ actor ACPAgentSessionController {
...
-            else { return }
-            let text = Self.strippingANSIEscapeSequences(from: rawText)
-            guard !text.isEmpty else { return }
-            stderrLineCount += 1
-            lastStderrPreview = Self.truncatedDiagnosticPreview(text)
-            recordActivePromptStderrLine(text)
-            diagnose(.stderrLine(text))
-            guard provider.shouldEmitStderrLine(text) else { return }
-            emit(.stream(AIStreamResult(type: "system", text: text)))
+            handleStderrLine(lineData)
```
dsebban: deleteing all of that to inline is not needed i prefer keeping the original code to avoid additional changes , keep this in mind , blin inlining is not prefered against keep original code

Sources/RepoPrompt/Infrastructure/AI/Providers/GrokBuild/GrokBuildACPHeadlessAgentProvider.swift
```diff
@@ -67,7 +67,8 @@ final class GrokBuildACPHeadlessAgentProvider: HeadlessAgentProvider {
             beforePrompt: { controller, _ in
                 try await Self.applySelectedModelIfNeeded(config: config, controller: controller)
             },
-            approvalPolicy: .declineUnsupported
+            approvalPolicy: .declineUnsupported,
```
dsebban: not sure why grok files are changed her eit should only be about oracles

Sources/RepoPrompt/Infrastructure/AI/Providers/ACPHeadlessAgentProviderBridge.swift
```diff
@@ -12,6 +12,11 @@ final class ACPHeadlessAgentProviderBridge: HeadlessAgentProvider {
         case acceptForSession
     }
 
+    enum PromptCompletion {
+        case shutdownImmediately
```
dsebban: I dont undersatnd why acp related stuff are inside this PR it might be left overs from another pr

Sources/RepoPromptDomainRuntime/MCPDomainStandaloneCapabilityProvider.swift
```diff
@@ -274,7 +274,37 @@ package enum MCPDomainStandaloneToolInstaller {
             binding(MCPWindowToolName.fileActions, backends.filesystem.manageFiles),
             binding(MCPWindowToolName.applyEdits, backends.filesystem.applyFileEdits),
             binding(MCPWindowToolName.oracleUtils, backends.conversation.accessOracleUtilities),
-            binding(MCPWindowToolName.askOracle, backends.conversation.startOracleConversation),
+            binding(MCPWindowToolName.askOracle) { request in
+                let args = try request.mcpArguments()
```
dsebban, line 278: what haeeped here why so much adde dnew code

Sources/RepoPromptMCP/CommandRunner/MCPCommandRunner.swift
```diff
@@ -599,10 +612,12 @@ actor MCPCommandRunner {
...
           oracle_send                    Send oracle conversation turn
             message      (required)      The message to send
-            new_chat     (bool)          true=start new, false=continue (important!)
+            chat_id      (string)        Continue an exact conversation
+            new_chat     (bool)          true=start new; omitted resumes selected/recent
+            model        (string)        Primary override for explicit new_chat=true only
             mode         (string)        chat|plan|review
               review mode includes git diffs when published via git artifacts
-            Example: oracle_send message="Review this" mode=review new_chat=true
+          Example: oracle_send message="Review this" mode=review new_chat=true
```
dsebban: yhis change is not needed right ? its a linting issue

Sources/RepoPromptMCP/CommandRunner/MCPCommandRunner.swift
```diff
@@ -599,10 +612,12 @@ actor MCPCommandRunner {
         Conversation:
           oracle_send                    Send oracle conversation turn
             message      (required)      The message to send
-            new_chat     (bool)          true=start new, false=continue (important!)
```
dsebban, line 602: why did you change that ?

docs/architecture/oracle-groups-rewrite.md
dsebban: Agreed that the working plan should stay out of the PR. The file on this line is the tracked architecture document (`oracle-groups-rewrite.md`), not the local `oracle-groups-clean-runtime-plan.md`; that local plan remains untracked and excluded. The PR description carries the implementation and validation summary instead.

Sources/RepoPrompt/App/AppDomainRuntimeComposition.swift
dsebban: The local `runtime` is needed during initialization: the same newly constructed instance is assigned to `self.runtime` and passed to the conversation store and claim manager. Using `self.runtime` for those dependent initializers before every stored `let` is initialized would violate Swift initialization rules, so the local binding is the smallest composition-root form.

Sources/RepoPrompt/App/AppDomainRuntimeComposition.swift
dsebban: Yes. This is the normal initialization order for dependent immutable properties: construct the runtime, assign it once to `self.runtime`, then build the stores from that exact instance. It avoids a second runtime authority and keeps all four app-wide dependencies immutable.

Sources/RepoPrompt/App/AppDomainRuntimeComposition.swift
dsebban: These properties belong in the app composition root because they are app-wide domain authorities built from the same `MCPDomainRuntime` persistence coordinator and identity: conversation store, claim manager, and grouped runtime. Keeping them together as immutable properties prevents callers from constructing divergent stores or claim authorities; the app and Context Builder integration tests exercise these shared instances.

Sources/RepoPrompt/Features/AgentMode/ViewModels/UI/AgentModelsSettingsViewModel.swift
dsebban: The `1` is the primary Oracle. `additionalOracleModelRaws` intentionally contains only the extra lanes, so the total roster size is `1 + additional.count`. `OracleRosterContract` defines the one-to-five invariant, and `AgentOraclePillRoutingTests.testAggregateOraclePillCountAndLabelsUseConfiguredOrLatestProjectedGroup` covers zero additional models as one Oracle and the five-lane cap.

Sources/RepoPrompt/Features/AgentMode/ViewModels/UI/AgentModelsSettingsViewModel.swift
dsebban: This is an immutable computed Boolean, consistent with the other computed Booleans in this view model (`hasWorkspace`, `isEditingWorkspaceSettings`, `hasConnectedCLIProvider`, etc.). It derives whether Add is currently valid from two invariants: a configured primary model and a roster below `OracleRosterContract.maximumCount`; storing separate mutable state would risk drift.

Sources/RepoPrompt/Features/AgentMode/Views/Components/AgentOraclePill.swift
dsebban: The guards are deliberate early exits in the pure selection helper: no eligible session returns `nil`, and a non-group session returns unchanged. For a group, the helper canonicalizes the selection to lane 0. `AgentOraclePillRoutingTests.testLatestStreamingSessionDoesNotFallbackToStaleCompletedSession` and the grouped continuation tests cover those routing boundaries.

Sources/RepoPrompt/Features/AgentMode/Views/Components/AgentOraclePill.swift
dsebban: Returning `nil` here preserves the function’s existing optional contract when there are no sessions; it does not manufacture a fallback session. The next guard handles the distinct non-group case. This separation is what lets callers hide the pill cleanly when no eligible session exists.

Sources/RepoPrompt/Features/AgentMode/Views/Components/AgentOraclePill.swift
dsebban: The clamping is enforcing the persisted contract, not adding general flexibility: a roster is always 1...5. The function must also reconcile current settings with an already-projected group, so it takes the larger valid count without allowing corrupted projection data or an invalid configured count outside that range. `testAggregateOraclePillCountAndLabelsUseConfiguredOrLatestProjectedGroup` explicitly covers projected values, an invalid projected size, and `configuredAdditionalCount: 99`.

Sources/RepoPrompt/Features/AgentMode/Views/Components/AgentOraclePill.swift
dsebban: The two inputs represent different authorities at different times: current configuration describes the next run, while the latest persisted group describes what is already on screen. The function selects only the latest group, derives its maximum recorded size, and returns the larger contract-bounded count so changing settings does not relabel an existing result incorrectly. The aggregate-count test covers both sources and stale history.

Sources/RepoPrompt/Features/AgentMode/Views/Components/AgentOraclePill.swift
dsebban: The durable group result is structured; this text check is only a presentation fallback for live/legacy `ChatSession` projections, which currently expose the assistant transcript but no terminal-status enum. It recognizes the exact existing error frame or an `Error:` prefix rather than arbitrary occurrences. `testAggregateOraclePillCountAndLabelsUseConfiguredOrLatestProjectedGroup` verifies framed/prefixed failures and that normal prose mentioning error handling remains completed. A bounded follow-up is to remove this fallback once terminal status is persisted on the projection itself.

Sources/RepoPrompt/Features/AgentMode/Views/Components/AgentOraclePill.swift
dsebban: Addressed in `31f0d374eaca9d37c6197c96a7fdaf00760374ed`: the empty-state roster configuration state and popover were removed, and the pill again renders nothing when there is no session; it only opens existing Oracle chats. Group lane navigation and Copy All remain. `AgentOraclePillRoutingTests.testExplicitRequestStateRejectsBlankStaleTabAndMismatchedSession` protects the no/stale-session routing boundary, while the aggregate-count test covers the retained group UI logic.

Sources/RepoPrompt/Features/AgentMode/Views/ToolCards/ContextBuilderToolCards.swift
dsebban: The call card has no result DTO yet; the completed result card does. Both intentionally use one formatter, whose optional DTO prefers durable per-lane result metadata when present and otherwise falls back to the view model’s frozen `mcpPlanModel`. `ContextBuilderOracleResultTests.testGroupReplyAndToolCardPreserveOrderedUnsynthesizedLanes` verifies the DTO path and the joined fallback model line, so `nil` here represents the pre-result phase rather than an unused parameter.

Sources/RepoPrompt/Features/Chat/Services/ChatHistoryManager.swift
dsebban: The guard is the fail-fast form of the important retention invariant: an unlimited history returns unchanged, and any unreadable header causes all files to be retained so one projection of a canonical group is never deleted independently. Applying the limit to logical conversations rather than files is the behavioral reason for this structure.

Sources/RepoPrompt/Features/Chat/Services/ChatHistoryManager.swift
dsebban: These are local mutable accumulators inside the `ChatDataService` actor: `sessionsByURL` is populated while decoding headers, and `orderedKeys` records first-seen logical conversations in modification order. Neither escapes the function, and actor isolation serializes access; immutable declarations would require repeated decoding or a less readable reduction without changing the invariant.

Sources/RepoPrompt/Features/Chat/Services/OracleGroupLanePayloadLoader.swift
dsebban: The name matches its bounded responsibility: it loads one canonical Oracle group and produces the per-lane payload consumed by `OracleLaneMarkdownFormatter`. It does not own execution, transport, or persistence. `OracleLaneMarkdownFormatterTests` exercises terminal durable results, prepared live projections, unavailable lanes, ordering, and exact execution metadata through this loader.

Sources/RepoPrompt/Features/Chat/Services/OracleGroupLanePayloadLoader.swift
dsebban: The concurrency boundary is intentional: `@MainActor` is required while reading `ChatSession` and live-message projections. The only suspension is one canonical store load; payload construction then uses bounded linear maps, with the roster itself capped at five lanes, and it starts no child tasks or per-lane I/O. `testTerminalPayloadUsesDurableResultsAndExactExecutionProfile` and `testPreparedPayloadUsesCurrentPartialTextAndExplicitRunningFailedUnavailableStates` cover both sides of that boundary.

Sources/RepoPrompt/Features/Chat/Services/OracleHeadlessRuntime.swift
dsebban: The state is needed because grouped execution changes the old one-stream-per-tab relationship into concurrent streams with invocation-scoped cancellation. The sets model the many-to-many registrations; the generation counters close the race where cancellation occurs while `sendPrompt` is still pending; and the pending count keeps `cancelAll` aware of multiple pending sends on one tab. `OracleHeadlessRuntimeTests` covers concurrent same-tab cancellation, pending-send cancellation, successor preservation, and concurrent pending `cancelAll`.

Sources/RepoPrompt/Features/Chat/Services/OracleLaneMarkdownFormatter.swift
dsebban: I checked for an existing formatter, but the required contract is specific: canonical lane order and labels, explicit running/completed/failed/cancelled/unavailable states, provider/model/effort metadata, exact response whitespace, partial output, and structured errors. Keeping that in this 84-line formatter avoids overloading generic Markdown helpers. `OracleLaneMarkdownFormatterTests` asserts those exact fields, ordering, wording, and whitespace preservation.

Sources/RepoPrompt/Features/Chat/ViewModels/Oracle/OracleReviewPackagingDiagnostics.swift
dsebban: `contextBuilder` is only an origin tag in the DEBUG packaging diagnostics. Context Builder can request a plan/question/review follow-up through the same Oracle send path, so omitting the case would misattribute that path as compatibility or direct MCP. It adds no Context Builder dependency to grouped execution and is used at the `OracleSendTabContext` creation site for the follow-up.

Sources/RepoPrompt/Features/Chat/ViewModels/Oracle/OracleViewModel+Groups.swift
dsebban: Provider identity is part of selecting and durably recording each lane, not a transport dependency. A model name alone is insufficient for custom providers and can be ambiguous; this adapter converts the already-resolved `AIModel` into the domain `OracleExecutionProfile` so the result records what actually ran. Dispatch remains through the existing model/provider services. `OracleHeadlessRuntimeTests.testResolvedExecutionProfileCapturesRuntimeModelAndEffectiveEffort` covers Codex, Anthropic, and a custom provider.

Sources/RepoPrompt/Features/Chat/ViewModels/Oracle/OracleViewModel+MCP.swift
dsebban: `implicitSessionID` is a frozen UUID chosen once before asynchronous group preparation. Without it, the later send could re-run “selected/recent” resolution after active-session changes or a short-ID collision and continue a different chat. `AgentOraclePillRoutingTests.testFrozenImplicitSessionIdentitySurvivesShortIDCollision` and `testFrozenImplicitSessionIdentityPreservesSessionOwnerMatchWithoutExplicitRejection` cover that exact invariant.

Sources/RepoPrompt/Features/Chat/ViewModels/Oracle/OracleViewModel+MCP.swift
dsebban: The continuation-selection behavior was moved, not removed. It now lives in `resolveImplicitOracleContinuationCandidate`, which preserves tab scoping, owner buckets, active-session preference, UI activation behavior, and most-recent fallback. Both group preselection and the actual send use the same helper, avoiding a time-of-check/time-of-use mismatch. The legacy active-then-recent and frozen-session tests cover this path.

Sources/RepoPrompt/Features/Chat/ViewModels/Oracle/OracleViewModel+MCP.swift
dsebban: These are internal dependency-injection points for reusing the established single-send pipeline per group lane: `resolvedModel` prevents a second model lookup, `implicitSessionID` freezes the chosen projection, and `onProgress` publishes lane progress. They are not public MCP arguments. The execution-profile, frozen-session, and grouped Context Builder progress/finalization tests cover the three responsibilities.

Sources/RepoPrompt/Features/Chat/ViewModels/Oracle/OracleViewModel+MCP.swift
dsebban: This change extracts the original activation block rather than inlining it. `shouldActivateOracleSendSession` keeps the existing focused-tab and non-streaming checks, adds the explicit activation policy, and is reused both when preselecting a continuation and when sending it. Keeping one helper prevents the two group-routing stages from drifting.

Sources/RepoPrompt/Features/Chat/ViewModels/Oracle/OracleViewModel.swift
dsebban: The precheck was removed because it was racy: while `sendPrompt` is suspended, no stream ID is registered yet, so `hasActiveStream` is false even though cancellation must fence the eventual stream. Calling cancel unconditionally increments the tab generation; a later returned stream is then cancelled before registration. `OracleHeadlessRuntimeTests.testCancellationWhileSendPromptIsPendingCancelsReturnedStream` covers this exact case.

Sources/RepoPrompt/Features/ContextBuilder/ViewModels/ContextBuilderAgentViewModel.swift
dsebban: This signature exists only under `#if DEBUG` as a test hook, not as product API. `@MainActor @Sendable` matches the state it observes, and `async` lets tests suspend exactly after the committed tab snapshot to reproduce post-commit cancellation races deterministically. `ContextBuilderMCPProgressTimelineTests.testPostCommitMCPCancellationReturnsCommittedPromptAndSelection` and `testPreCommitMCPCancellationRendersInitialSnapshotWithoutIngressBarrier` exercise that seam.

Sources/RepoPrompt/Features/ContextBuilder/ViewModels/ContextBuilderAgentViewModel.swift
dsebban: The highlighted `cancelAndDrainOracleGroup` helper is 22 lines; the large diff hunk also includes the separate grouped follow-up runner. The helper deliberately keeps one ordered cleanup sequence: invalidate ownership, cancel every member, await the group task, then clear state only if the same generation still owns cleanup. `testReplacementOracleRunOwnsStateWhilePredecessorResumesAfterReservationRelease` and the parent-cancellation tests protect that ordering. A bounded follow-up can extract pure packaging/projection helpers from the runner, but not split the transactional ordering without equivalent tests.

Sources/RepoPrompt/Features/ContextBuilder/ViewModels/ContextBuilderRunLifecycle.swift
dsebban: The profile is captured because a Context Builder run may suspend before its Oracle follow-up starts, while settings can change meanwhile. `planningModelRaw` alone is insufficient: the roster also includes ordered additional Oracle models and workspace inheritance. `ContextBuilderStrictFinalizationTests.testMCPGroupedFollowUpFreezesProfileBeforeSuspendedLaunch` verifies the captured primary and additional roster are used, and `testMCPFollowUpWithoutCapturedProfileFailsClosed` verifies missing capture does not fall back to live settings.

Sources/RepoPrompt/Infrastructure/AI/ACP/ACPAgentSessionController.swift
dsebban: Addressed in `31f0d374eaca9d37c6197c96a7fdaf00760374ed`: the unrelated ACP transport-drain cluster, including `isDrainingTransport`, was removed from this PR. That scope decision is not a judgment that transport draining is invalid; it should be evaluated separately with its own failure model and validation. The remaining Grok bridge tests still cover status/content separation and clean pre-terminal completion.

Sources/RepoPrompt/Infrastructure/AI/ACP/ACPAgentSessionController.swift
dsebban: Addressed in `31f0d374eaca9d37c6197c96a7fdaf00760374ed`: the added transport-settlement task group and timeout path were removed with the unrelated ACP/Grok transport cluster. The transport idea is not being rejected; it is simply no longer coupled to the Oracle-group PR. Current head restores the pre-existing shutdown path, while `GrokBuildACPHeadlessAgentProviderTests` retains the in-scope stream-completion coverage.

Sources/RepoPrompt/Infrastructure/AI/ACP/ACPAgentSessionController.swift
dsebban: Addressed in `31f0d374eaca9d37c6197c96a7fdaf00760374ed`: this cooperative-settlement timeout branch was removed together with the unrelated transport-drain work. That keeps this PR focused on grouped Oracle lifecycle; a transport change can be reviewed independently rather than inferred to be invalid. The current bridge tests continue to verify clean terminal delivery and short content handling.

Sources/RepoPrompt/Infrastructure/AI/ACP/ACPAgentSessionController.swift
dsebban: Addressed in `31f0d374eaca9d37c6197c96a7fdaf00760374ed`: the off-scope stderr helper refactor was removed and the prior direct `handleStderrChunk` logic restored. This follows the preference to avoid unrelated movement in this PR. It does not reject a future transport cleanup; it leaves that work for a focused change with dedicated tests.

Sources/RepoPrompt/Infrastructure/AI/Providers/GrokBuild/GrokBuildACPHeadlessAgentProvider.swift
dsebban: Addressed in `31f0d374eaca9d37c6197c96a7fdaf00760374ed`: the Grok-specific `promptCompletion: .drainTransportBeforeShutdown` change was removed, along with its off-scope drain scenarios/tests. The unrelated transport cluster was removed from this PR, not declared invalid. The retained tests still cover default/dynamic model selection, failure before prompt, status versus content, and clean completion.

Sources/RepoPrompt/Infrastructure/AI/Providers/ACPHeadlessAgentProviderBridge.swift
dsebban: Addressed in `31f0d374eaca9d37c6197c96a7fdaf00760374ed`: `PromptCompletion`, the transport-settlement actor, terminal rewriting, and drain-specific controller calls were all removed from `ACPHeadlessAgentProviderBridge`. They were unrelated transport work rather than a requirement of grouped Oracles. This scope correction does not invalidate the transport idea; it separates it for independent design and validation.

Sources/RepoPromptDomainRuntime/MCPDomainStandaloneCapabilityProvider.swift
dsebban: The standalone binding previously hardwired every `ask_oracle` call to “start,” so it could not honor `chat_id`, `new_chat`, or a model override consistently with the app binding. The added code validates argument types, delegates route semantics to shared `OracleConversationRoute`, and dispatches only start versus continuation. Commit `31f0d374eaca9d37c6197c96a7fdaf00760374ed` added `MCPDomainStandaloneCompositionTests.testInstalledAskOracleBindingRoutesAndRejectsMalformedArguments`, covering three valid routes and five malformed cases without backend mutation.

Sources/RepoPromptMCP/CommandRunner/MCPCommandRunner.swift
dsebban: This is help text rather than a lint-only change. It documents the now-supported `chat_id` and `model` parameters and corrects `new_chat` semantics; the example is aligned as a command-level example rather than another argument. The underlying behavior is covered at current head by `testInstalledAskOracleBindingRoutesAndRejectsMalformedArguments`.

Sources/RepoPromptMCP/CommandRunner/MCPCommandRunner.swift
dsebban: The previous help said `new_chat=false` meant continue, but the actual contract distinguishes exact continuation by `chat_id` from omitted `new_chat`, which resumes the selected/recent conversation. The text was updated so CLI users and agents do not accidentally start or continue the wrong chat, and it also documents that model override is valid only for an explicit new conversation. The standalone routing test added in `31f0d374eaca9d37c6197c96a7fdaf00760374ed` locks those semantics down.

dsebban:
The local working plan `docs/architecture/oracle-groups-clean-runtime-plan.md` remains excluded from this PR; the PR description carries the implementation summary instead.

I re-reviewed the complete 76-file PR diff with a narrow YAGNI / blast-radius lens. I did not find a reason to broaden the architecture or refactor unrelated code. One concrete Context Builder ownership race remained: `runFollowUpOracleGroup` discarded `cancelAndDrainOracleGroup`’s ownership result, so an older replacement request could resume after a newer run took ownership, call `beginRun()`, and clear the newer answer/route. A request already cancelled before startup could also start draining current work before observing cancellation.

Fixed in `629ef4d43e5a11fc018058802a60ca9ece923a12` with the narrow startup fence: check cancellation before draining and require cleanup ownership before beginning a new generation. The existing Context Builder lifecycle test now covers both superseded cleanup and already-cancelled startup. No persistence, schema, provider, routing, or Oracle group contract behavior was changed.

The commit is limited to the Context Builder view model and its existing lifecycle test. The GitHub connector reconstructs full blobs rather than applying hunks, so the final two-file diff also contains two comment-only wording deltas in the same production source file; they do not alter executable behavior, APIs, schemas, or test semantics. I intentionally stopped further whole-file reconstruction once the remaining differences were comment-only to avoid introducing higher-risk source churn.

Validation for this fix: an isolated Swift 6.2.1 ownership harness reproduces the failure before the fix and passes after it; patch whitespace/apply/reverse checks passed against the reviewed source. I could not run the repository’s macOS `make dev-test FILTER=ContextBuilderOracleGroupStateTests`, SwiftFormat/SwiftLint, or mandatory contribution preflight from this connector-only environment, so those remain CI/local follow-up. Previously reported broader PR validation at `31f0d374...` is historical and should not be read as validation of the new head.

Current head: `629ef4d43e5a11fc018058802a60ca9ece923a12`.

Sources/RepoPromptMCP/DirectHeadlessOracleAdapter.swift
```diff
@@ -0,0 +1,603 @@
...
+            if let group = try await store.load(member: OracleMemberLookup(publicChatID: chatID), owner: owner) {
+                return try await groupContinuationPlan(
+                    group: group,
+                    configuredRoster: configured,
+                    invocationID: invocationID,
+                    runID: runID,
+                    input: input
+                )
+            }
+            if configured.count == 1 {
```
dsebban: **[P1] Keep named direct chats on the direct path after grouped settings are enabled**

This makes an explicit non-group `chat_id` routable only while `configured.count == 1`. A direct chat can legitimately exist while the current roster is `N > 1`: message-only `oracle_send` deliberately uses the primary direct fallback, and an existing `N = 1` chat also survives a later roster settings change. In either case, the next explicit continuation misses the durable group store and throws `unknownChatID`, contrary to this PR's stated contract that a named single chat remains on the existing single path and is never promoted.

The minimal fix is to make the group-store miss unconditionally fall back to a one-lane direct plan:

```swift
let directRoster = try OracleRoster(primary: configured.primary)
try await provider.validateOracleRoster(directRoster)
let childPlan = try Self.childPlan(runID: runID, roster: directRoster)
return InvocationPlan(
    invocationID: invocationID,
    runID: runID,
    claimID: nil,
    input: input,
    route: .direct(modelID: directRoster.primary.modelID, implicitConversationID: nil),
    childLaunchPlan: childPlan
)
```

No new lookup API or promotion logic is needed. `DirectHeadlessConversationBackend` already parses the UUID, and `DirectHeadlessProviderCoordinator.continueConversation` is already the authority that rejects unknown IDs and resumes known direct chats with their stored provider/model.

Add one focused regression to `DirectHeadlessOracleGroupTests`: create a direct chat under an `N > 1` roster (or create it at `N = 1` and then enable an additional model), explicitly continue the returned `chat_id`, and assert the same `chat_id`, no `oracle_group_id`, exactly one lane/provider dispatch, and no durable group document. The current single-direct and implicit-direct tests cover each half separately, but not this transition.

dsebban (commented):
Reviewed the full 74-file diff at `ffac9684facd8f74e370fa1dcfa876cbeb1315f6`, including settings/schema authority, app routing and projections, Context Builder integration, the shared contracts/runtime/claims/persistence/terminal-publication lifecycle, direct-headless carrier/token routing, and the focused test suites.

Validation at this exact SHA is green across all eight hosted CI jobs: the four root build-and-test shards, Provider Tests, the Sentry-enabled build/privacy check, Style/guardrails, and Secret Scan.

**Verdict: one blocking compatibility defect.** I intentionally omitted style-only and speculative/YAGNI comments; the rest of the grouped lifecycle is coherent in the execution and recovery paths I traced. I also did not carry forward the automated prepared-plan retention warning: grouped execution removes the plan in `consumePlan` before provider execution, while authorization, preparation, cancellation, and binding-failure paths revoke/discard prepared state.

The inline finding gives the smallest fix and the focused regression test needed to validate it. GitHub does not permit the PR author's account to submit a formal “Request changes” review on its own PR, so this is submitted as a blocking comment review.

Sources/RepoPromptMCP/DirectHeadlessOracleAdapter.swift
dsebban: Fixed in `4f1b4b2a5831a62a773b7fb7fa90685154662a34`. A named `chat_id` that is not a canonical group member now always plans the existing one-lane direct path from the configured primary, regardless of additional roster entries. `testNamedDirectContinuationStaysSingleLaneWithGroupedSettings` covers the exact N>1-settings transition and asserts there is no group or lane identity on the launch plan.

Sources/RepoPrompt/Features/ContextBuilder/ViewModels/ContextBuilderAgentViewModel.swift
```diff
@@ -4404,6 +4460,328 @@ final class ContextBuilderAgentViewModel: ObservableObject {
...
+        session.isBackgroundPlanGenerating = true
+        session.backgroundPlanError = nil
+        session.backgroundPlanResponseText = nil
+        session.backgroundPlanReasoningText = nil
+        session.followUpOracleSessionID = nil
+        updateRuntimeBindings(from: session)
+
+        await progressReporter?(.payloadPackaging)
+        try requireCurrentOracleRun(session: session, generation: generation)
+        let aiMessage = try await promptManager.buildHeadlessAIMessage(
```
dsebban: **[P2] Run pre-launch packaging through the same generation-owned cleanup path**

`isBackgroundPlanGenerating` is set before this call, but the existing `do/catch` starts only after the frozen pack, packaging context, callbacks, and task have all been created. A failure from `buildHeadlessAIMessage`, `ContextBuilderFrozenOraclePack.make`, or either prelaunch generation check therefore escapes without clearing the generating flag, invalidating the group generation, clearing previews/routes, or updating runtime bindings. The MCP caller propagates that error directly, so this tab can remain visibly stuck in a generating state even though no group task exists.

Keep the current cleanup ordering, but widen its ownership boundary: start the `do` immediately after the initial state setup, track the frozen pack/reservation as optional state, and route every pre-launch and post-launch error through the existing generation check + `cancelAndDrainOracleGroup` + UI reset. Release an acquired reservation only when one exists and was not durably retained.

Add one focused regression that forces packaging to throw before `followUpOracleGroupTask` is installed and asserts: `isBackgroundPlanGenerating == false`, no generated route, no group task, the generation is invalidated/finished, and no artifact reservation remains.

Sources/RepoPrompt/Infrastructure/MCP/ToolOutputFormatter.swift
```diff
@@ -4372,6 +4382,62 @@ extension ToolOutputFormatter {
...
+        }) else { return nil }
+
+        var lines = [heading]
+        if let status = dto.status {
+            lines.append("- Oracle group status: \(status)")
+        }
+        if let groupID = dto.oracleGroupID {
+            lines.append("- Oracle group: `\(groupID)`")
+        }
+        let payload = OracleLaneMarkdownPayload(lanes: ordered.map { lane in
```
dsebban, line 4408: **[P2] Preserve the structural cancelled lane status in formatted MCP output**

This collapses `.cancelled` into `.failed`, even though the shared result contract and `OracleLaneMarkdownPayload.Status` both distinguish them and the PR explicitly promises that cancellation remains visible as a structural lane outcome. A correctly persisted cancelled lane is therefore shown to agents/users as `Status: Failed`.

The minimal fix is only to split this switch arm:

```swift
case .failed: .failed
case .cancelled: .cancelled
```

Add a formatter regression with a grouped result containing one cancelled lane and assert the rendered lane says `- Status: Cancelled` rather than `Failed`. No DTO or domain-contract change is needed.

dsebban (commented):
Completed five independent review passes against the full PR and pushed the three bounded headless fixes in `4f1b4b2a5831a62a773b7fb7fa90685154662a34`.

1. **Compatibility/routing:** fixed named direct-chat continuation when additional Oracle models are configured.
2. **Concurrency/cancellation:** replaced the child bridge's ineffective blocking-write deadline with the existing bounded nonblocking writer.
3. **Durability/recovery:** re-audited claims, prepared-turn recovery, CAS, journals, artifact references, and exact terminal publication; no additional material defect found. The automated prepared-plan retention warning is not valid because grouped plans are removed by `consumePlan`, while pre-dispatch failures revoke/discard them.
4. **Authorization/security:** restricted grouped headless child MCP connections to a strict read-only allowlist (`get_code_structure`, `get_file_tree`, `read_file`, `file_search`, `git`), with future tools denied by default.
5. **Product/API/presentation:** found the two remaining material defects below.

The pushed commit includes focused regressions for all three repaired paths. The new hosted CI run is active. GitHub still prevents the author account from issuing a formal Request Changes review on its own PR, so the two remaining blockers are submitted as a comment review.

Sources/RepoPrompt/Infrastructure/MCP/ToolOutputFormatter.swift
dsebban: Fixed in `3f80bebb73adca21637cf3c7c019a900cbdb0d14`. The shared lane formatter now canonicalizes a projected `.failed` status back to `.cancelled` when the structural lane error code is `cancelled`, so every formatting surface renders the durable outcome correctly. `testFormatterRecoversCanonicalCancelledStatusFromErrorCode` asserts `Status: Cancelled` and rejects `Status: Failed`.

Sources/RepoPromptMCP/DirectHeadlessOracleAdapter.swift
```diff
@@ -0,0 +1,603 @@
...
+        }
+        guard plan.claimID != nil else {
+            plansByInvocationID.removeValue(forKey: invocationID)
+            guard case let .direct(modelID, implicitConversationID) = plan.route else {
+                throw AdapterError.missingPreparedInvocation
+            }
+            return .direct(modelID: modelID, implicitConversationID: implicitConversationID)
+        }
+        return .group
+    }
```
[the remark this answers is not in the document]
dsebban: Verified as not a defect. For a grouped route, the backend immediately enters `start`, `continue`, or `buildContext`; each calls `consumePlan`, which removes the invocation entry before `execute` reaches any provider/network suspension. Failures before binding execution are covered by the long-running wrapper's `revokeChildLaunches`, which calls `discardPreparedInvocation`, and shutdown clears any remaining preparations. A provider error therefore cannot retain this dictionary entry.

Sources/RepoPrompt/Features/ContextBuilder/ViewModels/ContextBuilderAgentViewModel.swift
dsebban: Fixed in `ce01d007e27b9c93159f625b23d0b2a271d459d1` without restructuring the large coordinator. `MCPAppPhysicalCapabilityAdapters.Execution` now wraps the injected grouped follow-up call, captures the exact pre-run session and generation, and cleans only when an N>1 invocation fails before installing its group task. It invalidates that owned generation, clears stale route/preview state, records the error (or clears it for cancellation), and reuses `setBackgroundPlanGenerating(false)` for preview/runtime binding refresh. A newer replacement generation is explicitly preserved. `testPrelaunchFailureCleanupClearsOnlyItsOwnedGeneration` covers both the cleanup and successor-fencing cases.

baron (approved):
## Review verdict: approve — no merge-blocking findings

Reviewed head `4f12abac383d9135ca34b979d48b91ee24bdbe02` against current main `f5b962d1b27da684c7fb0456ebc353ef8826f0ee`, using clean integration commit `c160a91b2e72b065be29f76454a5b0aa1edea62a` with both parents verified. Head and base were refreshed again before submitting this review and are unchanged.

### Findings
- The implementation is reasonable for the intended workflow: ordered independent Oracle answers, not synthesis or arbitration.
- Existing named single-chat continuations retain the single-chat path rather than being silently promoted when additional models are enabled.
- Shared grouped execution preserves lane ordering and separate outcomes. Cancellation drain, durable claims, and exact terminal publication have explicit ownership and recovery paths.
- Stored-roster restrictions and the frozen-pack requirement for grouped direct-headless Context Builder are intentional boundaries, not defects.
- Read-only app/Context Builder and headless reviews, plus checks of shared lifecycle, persistence, settings, and result presentation, did not establish a concrete regression warranting changes. Speculative edge cases and unmeasured performance concerns are not treated as blockers.

### Maintainer-guidance check
The user-facing invariant is independent, reviewable lane results without losing successful output when another lane fails. Canonical group storage owns the grouped lifecycle; no concrete state-safety regression was established. Scale and live UI/provider observability were not independently measured, so this approval does not claim runtime or performance certification. No additional redesign is requested.

### Validation and limits
- Exact-head CI is green: four root build/test shards, provider tests, Sentry-enabled build, style, and secret scan: https://github.com/repoprompt/repoprompt-ce/actions/runs/34138169117
- Integration diff whitespace check passed; the review worktree remained clean.
- I did not run local builds/tests or independently exercise the live multi-Oracle UI/provider flow. This approval is based on source review and verified hosted CI, not a claim of live smoke coverage.

Issue #158, opened 2026-06-11
Cursor CLI ACP runs never receive RepoPrompt MCP tools: cursor-agent 2026.06.04 ignores session/new mcpServers (stdio and http)


baron:
Follow-up to #120 / PR #126. Scoped to the remaining gap after that fix lands: **Cursor CLI ACP runs never receive RepoPrompt MCP tools**, because `cursor-agent` (latest) does not load externally supplied MCP servers in ACP mode through any channel we could find.

PR #126 fixes #120 as scoped there (Composer runs no longer fail before prompt submission; verified live with an A/B against a build without the fix). After it merges, Cursor CLI works in Agent Mode with Cursor-native tools only. This issue tracks the missing half: RepoPrompt tool injection (`read_file`, `file_search`, `ask_oracle`, …) into Cursor runs.

## Environment Tested

- RepoPrompt CE debug build from PR #126 head (`607e20b`), and isolated probes with no RepoPrompt involvement at all
- Cursor CLI: `2026.06.04-5fd875e` (confirmed latest via `cursor-agent update`)
- Workspace root: `/Users/baron/projects/oss/opencode` (latest `dev`)
- Test date: 2026-06-11 JST

## Evidence

All probes drive `cursor-agent --approve-mcps acp` directly over stdio JSON-RPC (initialize → session/new → session/prompt) with a fake MCP server that logs every spawn/contact, so RepoPrompt code is fully out of the loop.

| Channel tested | Result |
| --- | --- |
| `session/new` `mcpServers` entry, stdio shape per ACP spec (`name`/`command`/`args`/`env`), with `type: "stdio"` | Accepted without error; server **never spawned**; model reports the tool does not exist |
| Same, without `type` (pre-#126 payload shape) | Identical — never spawned |
| `session/new` `mcpServers` entry, `type: "http"` with reachable local URL | Accepted without error; server **never contacted** — despite `initialize` advertising `"mcpCapabilities": {"http": true, "sse": true}` |
| Project `<repo>/.cursor/mcp.json` (stdio) | Not loaded in ACP sessions |
| Global `CURSOR_DATA_DIR/mcp.json` (stdio) | Not loaded in ACP sessions |

Additional findings from the investigation (useful to rule out red herrings):

[...]

PR #902, opened 2026-08-29
fix: support verified Cursor agent ACP entrypoint

najibninaba:
## Summary

RepoPrompt can connect to current Cursor CLI installs that expose ACP as `agent acp`, while continuing to prefer the legacy `cursor-agent` entrypoint.
Production Cursor integrations use automatic discovery, while explicit command selections remain exact.
The generic `agent` fallback is accepted only when its trusted canonical executable resolves to `cursor-agent`, which prevents RepoPrompt from launching Grok or another unrelated CLI by mistake.
If a stale legacy executable fails ACP probing, discovery continues to the verified current entrypoint.

Capability probes share one aggregate ten-second budget, reserve synchronous termination cleanup inside that budget, and reject timed-out results even if the process reports exit status zero.

## Validation

- Fork staging PR [najibninaba/repoprompt-ce#3](https://github.com/najibninaba/repoprompt-ce/pull/3) passed all nine exact-head checks at `f12e34b4`, including all four app test shards, provider tests, style, secret scan, the Sentry-enabled build, and Greptile 5/5 review.
- Review baseline `6e60863f` is a direct ancestor of the fork-tested fixes and current upstream integration head `36b2dc8b`.
- `make dev-test FILTER=RepoPromptTests.CursorACPLaunchResolverTests` - 5 focused contracts passed against current `upstream/main`.
- `make dev-test FILTER=RepoPromptTests.CursorProductionFactoryTests` - production factory contract passed.
- `make dev-test FILTER=RepoPromptTests.CLIProcessRunnerTimeoutCleanupTests` - timeout-cleanup contract passed.
- `make dev-lint` - strict SwiftFormat and SwiftLint checks passed with repository-pinned tooling.
- `make dev-swift-build PRODUCT=RepoPrompt` - RepoPrompt product build passed.
- Mandatory commit and push preflights passed, including guardrails and redacted secret scanning.
- Live CE Agent Mode on the original implementation connected to Cursor CLI 2026.08.25, discovered 36 models, and used RepoPrompt `file_search` through the ACP session to return a synthetic marker.

The current integration merges upstream `834b43e0` without production-code conflicts.
Upstream PR #908 intentionally removed the prior broad Cursor and process lifecycle suites, so the merge retains only the five focused Cursor contracts, one production factory contract, and one timeout-cleanup contract added for this change.

## Risk Assessment

✅ **Medium**

- This changes Cursor executable discovery across Agent Mode, model polling, Settings preflight, headless discovery, and Cursor chat.
- Automatic fallback rejects unrelated generic `agent` executables before execution.
- Explicit command selections do not gain cross-name fallback.
- Existing process-runner callers retain their default cleanup behavior; Cursor probes opt into the explicit bounded policy.
- A future Cursor installer that ships only a standalone canonical binary named `agent` will fail closed until stronger static provenance is available.
- Cursor changing its ACP help wording would disable the fallback instead of launching an unverified executable.

## Related

Related: #158


baron (changes requested):
Requesting changes on exact head `6e60863fa5b1e3af793e36c5268b9a462e98c9b4`.

**Blocking — current `agent acp` support is unreachable in production.**

The hardening commit changes `launchCandidates` so a `cursor-agent` configuration only probes `.cursorAgentACP` (`CursorACPLaunchResolver.swift:361-365`), and the tests now explicitly assert that no cross-name fallback occurs (`CursorACPLaunchResolverTests.swift:54-127`). However, every production authority still configures Cursor as `cursor-agent`: `CursorAgentConfig.swift:14-17`, `ACPAgentProviderFactory.swift:27-35`, `AgentRuntimeProviderService.swift:59-69,292-303`, `CursorCLIProvider.swift:119-126`, and `CursorACPModelPollingService.swift:15-24`. There is no production `commandName: "agent"` path.

Consequently, a current Cursor installation exposing only the documented `agent acp` entrypoint cannot connect through Agent Mode, model polling, Settings preflight, headless discovery, or Cursor chat. A stale non-ACP `cursor-agent` also cannot fall through to the current entrypoint, contrary to the PR summary. Please add an authoritative, verified production command-selection/fallback path and factory-level regression coverage.

Non-blocking follow-ups: the per-candidate 10-second probes lack an aggregate deadline, and support acceptance should require `!result.timedOut` in addition to `status == 0`.

najibninaba:
Addressed the requested changes on current head `36b2dc8b`.

The reviewed head `6e60863f` remains a direct ancestor. The two fork-tested fixes add:

- An authoritative automatic production policy that tries `cursor-agent`, then a canonically verified `agent` alias.
- Exact behavior for explicit command selections, with no cross-name fallback.
- Automatic discovery across interactive Agent Mode, headless runtime, Cursor chat, model polling, and Settings preflight.
- Factory-level regression coverage for the production paths.
- One aggregate capability-probe deadline and rejection of `timedOut == true`, including status-zero results.
- Synchronous termination cleanup reserved and applied inside that aggregate budget.

I merged current upstream `834b43e0` and resolved the PR #908 test-suite deletion by retaining only the focused contracts introduced for this change, rather than restoring the removed omnibus suites.

Local validation on the integration head:

- Cursor resolver contracts: 5/5 passed.
- Production factory contract: 1/1 passed.
- Timeout-cleanup contract: 1/1 passed.
- Strict lint passed.
- RepoPrompt product build passed.
- Commit and push preflights passed.

Fresh hosted CI is running on `36b2dc8b`.

baron (changes requested):
Requesting changes on exact head `36b2dc8b1e42178f7175c5e5bd66a4d93e124d78`, refreshed against current `main` `29477196bdb123e2144af0558db2ff469340534c` via a clean conflict-free synthetic merge. The hosted merge ref is stale.

**Blocking — replace the timing-sensitive timeout-cleanup test oracle.**

`Tests/RepoPromptTests/AI/CLIProcessRunnerTimeoutCleanupTests.swift:14-29` measures the entire `CLIProcessRunner.run` wall time and requires it to stay below 2 seconds, although the explicit policy controls only the child termination phase. Environment resolution, command/process startup, scheduling, and output-reader settlement are also inside that measurement, so the assertion is host-sensitive rather than a deterministic certificate of the cleanup contract.

The exact focused command failed twice on the reviewed PR integration:

- conductor ticket `379892f7-c617-49e8-9313-53afaa259d6c`: measured `2.617s`
- conductor ticket `2ec9dcf4-362e-4062-b31b-164de47777bf`: measured `3.600s`

Both exceeded the fixed `2.0s` ceiling. The same exact-head check passed hosted CI, which further demonstrates environment sensitivity. Current `main` is one unrelated commit beyond the tested base and changes none of the PR files; the refreshed integration remains clean.

Please replace the wall-clock threshold with a deterministic barrier, termination/reap observer, or injected clock at the owning boundary. Do not fix this by simply inflating the timeout.

The previous production-wiring blocker is addressed: automatic production configuration now reaches the verified `agent acp` alias while unrelated generic `agent` executables fail closed.

najibninaba:
> Please replace the wall-clock threshold with a deterministic barrier, termination/reap observer, or injected clock at the owning boundary. Do not fix this by simply inflating the timeout.

Addressed in follow-up commit `258d493b`, directly on reviewed head `36b2dc8b`.

- Removed `CLIProcessRunnerTimeoutCleanupTests.swift` entirely. The host-sensitive threshold was not increased or replaced with another wall-clock assertion.
- Retained the deterministic resolver coverage for reserving the cleanup allowance and forwarding the selected timeout policy.
- While verifying the automatic discovery invariant, I also closed a reverse-resolution gap where `cursor-agent` could canonicalize to an unrelated generic `agent` while retaining its original lookup provenance. The guard now uses the canonical resolved launch classification.
- Added a marker-writing `cursor-agent -> agent` regression test proving the unrelated executable is rejected before probing. Explicit `agent` selection and the verified `agent -> cursor-agent` fallback remain supported.

Validation passed:

- Cursor resolver tests: 6/6
- Strict Swift lint
- Full root test suite
- RepoPrompt and `repoprompt-mcp` builds
- Conductor and CI-runner self-tests
- Generated Xcode workspace validation
- Secret scans and repository guardrails
- Complete PR-ready preflight

Thanks for catching the timing-sensitive oracle. @baron, could you take another look at the updated head?

aAtila (changes requested):
## Revision coverage

**Reviewed head SHA:** `11b08a884a5761dbca625b622a174ea65f6e90ed`
**Compared base / merge base:** `542b111e3074ed7126c0dea8a46572fa569bf51a`

No separate SHA was supplied in the request, so I pinned the PR head at the start and rechecked it before submitting. It remained unchanged. I inspected the complete ten-file PR diff, compared the immutable base/head commits (nine commits ahead, no divergence), and matched all ten changed files' post-image blob IDs to source fetched explicitly at the reviewed SHA. Surrounding source used for findings was also fetched at that SHA, not from a moving branch or the synthetic merge ref.

## Summary and proportionality

The change makes production Cursor configurations automatic while keeping explicit selections distinct, adds fallback to a trusted `agent` alias whose canonical target is `cursor-agent`, rejects timed-out capability results, and retains ownership of cancelled/timed-out probe producers until they settle. It affects interactive Agent Mode, headless/discovery, Cursor chat, model polling, and Settings preflight. It also threads an optional cleanup policy through the existing process runner without changing other callers' default policy.

The footprint is substantial for entrypoint support (+1,378/-114 across ten files), but much of it implements and tests the executable-trust and probe-lifetime contracts. I do not see a need to redesign the feature. The blocking issue is local to discovery ordering.

Assumption: preserving preference for a healthy legacy `cursor-agent` means its successful preflight should not depend on discovering an unused secondary executable. The fail-closed policy for an automatically discovered standalone canonical `agent` is explicit in the PR description; I have not treated that intentional restriction as a defect.

## Blocking

### B1 — Major · Verified — Resolve the secondary entrypoint only when it is needed

**Location:** `Sources/RepoPrompt/Infrastructure/AI/Providers/Cursor/CursorACPLaunchResolver.swift:815-823` (`launchCandidates`), together with `probeSupportSerially`'s discovery-before-deadline ordering.

A supported legacy-only installation now takes an unnecessary slow path. With a working ACP-capable `cursor-agent` on the captured PATH and no `agent` in PATH/hints, `launchCandidates` eagerly resolves both names before the first capability probe. Resolving the missing `agent` calls `CommandPathResolver.resolve(..., shellLookupMode: .fallbackOnly)`, which falls through to interactive-shell lookup.

The existing shell resolver can execute four sequential lookup queries; each has a five-second wait before termination cleanup. A slow or stuck shell startup can therefore add four five-second lookup waits, plus cleanup, before the working `cursor-agent` is even probed. This work is outside the aggregate deadline, which is created only after `resolveLaunchesForProbe` returns. It also holds the newly shared probe mutex, delaying other default Cursor preflights, and the synchronous shell lookup does not promptly observe task cancellation.

Before this diff, a healthy `cursor-agent` found in PATH did not trigger discovery of `agent` at all. This is a regression on an already-supported installation, not merely a request to include all environment discovery in the capability timer.

**Fix:** Resolve and probe candidates in preference stages. Try the legacy candidates first; only discover the `agent` entrypoint after the legacy stage fails. Preserve the static canonical-identity check before executing any automatic generic alias, and retain one aggregate capability-probe budget. Avoid eagerly doing shell work for a fallback that a successful primary path never needs.

**Coverage gap / regression assertion:** Use a working legacy executable, no `agent`, and a fake shell that records secondary lookup. Assert that preflight succeeds, caches the legacy launch, and never invokes the secondary lookup. Use a marker or barrier rather than a wall-clock upper bound. Existing resolver fixtures set `SHELL=/bin/false`, so they do not expose this regression.

**Why it matters:** A working Cursor installation can acquire a long, cancellation-insensitive connection delay, and unrelated Cursor preflights queue behind that unnecessary work.

Evidence: [eager candidate discovery](https://github.com/repoprompt/repoprompt-ce/blob/11b08a884a5761dbca625b622a174ea65f6e90ed/Sources/RepoPrompt/Infrastructure/AI/Providers/Cursor/CursorACPLaunchResolver.swift#L788-L837), [deadline placement](https://github.com/repoprompt/repoprompt-ce/blob/11b08a884a5761dbca625b622a174ea65f6e90ed/Sources/RepoPrompt/Infrastructure/AI/Providers/Cursor/CursorACPLaunchResolver.swift#L502-L540), [shell lookup and its waits](https://github.com/repoprompt/repoprompt-ce/blob/11b08a884a5761dbca625b622a174ea65f6e90ed/Sources/RepoPrompt/Infrastructure/Process/CommandPathResolver.swift#L237-L455).

## Scope revision required

None identified.

## Non-blocking

### N1 — Minor · Verified — Deduplicate the canonical probe targets

**Location:** `Sources/RepoPrompt/Infrastructure/AI/Providers/Cursor/CursorACPLaunchResolver.swift:849-865` (`validLaunches`).

The `seen` set in `launchCandidates` deduplicates expanded entry paths only. Distinct symlinks in different PATH directories can resolve to the same canonical executable; `validatedLaunch` canonicalizes them, but `validLaunches` appends every result. After a failed help probe, the loop can consequently execute the same canonical binary again with the same arguments and environment.

**Fix:** Preserve the first occurrence and deduplicate by canonical `launch.command` after trust validation, before appending. This keeps precedence intact while removing redundant attempts.

```swift
guard seenCanonicalPaths.insert(launch.command).inserted else { continue }
launches.append(launch)
```

Declare the set once for the ordered discovery/probe sequence.

**Coverage gap / regression assertion:** Provide two legacy symlinks to one failing canonical binary and a verified fallback pointing to a different binary. Record probe commands and assert that the stale binary is probed once and the distinct fallback is subsequently selected.

**Why it matters:** Repeating an identical failed probe wastes the shared ten-second budget and can reduce the opportunity to try a genuinely different installation.

Evidence: [entry-path deduplication and unconditional canonical append](https://github.com/repoprompt/repoprompt-ce/blob/11b08a884a5761dbca625b622a174ea65f6e90ed/Sources/RepoPrompt/Infrastructure/AI/Providers/Cursor/CursorACPLaunchResolver.swift#L788-L870).

## Pre-existing

No additional pre-existing findings raised. The shell resolver's synchronous waits predate this PR; B1 concerns the new unnecessary invocation of that path and its expanded effect through shared probe serialization.

## Validation and verdict

I reviewed all nine new resolver tests and the production-factory contract. The timeout/cancellation tests use explicit barriers and check late-result rejection and recovery after producer settlement. GitHub reports [CI run 2480](https://github.com/repoprompt/repoprompt-ce/actions/runs/33989866398) as successful with `head_sha` equal to the reviewed SHA. I did not run the macOS application build or native test suite in this Linux environment; findings marked Verified above are verified against the diff and source call paths, not claimed as locally executed reproductions.

**Verdict: Needs Work.** B1 is a verified reliability/regression issue that can be fixed within this PR's scope. N1 is non-blocking. CI, required approvals, and other merge gates remain separate. No code was modified.

aAtila:
## Revision-specific review delivery

**Review ID:** `5124224297`
**Actual reviewed head SHA:** `11b08a884a5761dbca625b622a174ea65f6e90ed`
**Compared base / merge base:** `542b111e3074ed7126c0dea8a46572fa569bf51a`
**Verdict:** **Needs Work** — submitted as `REQUEST_CHANGES`.

[Full review, evidence, fixes, and targeted regression assertions](https://github.com/repoprompt/repoprompt-ce/pull/902#pullrequestreview-5124224297)

The complete ten-file diff and SHA-pinned source were inspected; all changed-file post-image blob IDs matched. The PR head was unchanged when rechecked before submission. This review covers the commit above, not an older head mentioned in the PR description or a synthetic merge revision.

### Summary

Production Cursor flows now use automatic discovery, with a canonically verified `agent` fallback, explicit command-selection behavior, and bounded logical capability probes that retain ownership while cancelled producers drain. The necessary correction is local to discovery ordering, not a feature redesign.

### Blocking

**B1 — Major · Verified — `Sources/RepoPrompt/Infrastructure/AI/Providers/Cursor/CursorACPLaunchResolver.swift:815-823`: eager fallback discovery delays a healthy primary entrypoint.** With working `cursor-agent` on PATH but no `agent`, discovery runs interactive-shell lookups for the missing secondary name before probing the primary. Four five-second lookup waits plus cleanup can occur before the aggregate probe deadline even starts; the shared probe mutex also queues other Cursor preflights behind this work. Resolve/probe the legacy stage first and discover `agent` only when needed, retaining canonical trust checks. Add a regression assertion that successful legacy preflight never invokes secondary shell lookup. This prevents a connection-latency and cancellation-responsiveness regression on already-supported installations.

### Scope revision required

None.

### Non-blocking

**N1 — Minor · Verified — `Sources/RepoPrompt/Infrastructure/AI/Providers/Cursor/CursorACPLaunchResolver.swift:849-865`: duplicate canonical targets are probed repeatedly.** Distinct symlink paths survive entry-path deduplication and are appended even when they resolve to the same command. Deduplicate canonical commands while preserving first occurrence; assert that two links to one failing binary produce one attempt before a distinct verified fallback. This avoids wasting the aggregate budget.

### Pre-existing

No additional findings raised. The shell helper's waits predate the PR; B1 is its newly unnecessary invocation on the successful legacy path.

GitHub reports [CI run 2480](https://github.com/repoprompt/repoprompt-ce/actions/runs/33989866398) successful for the exact reviewed head. The macOS build/native tests were not run locally in this Linux environment. CI and other merge gates remain separate. No code was modified.

najibninaba:
@aAtila Thanks for catching these.

> Resolve/probe the legacy stage first and discover `agent` only when needed, retaining canonical trust checks.

Addressed in `e2716903`, directly on reviewed head `11b08a88`. Automatic discovery now resolves and probes `cursor-agent` first. It only looks for `agent` after the legacy candidates fail, so a healthy legacy install does not pay for fallback shell discovery. Canonical trust checks are preserved.

> Deduplicate canonical commands while preserving first occurrence; assert that two links to one failing binary produce one attempt before a distinct verified fallback.

Also addressed in `e2716903`. Candidates are deduplicated by canonical executable path across both stages. Added a regression test proving that aliases to the same failing binary produce one probe before the distinct fallback.

Discovery time stays outside the shared probe budget, while time spent on failed probes remains charged. The cleanup allowance is preserved, and cancellation is checked again after discovery before starting a probe.

The regression tests reproduce the failures before the fixes, without wall-clock thresholds.

Validation passed:

- All 14 resolver tests and the full 1,021-test suite.
- Strict lint and repository guardrails.
- Debug packaging, live CE MCP smoke, and a real Cursor session through the updated app.
- Commit and push preflights.

I also ran a second review pass and addressed the remaining findings.

#960 still covers the model and Effort/Speed controls separately. It can pick up these resolver fixes when its dependency on #902 is refreshed.

Could you take another look at the updated head?

aAtila (approved):
// duplicate

aAtila:
## Revision-specific re-review delivery

**Review ID:** `5125323341`
**Actual reviewed head SHA:** `e27169032ed2de8b0a1ea597a0340996e72f5473`
**Previous review:** `5124224297` at `11b08a884a5761dbca625b622a174ea65f6e90ed`
**PR base / merge base:** `542b111e3074ed7126c0dea8a46572fa569bf51a`
**Verdict:** **Ship** — submitted as **APPROVE**.

[Full revision-specific review](https://github.com/repoprompt/repoprompt-ce/pull/902#pullrequestreview-5125323341)

### Summary

The one-commit follow-up stages Cursor entrypoint discovery and deduplicates canonical probe targets. I reviewed the follow-up diff, refreshed the complete PR diff, and matched the two revised source blobs fetched at the pinned head to the diff's post-images. The other eight PR files are unchanged from the previous review. The head was unchanged when rechecked before submission.

Both previous findings are **resolved, Verified**:

- **B1:** `CursorACPLaunchResolver.swift::probeSupportSerially` returns after a successful legacy probe without discovering the unused `agent` entrypoint. The regression test asserts support, the cached legacy launch, and no shell-lookup marker.
- **N1:** Canonical deduplication spans the entire probe sequence. The regression test asserts one attempt for the duplicated stale binary, followed by the distinct verified fallback, including a duplicate appearing in the second stage.

### Blocking

None remaining.

### Scope revision required

None.

### Non-blocking

No new findings; previous N1 is addressed.

### Pre-existing

No additional findings raised. Discovery remains outside the capability-probe budget; the ten-second budget is not an end-to-end connection deadline.

The added tests also cover discovery-budget accounting without replenishing consumed probe time and cancellation preventing producer admission after discovery. No additional coverage gap was identified for this follow-up.

GitHub reports [CI run 2519](https://github.com/repoprompt/repoprompt-ce/actions/runs/34023200234) successful for the associated head SHA. I did not run the macOS build/native suite locally. CI, required approvals, and other merge gates remain separate. No code was modified.

baron (approved):
Re-reviewed the requested-change findings against exact head e27169032ed2de8b0a1ea597a0340996e72f5473. The findings in my previous reviews are addressed, so I am updating my review to APPROVE.

- Production wiring: CursorAgentConfig defaults to automatic selection; the resolver stages legacy cursor-agent before the canonically verified agent fallback. Explicit selection remains exact. CursorProductionFactoryTests covers interactive, headless, chat, and polling factories.
- Timeout-cleanup test: the host-sensitive CLIProcessRunnerTimeoutCleanupTests file is removed. CursorACPLaunchResolverTests retains an injected-clock contract that checks execution budget and cleanup allowance directly, plus barrier-driven deadline/cancellation and producer-settlement coverage. Timed-out results are rejected even with exit status zero.
- Latest discovery findings B1/N1: successful legacy probing returns before secondary lookup; a shared canonical-path set suppresses duplicate probes across stages. The tests assert no secondary-shell marker and exactly one probe per distinct canonical target. This agrees with aAtila's approval of this same head.

Validation: all eight hosted checks are successful for e2716903 (CI run 34023200234). This pass inspected the pinned source and tests; it did not rerun the native suite or packaged Cursor acceptance. Current-main integration and release acceptance remain separate gates. There are no inline review threads to resolve on this PR.

Maintainer-guidance check: the user-facing invariant is verified automatic Cursor discovery without unnecessary secondary lookup or execution of unrelated binaries. Root-cause confidence is confirmed from the owning resolver and production configuration. Canonical trust and retained probe ownership preserve the safety boundary; canonical deduplication avoids repeated work. Scope is closure of these existing findings; validation is pinned-source review plus exact-head hosted CI.

PR #906, opened 2026-08-29
Fix: restore global Codex instructions in Agent Mode

### LINKED ISSUE #831: Agent Mode ignores the user's global ~/.codex/AGENTS.md
As @w-winter reported in https://github.com/repoprompt/repoprompt-ce/issues/665#issuecomment-5173931965: Codex normally composes `$CODEX_HOME/AGENTS.md` with project-local `AGENTS.md` files, but CE's app-private `CODEX_HOME` (#612) contains no such file, so Agent Mode sessions silently lose the user's global instructions. Losing global instructions quietly breaks people's standing workflows. Filing this as its own issue so it's tracked independently of #665's project-hook scope.

Preliminary investigation notes (candidate direction only — not a settled design):

- Unlike global skills (fixed in #824 via Codex's dedicated `skills/extraRoots/set` request), Codex 0.147.0 offers no protocol method or config key that reproduces AGENTS-style composition: `thread/start.developerInstructions` / `developer_instructions` can inject text but without AGENTS precedence, composition, or `instructionSources` provenance, and `baseInstructions`/`instructions`/`model_instructions_file` replace the model's base prompt (a channel CE already uses for its system prompt).
- Codex's native behavior (checked at rust-v0.147.0: `codex-home/src/instructions/mod.rs`, `core/src/agents_md.rs`) reads `$CODEX_HOME/AGENTS.override.md` then `$CODEX_HOME/AGENTS.md` from disk and prepends it to the project-doc chain — so preserving native semantics appears to require a real file at the managed home path (projection of the user's file), with copy-vs-link and refresh timing as the main open questions.

@w-winter offered to take this on; happy for that, and I can also pick it up within the week otherwise.

w-winter:
Fixes https://github.com/repoprompt/repoprompt-ce/issues/831.  RepoPrompt CE uses an app-private `CODEX_HOME`, so Codex Agent Mode sessions load project-local `AGENTS.md` files but were silently missing the user's global `~/.codex/AGENTS.override.md` or `~/.codex/AGENTS.md`.

This patch projects both of those global instruction files into the managed Codex home before launch, leaving precedence and global-plus-project composition to Codex.

PR https://github.com/repoprompt/repoprompt-ce/pull/837 contains earlier work in this area as part of a broader managed-home change.  @baron agreed that this focused PR should proceed.  Afterward, we'll evaluate a pared-down version of #837 for any remaining work that doesn't overlap.

## Summary

- Project `AGENTS.override.md` and `AGENTS.md` independently into the managed Codex home
- Preserve Codex's native empty-override fallback and global-plus-project instruction composition
- Refresh the projection before app-server and `codex exec` launches, including cached app-server runtimes
- Use atomic regular-file copies rather than writable links into the user's real Codex home
- Track projected files with a small ownership manifest
- Fail closed instead of overwriting unowned, modified, or symlinked managed-home files

## Implementation Notes

- `CodexRuntimeAuthority.Runtime.prepareState()` remains the single preparation point for managed Codex state
- Source symlinks are supported, matching normal Codex file loading; managed-home destination symlinks are rejected
- Projection errors use the existing app-server and exec launch error paths
- No settings, logging, watcher, or general managed-home framework was added

## Review Approach

- The patch received focused reviews from 5.6-Sol-Pro and Fable against the scope of #831
- Findings around manifest hash parsing, ownership assertions, and cached-runtime recovery were fixed with focused regressions

## Validation

Passed locally:

- `make dev-test FILTER=CodexGlobalInstructionsProjectionTests`
- `make dev-test FILTER=CodexRuntimeAuthorityTests`
- `make dev-test FILTER=CodexMCPBootstrapReadinessTests`
- `make dev-test FILTER=CodexAppServerClientProcessExitTests`
- `make dev-test FILTER=CodexExecAgentProviderRuntimePreparationTests`
- `make dev-test FILTER=CodexNativeSessionControllerGoalConfigTests`
- `make dev-test FILTER=CodexNativeSessionControllerHookApprovalTests`
- `make dev-lint`
- Repository guardrails and the exact one-commit outgoing-range secret scan

The full root suite was also run.  Three Codex coordinator timing tests outside this patch failed under full-suite load and all three passed on focused rerun.  All projection and launch-path tests passed.  I also confirmed via a Debug app canary that global and project-local instructions were both loaded in the expected order.

### DIFF
--- a/Sources/RepoPrompt/Infrastructure/AI/Providers/Codex/AppServer/CodexAppServerClient.swift
+++ b/Sources/RepoPrompt/Infrastructure/AI/Providers/Codex/AppServer/CodexAppServerClient.swift
@@ -443,6 +443,7 @@ actor CodexAppServerClient {
     private let livenessProbe: @Sendable (SpawnedProcess) -> Bool
     private let processSpawnPreparation: @Sendable () async throws -> Void
     private let processEnvironmentBuilder: @Sendable (ProcessEnvironmentRequest) async -> ProcessEnvironmentResult
+    private let runtimeStatePreparer: @Sendable (CodexRuntimeAuthority.Runtime) throws -> Void
     private let provisionsRepoPromptMCPOnStart: Bool
     private let processExitObserverFactory: @Sendable (pid_t) -> ChildProcessExitObserver
     private let expectedAgentPIDRegistrar: ExpectedAgentPIDRegistrar
@@ -466,6 +467,9 @@ actor CodexAppServerClient {
         processEnvironmentBuilder: @escaping @Sendable (ProcessEnvironmentRequest) async -> ProcessEnvironmentResult = {
             await ProcessEnvironmentBuilder.build($0)
         },
+        runtimeStatePreparer: @escaping @Sendable (CodexRuntimeAuthority.Runtime) throws -> Void = {
+            try $0.prepareState()
+        },
         provisionsRepoPromptMCPOnStart: Bool = true,
         processExitObserverFactory: @escaping @Sendable (pid_t) -> ChildProcessExitObserver = {
             ChildProcessExitObserver(pid: $0)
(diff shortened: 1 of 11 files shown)


baron (approved):
Reviewed against current `origin/main` in a clean synthetic merge at PR head `e28ee9c87ee4306abea4ec75481eca81c2a820ac`.

No actionable findings. The change preserves Codex global-instruction precedence while keeping projection ownership and conflict handling fail-closed, and the relevant launch paths are covered. All hosted checks are green.

Local focused-test attempts were canceled during compilation before any test case executed, so those attempts are not counted as validation.

PR #913, opened 2026-08-30
Fix: codemaps to index nested repo worktree sources

### LINKED ISSUE #886: Codemap indexing retries forever when a workspace root contains nested Git repositories
## Impact

A workspace root that contains an embedded Git repository can leave background codemap indexing permanently stuck on one catalog page.  In my observed case, progress stayed at 64 of 1,774 candidates while the same page retried with increasing backoff.  Resetting codemap caches and reloading the root reproduced the failure.

## Expected behavior

The parent root should terminally exclude source paths owned by nested repositories, finish its own codemap, and tell the user which nested repositories need to be added as separate workspace roots.

## Actual behavior

RepoPrompt discovers supported source files below the nested repository boundary.  Git identity classification returns `unavailable(repositoryUnavailable)` for one of those files.  The graph index converts that result to a transient resolution, discards the completed page, and retries the same cursor indefinitely.  Between retries, the app has no active or draining graph-index batch, so the index appears idle at a fixed percentage.

## Reproduction

1. Open a Git repository as a RepoPrompt workspace root
2. Place another Git repository beneath it, for example `<outer-root>/<nested-repo>/.git`, with at least one supported source file
3. Don't add the nested repository as a separate workspace root, and leave its source visible to the parent root
4. Start a fresh codemap index for the parent root
5. Observe that indexing completes one page and repeatedly retries it instead of advancing

The failure reproduced consistently in the investigated workspace.

## Environment

- RepoPrompt CE release 1.3.0, build 35.15.18
- macOS 26.6.2
- arm64

## Focused evidence

A fresh DEBUG artifact runtime using the actual root with the production workspace store, catalog, and Git-classification path reproduced the failure.  The retained job stayed on page 1 with 64 processed candidates and entered `suspendedBusy` repeatedly.  The same nested source path returned `repositoryUnavailable` on every attempt.

A separate parser sweep processed all 898 supported source files, including generated files that normal ignore policy omits, in 5.97 seconds.  This ruled out a parser hang.

After nested repositories were excluded from the parent root, the same fresh-runtime diagnostic completed 303 of 303 candidates across five pages in 5.88 seconds with no retries.

## Confirmed code path

- `GitBlobIdentityService.pathSecurityState` detects a nested `.git` marker and returns `.repositoryBoundary`.
- `GitBlobIdentityService.makeClassification` converts that state to `.unavailable(.repositoryUnavailable)`.
- `WorkspaceCodemapBindingEngine.resolveGraphIndexCandidate` converts `.unavailable` to `.transient`.
- `processGraphIndexBatch` returns `.retry`, preserving the same page cursor and repeating the whole page.
- `WorkspaceCodemapGraphTerminalExclusionReason.repositoryBoundary` already exists but is not used for this path.

## Product contract

Nested repos should retain independent Git ownership.  Their HEAD, index, attributes, ignore rules, blob namespace, manifests, and invalidation lifecycle should consequently come from separate RepoPrompt roots.  The parent root shouldn't treat nested files as raw parent-owned worktree sources.

## Acceptance criteria

- A repository-boundary candidate can't cause an otherwise completed catalog page to retry indefinitely
- The parent root records nested-repository candidates as terminally excluded with reason `repositoryBoundary` and completes indexing its remaining files
- Root status reports that nested repositories were excluded without exposing absolute paths
- The workspace UI provides an actionable way to add detected nested repositories as separate roots, with user confirmation and no duplicate roots
- A focused regression test creates an outer repository with a nested repository, verifies zero boundary-driven retries, and verifies that the parent graph reaches a terminal ready state

## Related issues

No duplicate found

w-winter:
Fixes https://github.com/repoprompt/repoprompt-ce/issues/886.  A workspace root containing nested Git repos could leave background codemap (CM) indexing stuck retrying the same catalog page indefinitely.

The likely regression came from the content-addressed CM source-acquisition path (https://github.com/repoprompt/repoprompt-ce/pull/314).  Nested `.git` detection correctly prevented the outer repo's blob identity from authorizing those files, but classified them as repo-unavailable instead of using the existing validated-worktree path.  The graph treated that stable condition as transient, discarded the completed page, and retried the same cursor.

Instead of excluding nested source this patch restores CM eligibility for every supported file admitted by the workspace catalog.  Nested files are built from their current fingerprint-validated worktree bytes and normal ignore policy is unchanged.

## Summary

- Continue descriptor-safe traversal beyond nested `.git` directory and pointer-file markers
- Preserve the requested leaf fingerprint instead of the repo marker fingerprint
- Route regular nested source through the existing validated-worktree CM path
- Prevent outer Git blobs, attributes, status, and path-local metadata from becoming authoritative after a nested boundary
- Preserve existing missing, symlink, nonregular, and probe-failure outcomes
- Complete graph indexing without boundary-driven retries or terminal exclusions

## Implementation Notes

- `GitBlobIdentityService` remains the sole source-provenance classifier
- The graph index, workspace catalog, root topology, ignore handling, and UI are unchanged
- No nested Git commands, repo lifecycle, retry mechanism, compatibility path, or new source-acquisition path was added
- Outer index entries remain available as non-authoritative evidence that the nested boundary takes precedence

## Review Approach

- The patch received one focused maintainability review and complete-patch correctness reviews tethered to the originating issue's scope
- Findings around nested-owned Git metadata and retaining boundary state across terminal leaf outcomes were fixed with focused regressions

## Validation

Passed locally:

- `make dev-test FILTER=GitBlobIdentityServiceTests`
- `make dev-test FILTER=CodemapAutomaticSelectionGraphNativeTests`
- `make dev-lint`
- PR-ready preflight, including the matching path-selected heavyweight lanes
- Commit and push safety preflights, including staged-index and outgoing-range secret scans

I also validated the full flow in the Debug app with a real workspace root containing nested Git repos.  Once the repos were admitted by the workspace's existing ignore policy, they remained visible through the parent root and agents could retrieve their CMs while the parent graph completed normally.

### DIFF
--- a/Sources/RepoPrompt/Infrastructure/VCS/GitBlobIdentityModels.swift
+++ b/Sources/RepoPrompt/Infrastructure/VCS/GitBlobIdentityModels.swift
@@ -251,6 +251,7 @@ enum GitBlobValidatedWorktreeReason: String, Codable, Hashable {
     case indexFlag
     case checkoutTransformation
     case changedDuringClassification
+    case nestedRepository
     case generatedOrExplicit
 }
 
--- a/Sources/RepoPrompt/Infrastructure/VCS/GitBlobIdentityService.swift
+++ b/Sources/RepoPrompt/Infrastructure/VCS/GitBlobIdentityService.swift
@@ -15,13 +15,23 @@ actor GitBlobIdentityService {
         let unstable: Bool
     }
 
-    private enum PathSecurityState {
+    private struct PathSecurityAssessment: Equatable {
+        let state: PathSecurityState
+        let crossedNestedRepository: Bool
+
+        init(_ state: PathSecurityState, crossedNestedRepository: Bool = false) {
+            self.state = state
+            self.crossedNestedRepository = crossedNestedRepository
+        }
+    }
+
+    private enum PathSecurityState: Equatable {
         case regular(GitBlobLStatFingerprint)
         case missing
         case symlinkLeaf(GitBlobLStatFingerprint)
         case symlinkComponent
         case nonRegular(GitBlobLStatFingerprint)
-        case repositoryBoundary(GitBlobLStatFingerprint?)
+        case repositoryProbeFailed
     }
 
     private let gitService: GitService
@@ -134,10 +144,16 @@ actor GitBlobIdentityService {
         let preSecurity = relativePaths.map { path in
             pathSecurityState(workspaceRoot: workspaceRoot, relativePath: path)
         }
+        let worktreeMetadataRepositoryPaths = repositoryPaths.indices.compactMap { index -> String? in
+            guard let repositoryPath = repositoryPaths[index],
+                  !preSecurity[index].crossedNestedRepository
+            else { return nil }
+            return repositoryPath
+        }
         let preFileFingerprints = preSecurity.map(Self.fingerprint)
         let preMetadata = Self.repositoryMetadataFingerprint(
             layout: layout,
-            repositoryRelativePaths: requestedRepositoryPaths
+            repositoryRelativePaths: worktreeMetadataRepositoryPaths
         )
         let preLayout = Self.repositoryLayoutFingerprint(layout)
         let preIndexFingerprint = Self.lstatFingerprint(at: layout.gitDir.appendingPathComponent("index"))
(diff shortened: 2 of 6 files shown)


baron (approved):
Reviewed exact PR head ad0112af62c7f7184d787550804bff331d299d07 against freshly fetched origin/main 64b49fbc07b40df421bb0be4fa12540eeb8733d0 via synthetic merge d5c54f172f59634734008ba96e20a663a7b4ab59.

No actionable P0-P3 findings. Focused validation passed on the synthetic merge:
- GitBlobIdentityServiceTests (1 test, 0 failures)
- CodemapAutomaticSelectionGraphNativeTests (1 test, 0 failures)

Hosted raw-head checks are green. Validation limits: no separate full suite, lint, product build, or live-app validation was run on the synthetic merge.

PR #915, opened 2026-08-31
Fix duplicate workspace consolidation persistence

crstdr:
## Summary

- persist duplicate consolidation as a recoverable workspace retirement, so exact duplicates stay consolidated across authority projections and app restarts
- rehome Oracle chat and Agent Mode session sidecars into canonical storage before merging tabs and retiring a duplicate, while preserving each source as a recovery copy
- preflight **both** sidecar families before either family writes; divergent collisions now fail the duplicate closed without a one-family partial migration
- report fail-closed saves by phase, distinguish a committed working document from a completed saved document, and reconcile presentation state from authority after partial failure
- fence consolidated-workspace restoration until its exact saved phase and index flush complete, and refresh authoritative lifecycle state at the final activation boundary so stale saved documents cannot reopen a retired row
- preserve the window session snapshot when a window's restore target cannot be resolved, instead of overwriting it with the Default fallback layout
- serialize sidecar inventory validation with the existing Agent/Chat write boundaries, including pending writes, pruning, deletes, and tracked Oracle autosaves
- fail closed when dirty recovery state cannot be verified, and preserve newer local edits across delayed authority self-echo projections
- hold the existing workspace activation/deletion lease through explicitly targeted headless Oracle autosaves
- keep retired recovery records out of the ordinary Manage Workspaces list while preserving them in a collapsed **Consolidated workspaces** section with explicit restore controls

## Scope / YAGNI

This stays inside the existing workspace authority, save, sidecar, and presentation seams. It adds no UUID remapping, migration framework, retry queue, new persistence system, or settings surface.

**Rename-safe storage is deliberately not part of this PR.** An earlier revision resolved storage by scanning the workspace root for a matching UUID on every lookup. On a profile with 15,543 workspace directories that cost roughly 7 s of blocking main-actor I/O at startup (43 lookups x 3 windows) and crashed release launch, so it was reverted here. A replacement that resolves by identity through the already-persisted `workspace-catalog.json` is deferred to its own PR, because review showed it must land together with catalog persistence of recovered entries and identity-resolved sidecar rehoming -- without those, making renamed workspaces loadable lets them enter duplicate cleanup, where sidecars still resolve by display name, copy nothing, and report success.

Consequence for this PR: a renamed workspace still fails to load, exactly as on `main`. This PR neither causes nor fixes that. It does stop that failure from destroying the user's window layout.

Ordinary workspace save conflict recovery is unchanged. Source sidecars and duplicate directories are never deleted by consolidation. Unhiding a retired workspace clears its consolidation marker and makes it eligible for duplicate detection again.

## Review findings addressed

1. **Cross-family sidecar preflight:** Agent and Chat batches are both prepared before either commits.
2. **Partial authority save visibility:** the bridge returns working/saved phase outcomes and cleanup reports/reconciles a working-only commit accurately.
3. **Restore and activation races:** a transient exact-workspace fence blocks switching until restore persistence is complete; failed or partial saves reconcile from authority, and authority-backed activation performs a fresh final lifecycle classification after disk load. Local ephemeral workspaces retain their predecessor path because they are intentionally absent from the authority catalog and excluded from consolidation.
4. **Window layout preservation:** a window whose restore target cannot be resolved falls back to the system Default workspace, and capture previously persisted that observed state over the real snapshot (one field profile went from 14 real windows to 1 real + 14 Default, permanently). The failed entry is now armed *before* the switch is attempted, so a resolved target whose switch is refused is covered too, re-emitted while the window sits on a system workspace, and cleared once the window lands on any non-system workspace so a later deliberate Default arrangement still persists.
5. **Serialized sidecar safety:** complete source inventory/content and destination state are revalidated inside the existing Agent/Chat serialization boundaries; pending writes and tracked autosaves cannot race retirement.
6. **Recovery lifecycle fencing:** stale snapshots, ordinary saves, projections, cross-window hidden-state fan-out, and unverifiable saved documents cannot clear or bypass a retirement/restore lifecycle transition.
7. **Headless history safety:** explicit-workspace Oracle runs hold the existing activation lease through awaited autosave and revalidate recovery state before publication.
8. **Self-echo preservation:** a delayed accepted self-echo advances authority baselines without replacing the bridge cache's newer local model.

## Validation

- `make dev-test FILTER=WorkspaceDuplicateCleanupTests` — **7/7 passed**
- `make dev-format` — **0/1,283 files changed**
- `make dev-lint` — strict SwiftFormat / SwiftLint passed
- `make dev-swift-build PRODUCT=all` — both `RepoPrompt` and `repoprompt-mcp` passed
- `git diff --check` — passed
- contribution commit and push guardrails / secret scans — passed

Coverage remains intentionally focused. The rename-resolver test was removed with the resolver it exercised; it created 1,025 files to assert a single lookup and did not model the repeated startup lookups that caused the regression.

Also in this branch: a dead `if`/`else` in `suppressSelfEcho` whose arms both returned `model` was collapsed (it paid a full document encode per self-echo event and was the only caller of an accessor widened from `private` to `fileprivate`, now restored), and a comment claiming ordinary switches do not pay for a catalog snapshot was corrected, since authority-mode activation checks every non-ephemeral switch.


baron (changes requested):
Requesting changes after reviewing a local synthetic merge of current `origin/main` `5bc82b7f43ae66afe5a3bdfdb186b1efb19ab985` with PR head `7ace2961d58fc5a37f106fd41d76b0778fb72099` (synthetic merge `3c58be65f81f42b5b18f42ef4377f4ba5fe60c55`). GitHub's hosted merge ref was still based on the previous main, so the findings below were checked against the current-base synthetic merge.

1. **P1 — Restored copies become switchable before restoration is durable.** `setWorkspaceHidden` clears `consolidatedIntoWorkspaceID` in the in-memory model immediately (`WorkspaceManagerViewModel.swift:7548-7554`) and only then starts the asynchronous save (`:7560-7595`); failure merely prints. The management view consequently moves that row into the ordinary list (`ManageWorkspacesView.swift:476-480`) and enables `requestWorkspaceSwitch` (`:553-564`), whose admission path has no consolidated-marker check (`WorkspaceManagerViewModel.swift:2986-3028`). With a delayed or failed save, a user can click the eye and then switch into a record that durable authority still considers a consolidated recovery copy. Please keep the record non-switchable until the restore save succeeds, or make activation revalidate durable authority.

2. **P1 — A failed second phase of `saveFailClosed` is reported as total cleanup failure after the first phase was durably committed.** `DomainWorkspacePresentationBridge.saveFailClosed` persists `replaceWorkingDocument` and then independently submits `saveWorkspaceDocument` (`DomainWorkspacePresentationBridge.swift:105-135`). The working commit is journaled before the second command (`DomainWorkspaceContextAuthority.swift:1516-1527`), and bootstrap restores `journal.workingDocument` (`DomainPersistence.swift:1078-1098`). If the saved phase conflicts/fails, the cleanup catch reconciles but still records every duplicate as `persist_failed` and does not count the group (`WorkspaceManagerViewModel.swift:6667-6687`; the retirement path similarly rethrows at `:7142-7157`). A restart can therefore expose the merged/retired working document even though the result says nothing was consolidated. Please make the cleanup result reflect the reconciled durable outcome, or make the operation atomic/explicitly partial-success.

3. **P2 — Sidecar migration can partially commit before a later store rejects the duplicate.** Agent sessions are fully rehomed before chat migration begins (`WorkspaceManagerViewModel.swift:6606-6618`); the agent service writes each prepared destination and rebuilds its index (`AgentSessionDataService.swift:982-1011`). A divergent chat collision then throws during the later chat preflight (`ChatHistoryManager.swift:108-125`, `WorkspaceSessionSidecarMigration.swift:66-76`), and the catch only marks the workspace skipped (`WorkspaceManagerViewModel.swift:6620-6629`). This leaves agent artifacts in canonical storage even though the duplicate was not consolidated, contrary to the stated whole-duplicate failure behavior. Please preflight both stores before either writes, or roll back newly created destinations on failure.

The focused `WorkspaceDuplicateCleanupTests` pass (3/3), but they cover successful restore/consolidation and do not exercise these delayed-save, second-phase-failure, or cross-store-collision paths. All hosted checks are green on the PR head.

crstdr:
Updated in `6a228a83` to address the requested consolidation safety fixes: both sidecar families now preflight before either writes; fail-closed saves preserve working-vs-saved phase evidence; restored consolidated workspaces remain switch-fenced until persistence finishes; and the small rename-storage fix uses stable UUID/catalog identity without adding a migration framework. Focused tests are 6/6, strict lint and the RepoPrompt product build pass. Fresh CI is running.

crstdr:
Follow-up `22b6c8f9` closes the final exact-review finding. UUID discovery now completes the full root scan before choosing even the current-name directory, discovered memos cannot mask a newly-created same-UUID sibling, and mutating Git-data, consolidation, and deletion paths fail closed instead of using the presentation fallback. The existing rename regression now uses a current-name + old-name ambiguity and asserts Git-data resolution refuses it. Focused suite remains 6/6; strict lint and RepoPrompt build pass.

crstdr:
Final exact-diff review after `22b6c8f9`: **Oracle APPROVE**.

The review verified that the prior ambiguity blocker is closed: full UUID-directory scanning now precedes fallback selection, discovered/custom memo entries cannot bypass that scan, current-name + legacy-name ambiguity fails closed, and Git-data/consolidation/deletion mutations use authoritative or throwing directory resolution. The existing focused regression remains six tests total and pins the ambiguity plus Git-data refusal. No new correctness or safety blocker was found.

Sources/RepoPrompt/Features/AgentMode/Runtime/AgentSessionDataService.swift
```diff
@@ -1503,20 +1570,20 @@ actor AgentSessionDataService {
...
-            if !FileManager.default.fileExists(atPath: root.path) {
-                try FileManager.default.createDirectory(at: root, withIntermediateDirectories: true)
-            }
-            let folderName = WorkspaceDirectoryName.directoryName(name: workspace.name, id: workspace.id)
-            let workspaceDir = root.appendingPathComponent(folderName)
-            if !FileManager.default.fileExists(atPath: workspaceDir.path) {
-                try FileManager.default.createDirectory(at: workspaceDir, withIntermediateDirectories: true)
-            }
-            return workspaceDir
+        let root = WorkspaceStoragePaths.configuredRoot()
```
dsebban: This also relocates Agent sidecar lookup from the pre-change identity application-support root to `GlobalCustomStorageURL`. Existing custom-root users still have Agent history in the old root, and there is no migration here; the root-agnostic catalog memo can redirect this resolver away from that old tree too. Please preserve the predecessor root and use UUID scanning only to replace the mutable-name directory lookup, or add a proven migration. The paired Chat change has the same compatibility boundary.

Sources/RepoPrompt/Features/Workspaces/ViewModels/WorkspaceManagerViewModel.swift
```diff
@@ -2980,6 +3018,11 @@ class WorkspaceManagerViewModel: ObservableObject {
 
     @MainActor
     func requestWorkspaceSwitch(to newWorkspace: WorkspaceModel, saveState: Bool = true, reason: String = "userOrInternal") async -> WorkspaceSwitchResult {
+        if pendingConsolidatedRestoreIDs.contains(newWorkspace.id) {
```
dsebban: This fences only an in-progress restore. A persisted retired record with `consolidatedIntoWorkspaceID != nil` is still accepted by `requestWorkspaceSwitch`; MCP can resolve hidden workspaces explicitly by UUID and reaches this method. That lets a recovery copy open before the user restores it. Please also reject the current authoritative model while its consolidation marker is set, and pin that with a pre-restore switch assertion.

Sources/RepoPrompt/Features/Workspaces/ViewModels/WorkspaceManagerViewModel.swift
```diff
@@ -7291,35 +7702,83 @@ class WorkspaceManagerViewModel: ObservableObject {
     }
 
     func setWorkspaceHiddenFromSnapshot(_ workspace: WorkspaceModel, hidden: Bool) async throws -> WorkspaceModel {
+        guard let index = workspaces.firstIndex(where: { $0.id == workspace.id }),
```
dsebban: This exact-equality guard now applies to every `setWorkspaceHiddenFromSnapshot` caller, not just consolidated restoration. MCP hide/unhide resolves a disk snapshot; if the in-memory workspace has unsaved working changes, the models differ and an ordinary hide/unhide that previously succeeded now fails with cleanup error 9. Please keep the base body for ordinary hidden-state mutations and use this exact/fenced path only when clearing a consolidation marker (with a distinct save source).

Sources/RepoPrompt/Features/Workspaces/ViewModels/WorkspaceManagerViewModel.swift
```diff
@@ -8841,25 +9300,68 @@ class WorkspaceManagerViewModel: ObservableObject {
...
+        )
+        if requiresExactSnapshotAttempt {
+            guard let current,
+                  current == workspace,
+                  domainWorkspaceRevisionsByID[workspace.id]?.workingRevision != nil,
+                  domainWorkspaceDigestsByID[workspace.id] != nil
+            else {
+                throw staleCleanupAttemptError
+            }
+        } else if let current, current != workspace {
```
dsebban: This changes every ordinary save: when the caller's captured payload differs from the current projection, the method silently substitutes the in-memory model. That semantic change is not required by the cleanup exact-save path, which is already handled by the branch above, and it changes which user state reaches durable storage. Please revert this ordinary-path substitution unless it is split out with its own reproduction and regression coverage.

dsebban (commented):
# Independent deletion-first review — PR #915 (`22b6c8f9` vs `03266e8b`)

Legend: **[E]** = evidence from the supplied patch/excerpts; **[I]** = inference. Marks: `KEEP` / `DELETE/REVERT` / `SHRINK`. Classes: *root* (root-cause requirement), *hard* (hardening), *spec* (speculation), *test*, *defect* (correctness defect).

Scope confirmed: 16 files, +2277/−220, head `22b6c8f9`, base `03266e8b`. Ranges below use **head** line numbers (approximate where the hunk context makes exact lines ambiguous).

---

## Pass 1 — Semantic-delta audit by changed range

### `Features/AgentMode/Runtime/AgentSessionDataService.swift`

| Range | Mark | Class | Delta vs predecessor / trace | Risk of deletion; smallest patch |
|---|---|---|---|---|
| ~957–994 `prepareWorkspaceSessionRehome` | KEEP | root | New; no predecessor. Read-only: enumerates `AgentSession-*.json`, waits `diskWriter` idle on source+destination, calls shared `prepareCopies`. Returns `nil` when no source files. Throws (not swallows) on parse/collision. No durable side effect [E]. | Deletion breaks the Agent copy family. Nit: `sessionFileURLs` enumerated twice (here + inside `prepareCopies`) — harmless. |
| ~996–1025 `commitWorkspaceSessionRehome` | KEEP | root (+hard) | Writes prepared copies through `diskWriter.enqueueAndWait`, restores mtime, then `reconcileMetadataIndex`. `deletedSessionFileURLs` tombstone check mirrors existing `saveAgentSession` invariant [E]. Sources untouched. Partial commit possible on I/O failure mid-batch (disclosed as `sidecar_migration_failed`). | Tombstone guard is consistent hardening; keep. |
| 1573–1589 `workspaceFolderURL` | **SHRINK** | root + **unverified semantic delta** | Predecessor root: `MCPFilesystemConstants.identity.applicationSupportRootURL()/Workspaces`, folder by raw name. Head root: `WorkspaceStoragePaths.configuredRoot()` (honors `GlobalCustomStorageURL`), folder via resolver [E]. Two independent changes: (1) folder by UUID — required for rename; (2) **root relocation** — not required. Test 4 asserts `manager.gitDataDirectory(for:)` throws ambiguity under `currentBaseRoot`, which only holds if the manager honors `GlobalCustomStorageURL` [E]; base services did **not** [E]. So for custom-root users base kept sidecars in the default root and head redirects them (both via `configuredRoot()` and via the process-wide catalog memo) to the catalog root → existing history becomes invisible without migration [I]. Also `identity.applicationSupportRootURL()` may differ from the hardcoded `"RepoPrompt CE"` in dev identities [I]. | Keep base `root` lines verbatim; replace only the folder-name line with `WorkspaceStorageDirectoryResolver.shared.resolveDirectory(…, catalogFileURL: nil, baseRoot: root)`; make the resolver's catalog memo root-scoped (see resolver). Requires owner evidence on custom-root/identity before adopting head's relocation. |

### `Features/AgentMode/WorkspaceDirectoryName.swift`

| Range | Mark | Class | Delta | Risk / patch |
|---|---|---|---|---|
| 2, 6–8, 14 | **DELETE/REVERT** | spec | `directoryName` now delegates to the domain sanitizer; doc comment rewritten; new cross-module import. Its only shown producer (`AgentSessionDataService`) no longer calls it [E] → the change is dead for this PR. Changes emitted names for special-character workspace names for any remaining caller [I]. | Zero risk to revert. Revert file to base. |

### `Features/Chat/Services/ChatHistoryManager.swift`

| Range | Mark | Class | Delta | Risk / patch |
|---|---|---|---|---|
| 85–125 `prepareWorkspaceSessionRehome` | KEEP | root | Read-only preflight on the existing serial `fileSaveQueue`; bypasses `listChatSessions` because that path enforces history limits and may delete files [E — comment]. Throws on malformed/collision. | Keep. Optional nit: drop the separate `sessionFileURLs` call and use `prepareCopies(...).isEmpty` (−6). |
| 127–145 `commitWorkspaceSessionRehome` | KEEP | root | Atomic writes + mtime on the serial queue. | Keep. |
| 517–533 `workspaceFolderURL` | **SHRINK** | root + unverified delta | Same two-part change as Agent: UUID folder (needed) + root relocation from hardcoded `~/Library/Application Support/RepoPrompt CE/Workspaces` to `configuredRoot()` (not needed) [E]. | Same patch as Agent: preserve base root computation; resolver for the folder only. |

### `Features/WorkspaceFiles/ViewModels/WorkspaceFilesViewModel.swift`

| 3372 `try workspaceManager.gitDataDirectory(for:)` | KEEP | hard | One-token consequence of `gitDataDirectory` throwing; enclosing function already `async throws` [E]. Under UUID ambiguity Git-data writes now fail closed instead of writing to a name-derived dir. Tested (test 4). | Keep if `gitDataDirectory` stays throwing (recommended for a write path). |

### `Features/Workspaces/ViewModels/WorkspaceManagerViewModel.swift`

| Range | Mark | Class | Delta / trace | Risk / patch |
|---|---|---|---|---|
| 57–62 `WorkspaceStoragePaths.configuredRoot` | **DELETE** | spec / data-twice | Duplicates the manager's `currentBaseRoot` policy (`GlobalCustomStorageURL ?? defaultRoot`) [I from test 4]; only consumers are the two service `workspaceFolderURL` rewrites. | Delete together with the service root relocation. |
| 316–322 `workingCommitted` on `DomainWorkspaceAuthorityOperationError` | KEEP (optional) | hard (reporting) | Adds a flag; default `false` preserves all existing constructions [E]. Consumed only by `duplicateCleanupPersistenceFailureReason`. | Keep; addresses baron #2's disclosure ask. Manager-level label is untested (see tests). |
| 410 `pendingConsolidatedRestoreIDs` | KEEP | root | New fence set; published for the view. | Required. |
| 953, 959–961 DEBUG backup-dir override | KEEP | test infra | Prevents tests writing to real Application Support. | Keep. |
| 1857–1867 `resolvedWorkspaceDirectoryForMutation` | KEEP | root | Throwing resolver call with catalog URL from `domainWorkspaceFileURLsByID`. | Keep. Name is misleading (also used for reads) — nit. |
| 1869–1878 `workspaceDirectory(for:)` | KEEP | root | Was pure name derivation; now resolver with logged fallback to name derivation on error [E]. Non-authority `workspaceFileURL(for:)` and flush URLs route here. Fallback preserves base behavior in the only new failure state (ambiguity). | Keep. |
| 1891–1926 `workspaceFileURL(for:)` / removed `chatsFolder`, `gitDataDirectory(baseRoot:)`, `gitDataTabDirectory(baseRoot:)` / new `resolvedWorkspaceDirectory(for:baseRoot:)` / throwing `workspaceFileURL(for:baseRoot:)`, `gitDataDirectory`, `gitDataTabDirectory` | KEEP | root (+hard on throws) | Removed overloads were unused (build passed) and composed on name derivation — leaving them would be stale-path traps [I]. `workspaceFileURL(for:baseRoot:)` throwing is required for the legacy direct save to land in the UUID directory (test 4). | Keep. |
| 2039 cold load `try Self.resolvedWorkspaceFileURL` | KEEP | root | Inside existing `do`; ambiguity now hits the existing catch instead of loading a name-derived path [E]. | Keep. |
| 2256–2268 `resolvedWorkspaceFileURL(for:baseRoot:)` | KEEP | root | Shared by 4 index-driven loaders. | Keep. |
| 2340–2345, 2440–2445, 2486–2491 reload/snapshot/presets | KEEP | root | `print` + `continue` on resolution error; base had no failure mode here. | Keep (three near-identical blocks; folding not worth churn). |
| 3021–3025 switch fence | KEEP + **add** | root / **defect (gap)** | Blocks switch only for `pendingConsolidatedRestoreIDs`. A record with `consolidatedIntoWorkspaceID != nil` (retired, not being restored) is still admissible; MCP `switch` by UUID resolves hidden records explicitly [E: schema text "UUID lookup remains explicit"] → a retired copy can be opened, violating "non-switchable until restored" [I]. | Add `|| workspace(withID: newWorkspace.id)?.consolidatedIntoWorkspaceID != nil` with a "restore it first" message (+3). Add one assertion to test 1 (switch to `projectedRetired` before restore is blocked). |
| 4517–4519 `noteCatalogFileURLs` | KEEP | root | Populates the process-wide resolver memo per projection so Agent/Chat (which pass `catalogFileURL: nil`) resolve by catalog. | Keep. |
| 4535 `reconcilePendingConsolidatedRestoreIDs()` after projection | KEEP | hard | Drops fence entries whose projected model still carries the marker or is fully saved; keeps while `dirtyRevision != nil` [E]. Coherent with the in-flight restore's exact check (a re-projected marker makes the restore fail closed). | Keep. |
| 6430, 6462, 6469, 6836 `deletedWorkspaceIDs`→`retiredWorkspaceIDs` | SHRINK (optional) | rename | Cosmetic field rename on a result struct; consumers are tests only [E]. | Optional revert (−4 changed lines). Low value either way. |
| 6438–6446 backup dir override plumbing | KEEP | test infra | | |
| 6516–6556 authority re-plan (vs legacy disk snapshot) | KEEP | root | Base re-planned from `workspacesIndex.json` in both modes; in authority mode that index is not maintained (`saveWorkspaceIndexAsync` guard) so a stale index silently dropped live records from `workspaces` and skipped consolidation [E]. Head plans from the authority snapshot; failure → `authority_snapshot_unavailable` skips and early return. Tested (test 1 deliberately stales the index). | Keep. |
| 6564–6587 `stablePostSwitchPlans` (`group_changed`) | KEEP (conditional) | hard, **untested** | Base iterated `postSwitchPlans` directly, which could consolidate a group not present in the backup written for `initialPlans` [E]. Head skips such groups. Fail-safe direction, but no regression test. | Keep only with a test, else drop (−24). Acceptable to keep given conservative failure mode; flag. |
| 6642–6658 canonical dir resolve → `sidecar_preflight_failed` | KEEP | root | | |
| 6659–6705 prepare both → commit both | KEEP | root | Both families prepared before either writes; collision/malformed detected in prepare [E]. Commit-phase I/O failure can still leave one family partially copied (disclosed). | Keep. |
| 6706–6718 `canonical_changed` guard | **SHRINK** | hard (redundant) | Same condition is re-checked at 6734–6740 (code 6, inside the `do`) and again by the exact save (code 8) [E]. Three layers. | Remove this first layer (−13); the code-6 guard covers both modes and reports via the catch. |
| 6720–6727 prepublish `merged` | KEEP | root | Required (a) so the exact-save precondition `current == workspace` can hold, (b) so the bridge caches the outbound model under the committed digest. | Keep. |
| 6729–6762 flush / code-6 guard / save / catch (reconcile or legacy rollback) | KEEP | root | Authority: full re-projection instead of local rollback (prevents poisoning the bridge cache) [E]; legacy: rollback if unchanged. | Keep. |
| 6764–6776 canonical existence guard | KEEP | minimal diff from base | | |
| 6777–6805 retirement loop; per-group full-retirement count | KEEP / SHRINK (optional) | root / accounting delta | Authority → `persistDuplicateWorkspaceRetirement`; legacy → base `preserveDuplicateWorkspaceStorage`. `groupsConsolidated` now counts only fully retired groups (base counted any commit) — unrequested, untested accounting change [E]. `workspaces.removeAll` of retired rows in authority mode is churn (projection re-adds them) but harmless. | Optional: revert accounting to base (`+= 1` when `!retiredInGroup.isEmpty`) (−3). |
| 6819 `!retiredWorkspaceIDs.isEmpty` | KEEP | consequence of above | | |
| 6842–6870, 6872–6893, 6895–6925 authority projection helpers | KEEP (SHRINK optional) | root | `load…` and `reconcile…` duplicate snapshot+decode; empty-catalog semantics deliberately differ. | Optional: let `reconcile…` reuse `load…` with an `allowEmpty` flag (−10). |
| 6945–6948 eligibility excludes marked records | KEEP | root | | |
| 7126–7159 backup: dir override, UUID suffix, `.withoutOverwriting` | KEEP | test infra + hard | UUID suffix removes same-second overwrite of an earlier backup; `.withoutOverwriting` is then redundant and drops `.atomic` [E]. | Keep UUID; use `[.atomic, .withoutOverwriting]` or `.atomic` (1 token). |
| 7176–7181 comment | KEEP | | Comment now accurate. | |
| 7183–7235 `persistDuplicateWorkspaceRetirement` | KEEP | root | Marks `isHiddenInMenus=true`, `consolidatedIntoWorkspaceID`, prepublishes, exact save, reconcile/rollback on failure. Guards (identity, marker nil, root-key equality) are reasonable for a marking write. | Keep; could be `private` (nit). |
| 7279–7286 `duplicateCleanupPersistenceFailureReason` | KEEP (optional) | hard (reporting) | Distinguishes `persist_incomplete_working_committed` [E]. Untested at this level. | Keep or drop with 316–322 as a unit. |
| 7367–7381, 7507–7509, 7533–7548, 7551–7577 deletion: resolve dir up front, thread `workspaceDirectory:` param, fail closed | **SHRINK** | root + hard | Required part: delete the **resolved** dir (base deleted the name-derived dir, orphaning a renamed workspace's real directory in authority mode [E]). Head additionally changes the signature, three call sites, and refuses deletion entirely on ambiguity (legacy/ephemeral) [E]. Base computed the dir **after** removing `domainWorkspaceFileURLsByID[id]`; the head's `workspaceDirectory(for:)` would then fall to the catalog memo — still correct. | Smallest: in `finalizeWorkspaceDeletion` compute `let workspaceDir = workspaceDirectory(for: workspace)` as the **first** statement (before map removal) and keep everything else as base (−25). Keep `pendingConsolidatedRestoreIDs.remove` + `resolver.evict`. Head's fail-closed is defensible for a destructive path; if kept, add a test. |
| 7590–7597 removed legacy folder move on rename | KEEP | root-adjacent | Legacy-only (`domainWorkspaceAuthorityClient == nil`) [E]. Move + resolver would also work, but a failed move in base already produced a split; with UUID resolution the move is unnecessary. Legacy mode is the isolated test owner [E]. | Keep. Note: `HistorySessionScanner` name parsing will show stale names for legacy renames (already true in authority mode). |
| 7641–7702 `setWorkspaceHidden` | **SHRINK** / **defect (minor)** | root + dead/harmful code | Early branch routes consolidated unhide to the fenced snapshot path — good [E]. The ordinary path then also clears `consolidatedIntoWorkspaceID` (7655–7657, 7675–7677, 7681) — unreachable unless the passed model is stale relative to memory, in which case it is exactly the unfenced clear baron #1 flagged [I]. | Delete the three ordinary-path marker clears (−7). Extend the early branch to `|| pendingConsolidatedRestoreIDs.contains(workspace.id)` (+1) so the view can revert to its base one-liner. |
| 7704–7763 `setWorkspaceHiddenFromSnapshot` | **SHRINK** / **defect** | root for restore; **regression for ordinary MCP hide/unhide** | Base: field-wise update or `append` if absent; ordinary save with CAS recovery. Head, for **all** callers: throws code 9 unless `workspaces[index] == workspace`; replaces the element; save source is now exact/fail-closed via `requiresExactSnapshotAttempt` [E]. MCP `hide`/`unhide` pass a disk-loaded model (`resolveWorkspaceForHiddenMutation` → disk snapshot) [E]; for a dirty (working≠saved) workspace, e.g. the active one, that model ≠ memory → MCP hide now fails where base succeeded [I]. No test covers ordinary hide/unhide. Violates the stated preservation invariant and the PR body's "ordinary … save conflict recovery is unchanged". | Compute `isConsolidatedRestore` first; for `false`, run the **base body verbatim**; for `true`, run head's fenced/exact body with a new source `.consolidatedWorkspaceRestore` (and `preserveDiskRepoPathsIfUnchangedSinceBaseline: false` for consistency). Route only that source to exact mode. |
| 7765–7771 `reconcilePendingConsolidatedRestoreIDs` | KEEP | root | Needed by the failure path. | |
| 7773–7783 `applyWorkspaceHiddenStateInMemory(+consolidatedIntoWorkspaceID:)` | KEEP | root | Parity with base's immediate cross-manager sync for `hidden`. | |
| 9303–9340 pre-check / exact gate / early CAS pin | KEEP except **9325–9333** | root / **unrelated semantic delta** | Exact gate + early pin are required for fail-closed cleanup saves [E]. The `else if let current, current != workspace { retry with current }` branch changes **ordinary** saves: any caller passing a model that differs from memory now silently saves the in-memory model (or throws `CancellationError` on the second mismatch) instead of the passed payload [E]. No test; not needed by cleanup/restore (both use the exact branch). | Delete 9325–9333 (−9). Early pin for the ordinary branch is harmless given durable replay semantics; keep. |
| 9348–9364 honor `preserveDiskRepoPathsIfUnchangedSinceBaseline` | KEEP | hard/bug-fix | Base ignored the flag in the domain path (always merged disk repoPaths) [E]; head honors it. Affects other `false` callers (only when local == baseline and disk differs) [I]. | Keep; note as an intentional fix. |
| 9381–9385, 9400–9402 exact re-checks | KEEP | root | | |
| 9413–9476 exact → `saveFailClosed` (no `snapshot()`/create discovery), apply phases, `workingCommitted` | KEEP | root | Skipping `exists`→`create` ensures a concurrently deleted record conflicts rather than being recreated [E]. `workingCommitted` second clause (`working == nil && dirtyRevision != nil`) reports "payload already in durable working state" — defensible. | Keep. |
| 9581 `try workspaceFileURL(for:baseRoot:)` | KEEP | root | | |
| 9744–9745 `ensureWorkspaceDirectoryExists` via resolver | KEEP | root | Legacy save resolves twice (URL + ensure); legacy only. | Keep. |

### `Features/Workspaces/ViewModels/WorkspaceSaveDiagnostics.swift`

| 44 `.duplicateCleanupRetireDuplicate` | KEEP | root | Needed for exact-mode routing and tracing. | Add `.consolidatedWorkspaceRestore` (+1) if the hidden-snapshot scoping is adopted. |

### `Features/Workspaces/Views/ManageWorkspacesView.swift`

| Range | Mark | Class | Delta | Risk / patch |
|---|---|---|---|---|
| 16 `showConsolidatedWorkspaces` | KEEP | root UI | | |
| 436–458 skip-reason → guidance mapping | **SHRINK** | UI bloat | ~20 lines of `hasPrefix` matching on a stringly protocol; base sentence ("active sessions or failed switches") is now wrong, so *some* change is warranted [E]. | Replace with one generic sentence: "… skipped; skipped duplicates and their session history were left unchanged — review the remaining entries and retry." (−18). Optionally append distinct reason tokens (+2). |
| 476–486, 549–550, 575–577 partition ordinary/consolidated; counts; empty state | KEEP | root UI | Retired or pending rows leave the ordinary list. | |
| 583–608 `consolidatedWorkspacesSection`, 611–618 `canonicalWorkspaceName` | KEEP (optional shrink) | root UI | Collapsed section, `onSwitch: nil`, eye restores. The "Consolidated into X" caption + helper are nice-to-have (−12 if cut). | Keep. |
| 742–744, 757–761, 797 selection-mode badge + a11y | KEEP | UX hard | Consolidated recovery copies are selectable for bulk delete; labeling them prevents accidental deletion. | Keep (small). |
| 949–958 `toggleHiddenState` routing | **SHRINK** | duplicated routing | Duplicates the manager's `setWorkspaceHidden` early branch, adding the pending check there instead of in the manager [E]. | Revert to base one-liner (−8) after moving the pending check into `setWorkspaceHidden`. |

### `Features/Workspaces/Views/OptimizedWorkspaceRow.swift`

| 6, 24–31 optional `onSwitch` | KEEP | root UI | Existing callers pass non-optional closures unchanged. | |

### `Features/Workspaces/WorkspaceModel.swift`

| 68 rename | SHRINK (optional) | see above | | |
| 355–360, 421, 443, 465, 512, 556, 577 `consolidatedIntoWorkspaceID` | KEEP | root | Optional field; `encodeIfPresent`/`decodeIfPresent` → absent key decodes `nil` (legacy compatible); included in `==` (required for exactness) [E]. Test 3 covers round-trip. | Keep. |

### `Features/Workspaces/WorkspaceSessionSidecarMigration.swift` (new)

| Range | Mark | Class | Delta / trace | Risk / patch |
|---|---|---|---|---|
| 1–29 types + errors | KEEP | root | | |
| 31–44 `sessionFileURLs` | KEEP | root | Regular `.json` files with prefix, sorted. | |
| 46–108 `prepareCopies` | KEEP | root | Per file: session ID from filename; normalize (`workspaceID`, `fileURL` rewritten; `sortedKeys` re-serialization); existing destination → normalized compare → `divergentCollision` on mismatch, skip if already rehomed, else rewrite destination [E]. Whole-document re-serialization via `JSONSerialization` (key order/number formatting may change) — decodable fields verified by test 1 [E]; number-format drift is a theoretical risk [I]. Chat sessions gain a `fileURL` key they never had [I from `ChatSessionHeader`]. | Keep. Nit: only set `fileURL` when the key exists (+1 guard) to avoid polluting Chat sidecars. |
| 110–118 `write` | KEEP | root | Atomic + mtime. | |
| 120–166 normalization/ID parsing | KEEP | root | | |

### `Features/Workspaces/WorkspaceStorageDirectoryResolver.swift` (new)

| Range | Mark | Class | Delta / trace | Risk / patch |
|---|---|---|---|---|
| 4–17 errors | KEEP | root | | |
| 22–37 class, `Provenance`, `MemoEntry`, lock, cap | **SHRINK / DELETE** | dead code + **defect** | `compatibleMemo` returns `nil` for `.custom` and `.discovered` [E] → those memo entries are written but never read; the three-case provenance and `MemoEntry` are dead weight. `maximumRootEntryCount = 1024` (110–116) throws `scanFailed` when the root has >1024 entries → **every** index-driven load (`cold load`, reload, snapshot, presets) skips **every** workspace, and Agent/Chat resolution fails [E]. Leaked-fixture tooling in this codebase indicates dev roots with many stray directories are realistic [I]. | Replace with `catalogDirectoryByWorkspaceID: [UUID: URL]`; delete the cap (−7); delete Provenance/MemoEntry/`store` (≈−25). |
| 39–63 custom / catalog / memo fast paths | SHRINK | root | Storing on the `catalogFileURL` path duplicates `noteCatalogFileURLs`. Memo hit is not root-scoped → a catalog memo from the manager's root redirects a service resolving in a different root [I, see services]. | Return `catalogFileURL` dir without storing (−3). Memo hit only if `memo.deletingLastPathComponent() == standardizedRoot` (+1). |
| 65–135 scan: root checks, enumerate, per-child UUID/dir/`workspace.json`, ambiguity, derived fallback | KEEP | root | Full scan before choosing; ambiguity throws; no match → derived creation path [E]. Tested (test 4, current-name + old-name). Per-save rescans for non-catalog workspaces (no discovered memo) are O(root entries) [I] — perf note only. | Keep. Simplify `containsWorkspaceDocument` to `fileExists(atPath:)` (−10) and `isDirectory` to `fileExists(atPath:isDirectory:)` (−6) — optional. |
| 137–156 `noteCatalogFileURLs` | SHRINK | root | Evict-stale-then-upsert equals replace-all for a single-process catalog [I]. | `catalogDirectoryByWorkspaceID = map.mapValues{…}` (−8). |
| 158–162 `evict` | KEEP | root | | |
| 164–170 `removeAllForTesting` | **DELETE** | unused | Not referenced by the test file [E]. | −7. |
| 172–185 `compatibleMemo`, 187–191 `store`, 193–216 `containsWorkspaceDocument` | SHRINK | see above | | |

### `Infrastructure/MCP/AppShared/DomainWorkspacePresentationBridge.swift`

| Range | Mark | Class | Delta / trace | Risk / patch |
|---|---|---|---|---|
| 14–25 `DomainWorkspaceFailClosedSaveOutcome` | KEEP | root | Phase outcomes; `workingCommitted` derived from `working?.isSuccessfulDomainMutation` [E]. | |
| 119–152 `saveFailClosed` | **SHRINK** | root, duplicated infrastructure | Copy of `save` with `conflictRecoveryPolicy: .failClosed` on both envelopes and phase-returning; extra `?? saveRevision` fallback [E]. | Extract one private `savePhased(_, fileURL:, expectedWorkspaceRevision:, expectedContentDigest:, policy:, operationIDs:) -> DomainWorkspaceFailClosedSaveOutcome`; `save` returns `.finalOutcome ?? working`, `saveFailClosed` returns the struct (≈−20). |
| 172 `fileprivate document` | KEEP | root-adjacent | Needed by the echo digest check. | |
| 259–261 `suppressSelfEchoForTesting` | KEEP (with 334–357) | test hook | | |
| 330 `workspaceSnapshot`→`canonicalWorkspaceSnapshot` | KEEP (evidence-light) | hard | Digest compared should be the canonical committed digest; excerpts don't show the difference between the two APIs [I]. | Ask author for one-line evidence. |
| 334–357 digest-verified cache | KEEP (optional split) | hard (general bridge defect) | Base cached the *current local model* under the committed digest even when they differed [E]; head encodes the local model, compares digests, and on mismatch decodes the canonical bytes (never pushes the older model to the manager). With cleanup's prepublication this path is not required for the PR outcome [I], but it is a tested fix of the same resurrection mechanism. Cost: one JSON encode per suppressed self-echo [E]. | Keep (tested); acceptable to split into its own PR. |

### `Infrastructure/MCP/WindowRoutingService.swift`

| 2611 pass `consolidatedIntoWorkspaceID` | KEEP | root | Signature consequence. | |

### `RepoPromptDomainRuntime/DomainWorkspaceCommand.swift`

| 33–37 enum, 43/51/60 envelope field, 69–71 fingerprint | KEEP | root | Optional Codable field → old envelopes decode with `nil` [E]; fingerprint gains a component only when set (existing fingerprints unchanged) [E]. | |

### `RepoPromptDomainRuntime/DomainWorkspaceContextAuthority.swift`

| Range | Mark | Class | Delta | Risk / patch |
|---|---|---|---|---|
| 380–389 `fail_closed_requires_workspace_revision` | KEEP | hard | Makes fail-closed meaningful (no revision → no CAS). | |
| 396–402 `saveWorkspace(allowsCASRecovery/ExternalRecovery: policy != .failClosed)` | KEEP | root | Default path unchanged (`true`) [E]. | |
| 1548–1561 replaceWorkingDocument fail-closed conflict | KEEP | root | Returns conflict instead of `isDurableReplay` continue; default path unchanged [E]. | |
| 1693–1695, 1731–1733 diagnostic wording by policy | KEEP | hard (accuracy) | Non-fail-closed strings unchanged [E]. | |

### `Tests/RepoPromptTests/Workspaces/WorkspaceDuplicateCleanupTests.swift` (new)

| Test | Mark | Covers | Note |
|---|---|---|---|
| 1 `testAuthoritativeRetirement…` (~L40–320) | KEEP | older-echo cache; stale legacy index re-plan; retirement marker in authority; merged tabs; sidecar copies rewritten; originals preserved; unrelated projection doesn't resurrect; reload keeps marker; detection empty; fence blocks switch during restore; restore completes; detection resumes | Integration test; splitting would duplicate ~100 lines of runtime setup. Add one assertion: `requestWorkspaceSwitch(to: projectedRetired)` is blocked **before** restore (marker gap). |
| 2 `testConcurrentAuthorityUpdateWins…` (~L322–430) | KEEP | early CAS pin, fail-closed, reconcile baseline, external winner stands, duplicate untouched | Regression for the exact-save path. |
| 3 `testWorkspaceRetirementMarkerRoundTrips…` (~L432–452) | KEEP | Codable + legacy nil default | Cheap; not duplicative enough to cut. |
| 4 `testRenameSafeStorageResolution…` (~L454–520) | KEEP | UUID scan; legacy direct save to old dir; catalog URL wins; ambiguity throws; Git-data refuses | Regression for the rename fix. |
| 5 `testBothSidecarFamiliesPreflightBeforeEitherWrites` (~L522–566) | **REPLACE** | Only proves `prepare` has no side effects (trivially true) by calling both prepares directly; the coordinator (6659–6705) is untested [E] | Rewrite as a **legacy-manager** cleanup (no runtime): duplicate with a valid Agent sidecar + malformed Chat sidecar → assert `skipped` has `sidecar_preflight_failed`, no Agent copy in canonical dir, duplicate still present, `groupsConsolidated == 0`. Same size. |
| 6 `testFailClosedSaveReportsAWorkingOnlyCommit` (~L568–630) | KEEP (paired with reporting) | bridge `workingCommitted` via operation-ID collision on the saved phase; dirty revision persists | Manager-level `persist_incomplete_working_committed` label remains untested; drop this test only if the reporting is dropped. |
| helpers + `SuspensionGate` | KEEP | | |

Missing regression tests for intentional deltas: `stablePostSwitchPlans`; deletion fail-closed on ambiguity; ordinary-save `current != workspace` substitution (recommended deletion instead); `setWorkspaceHiddenFromSnapshot` exactness for ordinary callers (recommended scoping instead).

---

## Pass 2 — Adjudication of PR rationale and prior findings

**PR body claims**
- "persist duplicate consolidation as a recoverable workspace retirement" — verified [E: marker field, `persistDuplicateWorkspaceRetirement`, test 1 reload/projection assertions].
- "preflight both sidecar families before either family writes" — verified for collision/malformed detection [E: 6659–6705]; commit-phase I/O failures can still be one-family partial (disclosed as `sidecar_migration_failed`). Test 5 does not exercise the coordinator.
- "report fail-closed saves by phase … reconcile from authority" — verified [E]; manager-level label untested.
- "fence consolidated-workspace restoration" — verified for the pending window [E: 3021, test 1]; **gap**: marker-set records remain admissible to `requestWorkspaceSwitch` (MCP UUID switch).
- "resolve workspace storage by stable UUID/catalog identity" — verified [E: resolver, test 4]; **unverified**: the simultaneous relocation of Agent/Chat roots to `configuredRoot()` plus a root-agnostic catalog memo changes where sidecars live for custom-root (and possibly dev-identity) users without migration.
- "Ordinary workspace save conflict recovery is unchanged" — **false**: `.setWorkspaceHiddenFromSnapshot` is now exact/fail-closed for all callers; ordinary saves substitute the in-memory model when the payload differs (9325–9333); the domain path now honors the repoPaths flag.
- "Ambiguous on-disk matches fail closed rather than guessing" — verified for mutation paths [E]; read/flush paths fall back with a log (fine).

**Codex P2 (MCP reachability)**: Head does not remove authority-mode records; in authority mode the legacy index is not rewritten at all (`saveWorkspaceIndexAsync` guard) [E], so rows are not dropped; the retirement marker keeps them discoverable in the authority catalog. Legacy mode still removes rows (base behavior; legacy = isolated test owner [E]). Finding is moot for authority mode as written — but MCP `unhide` now hits the new code-9 equality guard and MCP `switch` can open a retired copy (new findings above).

**baron #1 (restore switchable before durable)**: addressed within one window by the fence + admission check; verified by test 1 [E]. Cross-window: another window becomes able to switch only after the working commit projects, and that commit is journaled/durable and restored on bootstrap [E: DomainPersistence 1078–1098] — acceptable. Residual gap: retired-but-not-restoring records are switchable via MCP UUID switch.

**baron #2 (working-only commit reported as total failure)**: addressed by phased outcome + `workingCommitted` + reconcile [E]; disclosure via `persist_incomplete_working_committed` and UI guidance. Terminal state on restart (merged canonical from journal, duplicate unretired) is now truthfully labeled; re-running cleanup is idempotent for tabs/presets (ID-deduplicated) [E: `mergedCanonicalWorkspace`].

**baron #3 (partial sidecar commit)**: addressed for prepare-detectable failures; commit-phase partials remain possible and disclosed. Test 5 is not a coordinator-level regression.

**Oracle APPROVE / author follow-ups**: full-scan-before-choose and memo-can't-mask-sibling claims verified [E]. Missed: the 1024-entry cap failure mode; dead provenance memo; MCP hide/unhide exactness regression; MCP switch gap; ordinary-save substitution; unverified sidecar root relocation.

---

## 1. Exhaustive file-by-file accounting

| File | +/− | Verdict |
|---|---|---|
| AgentSessionDataService.swift | +80/−15 | KEEP prepare/commit; SHRINK `workspaceFolderURL` (preserve base root; resolver for folder only) |
| WorkspaceDirectoryName.swift | +5/−5 | DELETE/REVERT entirely |
| ChatHistoryManager.swift | +76/−16 | KEEP prepare/commit; SHRINK `workspaceFolderURL` as Agent |
| WorkspaceFilesViewModel.swift | +1/−1 | KEEP |
| WorkspaceManagerViewModel.swift | +735/−142 | KEEP core (marker, planning from authority, sidecar coordination, exact save, retirement, fence, resolver wiring); DELETE `configuredRoot`, ordinary-save substitution (9325–9333), ordinary-path marker clears in `setWorkspaceHidden`, `canonical_changed` guard; SHRINK deletion threading, `setWorkspaceHiddenFromSnapshot` scoping; ADD marker check in `requestWorkspaceSwitch`; optional: rename revert, accounting revert, helper dedupe |
| WorkspaceSaveDiagnostics.swift | +1 | KEEP; +1 `.consolidatedWorkspaceRestore` |
| ManageWorkspacesView.swift | +92/−14 | KEEP partition/section/badge; SHRINK guidance mapping to one sentence; REVERT `toggleHiddenState` |
| OptimizedWorkspaceRow.swift | +9/−5 | KEEP |
| WorkspaceModel.swift | +10/−4 | KEEP marker; optional revert of `retiredWorkspaceIDs` rename |
| WorkspaceSessionSidecarMigration.swift | +166 | KEEP |
| WorkspaceStorageDirectoryResolver.swift | +217 | SHRINK to catalog-only, root-scoped memo; DELETE 1024 cap and `removeAllForTesting`; simplify existence checks |
| DomainWorkspacePresentationBridge.swift | +85/−4 | KEEP fail-closed outcome + echo digest fix; SHRINK `saveFailClosed` into shared phased save |
| WindowRoutingService.swift | +1 | KEEP |
| DomainWorkspaceCommand.swift | +12 | KEEP |
| DomainWorkspaceContextAuthority.swift | +36/−4 | KEEP |
| WorkspaceDuplicateCleanupTests.swift | +761 | KEEP 1,2,3,4,6; REPLACE 5 with a coordinator-level legacy cleanup test; add one blocked-switch assertion to 1 |

## 2. Ranked deletion/shrink list

1. **Resolver `maximumRootEntryCount` cap** — DELETE (correctness: >1024 root entries disables all workspace loading). −7.
2. **`setWorkspaceHiddenFromSnapshot` exactness for all callers** — SHRINK to restore-only; restore base body for ordinary MCP hide/unhide (regression). ±0 lines, semantic preservation.
3. **Ordinary-save `current != workspace` substitution (9325–9333)** — DELETE (unrelated semantic delta, silent payload replacement). −9.
4. **Service root relocation + `configuredRoot`** — SHRINK/DELETE (unverified relocation of Agent/Chat storage; keep base roots; root-scope the memo). ≈−26 changed.
5. **`WorkspaceDirectoryName.swift`** — REVERT (dead for this PR). −10 changed.
6. **Resolver dead provenance memo / `store` / `removeAllForTesting` / verbose existence checks** — SHRINK. ≈−55.
7. **`saveFailClosed` duplication of `save`** — SHRINK via shared phased helper. ≈−20.
8. **Deletion `workspaceDirectory:` parameter threading** — SHRINK to one moved line in `finalizeWorkspaceDeletion`. ≈−25.
9. **`ManageWorkspacesView` guidance mapping** — SHRINK to one sentence. −18.
10. **`setWorkspaceHidden` ordinary-path marker clears** — DELETE (dead/harmful). −7; move pending check into the early branch (+1); revert view `toggleHiddenState` (−8).
11. **`canonical_changed` pre-flush guard** — DELETE (redundant with code-6 and code-8 checks). −13.
12. Optional: `stablePostSwitchPlans` (−24 unless tested); `groupsConsolidated` accounting revert (−3); `retiredWorkspaceIDs` rename revert (−4); `reconcile…` reuse of `load…` (−10); Chat double enumeration (−6).

Additions required: marker check in `requestWorkspaceSwitch` (+3), `.consolidatedWorkspaceRestore` source (+1), root-scoped memo guard (+1), replaced test 5 (≈±0), one assertion in test 1 (+2).

## 3. Line counts and complexity direction

- Current: **+2277 / −220** (2497 changed lines).
- Target after mandatory items (1–11): ≈ **+2090 / −205** (≈2295 changed, −8%); with optional items ≈ **+2040 / −200**.
- Structural complexity: still **up** vs base (new marker, resolver, two-phase sidecar, fail-closed policy are inherent to the outcome), but reduced: one save-path bifurcation instead of two (`saveFailClosed` merged), resolver drops a 3-way provenance state machine, deletion API keeps its base signature, ordinary save/hidden paths return to base control flow. `consolidateDuplicateWorkspaces` remains a ~400-line function (cognitive load high; not a minimality issue to fix here).

## 4. Tests: keep / remove / combine

- **Keep**: 1 (add blocked-switch-before-restore assertion), 2, 3, 4, 6.
- **Replace**: 5 → legacy-manager `consolidateDuplicateWorkspaces` with a malformed Chat sidecar; assert `sidecar_preflight_failed`, no Agent copy written, duplicate not retired, group still detected.
- **Remove**: none outright. If reporting (316–322, 7279–7286, UI branch) is dropped, drop 6 with it.
- **Combine**: none; setup cost argues against splitting 1.
- **Not needed after scoping**: ordinary MCP hide/unhide test (behavior returns to base).

## 5. Risks of each deletion

1. Cap deletion: none functional; scan cost already linear in root entries.
2. Hidden-snapshot scoping: none for restore (head body retained for that path); ordinary path regains base's tolerant semantics (including base's `append`-if-absent quirk).
3. Ordinary-save substitution deletion: reverts to base's pre-existing cross-window race (stale payload wins via durable replay) — pre-existing, not introduced here.
4. Service root preservation: if the owner proves all users' sidecars already live under the catalog root, head's relocation is a no-op and either choice is safe; if not, preservation avoids invisible history. Root-scoped memo: services in a different root fall back to scan (slower, correct).
5. `WorkspaceDirectoryName` revert: none (no producer in this PR).
6. Resolver simplification: none; dead entries were never read. Replace-all memo may transiently drop an entry another window just noted → one extra scan.
7. `saveFailClosed` merge: none if `save`'s working-failure early return is preserved via `.finalOutcome ?? working`.
8. Deletion threading revert: under UUID ambiguity, deletion falls back to the name-derived directory (base behavior) instead of refusing; both directories belong to the same UUID.
9. Guidance shrink: less specific user text; skip reasons remain in the result struct.
10. Marker-clear deletion: none; those lines are unreachable except in the harmful stale-model case.
11. `canonical_changed` deletion: identical terminal state (skip + reconcile), reason string becomes `persist_failed: …changed before merge commit`.
12. `stablePostSwitchPlans` deletion (if chosen): reintroduces base's possibility of consolidating a group absent from the backup.

## 6. Verdict

**trim then ship** — after fixing the two correctness items (1024-entry cap; ordinary MCP hide/unhide exactness regression), closing the MCP switch-to-retired gap, deleting the unrelated ordinary-save substitution, and either reverting or evidencing the Agent/Chat storage-root relocation.

Sources/RepoPrompt/Features/Workspaces/WorkspaceStorageDirectoryResolver.swift
```diff
@@ -0,0 +1,217 @@
...
+                options: [.skipsHiddenFiles]
+            )
+        } catch {
+            throw WorkspaceStorageDirectoryResolutionError.scanFailed(
+                workspaceID: workspaceID,
+                root: standardizedRoot,
+                reason: error.localizedDescription
+            )
+        }
+        guard children.count <= maximumRootEntryCount else {
```
dsebban: This cap makes a root with 1,025 entries fail resolution for every workspace. The index-driven cold-load/reload/snapshot loops then skip all records, so one large or debris-heavy root becomes total workspace disappearance rather than bounded degradation. The PR has no evidence for 1,024 as a safe product limit. Please remove this cap (or design a bounded lookup that does not invalidate unrelated workspaces).

Tests/RepoPromptTests/Workspaces/WorkspaceDuplicateCleanupTests.swift
```diff
@@ -0,0 +1,761 @@
...
+                baseRoot: storageRoot
+            )) { error in
+                guard case WorkspaceStorageDirectoryResolutionError.ambiguousDirectories = error else {
+                    return XCTFail("Expected an ambiguity error, got \(error)")
+                }
+            }
+            XCTAssertThrowsError(try manager.gitDataDirectory(for: renamed))
+        }
+
+        func testBothSidecarFamiliesPreflightBeforeEitherWrites() async throws {
```
dsebban, line 711: This calls the two prepare APIs directly, so it proves only that preparation itself has no writes; it does not regress the coordinator ordering in `consolidateDuplicateWorkspaces` that previously wrote Agent copies before Chat preflight failed. Please drive a manager cleanup with a valid Agent sidecar plus malformed/divergent Chat sidecar, then assert `sidecar_preflight_failed`, no Agent destination copy, no retirement, and zero consolidated groups.

Sources/RepoPrompt/Features/Workspaces/ViewModels/WorkspaceManagerViewModel.swift
```diff
@@ -7291,35 +7702,83 @@ class WorkspaceManagerViewModel: ObservableObject {
...
+                workspaces[rollbackIndex] = original
+                pendingConsolidatedRestoreIDs.remove(original.id)
+            }
+            throw error
+        }
+    }
 
-        return updated
+    private func reconcilePendingConsolidatedRestoreIDs() {
+        pendingConsolidatedRestoreIDs = pendingConsolidatedRestoreIDs.filter { workspaceID in
```
dsebban: This only filters IDs already present in this manager's transient set; it cannot reconstruct a pending restore in another window or after restart. Once the working phase durably publishes a marker-cleared dirty document, those managers have an empty set and expose the row before the saved phase/index flush. Please derive this state from authority working-vs-saved evidence (for example, dirty marker-cleared working data while the saved file still carries the consolidation marker) and revalidate it at switch admission.

Sources/RepoPrompt/Features/Workspaces/ViewModels/WorkspaceManagerViewModel.swift
```diff
@@ -6610,12 +6833,97 @@ class WorkspaceManagerViewModel: ObservableObject {
...
             reassignedWindowIDs: reassignedWindowIDs.sorted(),
-            deletedWorkspaceIDs: deletedWorkspaceIDs.sorted { $0.uuidString < $1.uuidString },
+            retiredWorkspaceIDs: retiredWorkspaceIDs.sorted { $0.uuidString < $1.uuidString },
             skipped: Self.deduplicatedSkippedItems(skipped),
             backupURL: backupURL
         )
     }
 
+    /// Reloads the cleanup planner from the same owner that will persist its decisions.
+    /// The legacy index is intentionally not updated in authority mode and may omit live records.
```
dsebban, line 7245: The test intentionally makes the legacy index omit the duplicate before cleanup. Authority re-planning correctly finds and retires it, but authority mode never repairs that index, while MCP listing and exact UUID resolution still use `loadWorkspaceSnapshotFromDisk()`. The recovery record therefore remains unreachable through MCP in this stale-index case. Please refresh the compatibility index atomically from this authority snapshot, or move MCP inventory to the owning authority, and add an `include_hidden`/UUID unhide regression.

Sources/RepoPrompt/Features/Workspaces/WorkspaceSessionSidecarMigration.swift
```diff
@@ -0,0 +1,166 @@
...
+        guard !sourceFiles.isEmpty else { return [] }
+
+        var prepared: [WorkspaceSessionSidecarPreparedCopy] = []
+        prepared.reserveCapacity(sourceFiles.count)
+        for sourceURL in sourceFiles {
+            let sessionID = try sessionID(
+                from: sourceURL,
+                filenamePrefix: filenamePrefix
+            )
+            let destinationURL = destinationFolder.appendingPathComponent(sourceURL.lastPathComponent)
```
dsebban: Please reject source/destination storage aliasing before normalization. Two duplicate records can resolve to the same custom/catalog directory; then `destinationURL == sourceURL`, preparation treats the existing source as a destination, and commit rewrites the promised recovery original to canonical identity. At minimum compare standardized/resolved physical directories and fail before either family writes.

Sources/RepoPrompt/Features/Workspaces/WorkspaceSessionSidecarMigration.swift
```diff
@@ -0,0 +1,166 @@
...
+                    data: normalizedSource.data,
+                    modificationDate: modificationDate
+                ))
+            }
+        }
+        return prepared
+    }
+
+    static func write(_ prepared: WorkspaceSessionSidecarPreparedCopy) throws {
+        try prepared.data.write(to: prepared.destinationURL, options: .atomic)
```
dsebban: Preparation and commit are separated by awaits/queue turns. `.atomic` replaces the destination unconditionally, so a canonical session updated after preflight can be overwritten with the stale prepared bytes. Carry the expected destination state into the batch and verify it at commit, or use no-overwrite creation for destinations that were absent; fail closed if an existing destination changed.

Sources/RepoPrompt/Features/Workspaces/WorkspaceStorageDirectoryResolver.swift
```diff
@@ -0,0 +1,217 @@
...
+            do {
+                guard try child.resourceValues(forKeys: [.isDirectoryKey]).isDirectory == true else { continue }
+            } catch {
+                throw WorkspaceStorageDirectoryResolutionError.scanFailed(
+                    workspaceID: workspaceID,
+                    root: standardizedRoot,
+                    reason: error.localizedDescription
+                )
+            }
+            if try containsWorkspaceDocument(child, workspaceID: workspaceID, root: standardizedRoot) {
```
dsebban: Requiring `workspace.json` rejects a legitimate legacy sidecar-only directory. Before this PR, Agent/Chat history could live under the identity application-support root while the workspace document lived under a global/custom root; UUID discovery must reuse that directory to avoid splitting history. A single UUID-bearing directory should be usable even when it contains only `AgentSessions`/`Chats`; multiple matches can still fail closed.

Sources/RepoPrompt/Features/Workspaces/ViewModels/WorkspaceManagerViewModel.swift
```diff
@@ -6823,16 +7137,24 @@ class WorkspaceManagerViewModel: ObservableObject {
...
         formatter.dateFormat = "yyyyMMdd-HHmmss"
-        let fileURL = directory.appendingPathComponent("workspace-dedup-\(formatter.string(from: backup.createdAt)).json")
+        let fileURL = directory.appendingPathComponent(
+            "workspace-dedup-\(formatter.string(from: backup.createdAt))-\(UUID().uuidString).json"
+        )
         let encoder = JSONEncoder()
         encoder.outputFormatting = [.prettyPrinted, .sortedKeys]
         encoder.dateEncodingStrategy = .iso8601
-        try encoder.encode(backup).write(to: fileURL, options: .atomic)
+        try encoder.encode(backup).write(to: fileURL, options: .withoutOverwriting)
```
dsebban: The UUID suffix removes the same-second collision, but this also drops the predecessor's atomic write. A crash can now leave a truncated recovery backup. Please retain no-overwrite behavior while restoring atomicity, e.g. `[.atomic, .withoutOverwriting]`.

dsebban (commented):
## Fable-priority synthesis

Primary review: https://github.com/repoprompt/repoprompt-ce/pull/915#pullrequestreview-5082924411

**Verdict: trim then ship, with changes requested before merge.** Fable is the primary judgment. Sol Pro additions are accepted only where the current source proves a boundary Fable missed.

### Required before merge

1. **Make switch admission recovery-aware, not just instance-set-aware.** Keep Fable's marker check: reject a workspace when the passed or current authoritative model still has `consolidatedIntoWorkspaceID`. Also accept Sol's stronger finding: `pendingConsolidatedRestoreIDs` is local to one manager and is only filtered, never reconstructed. After the working phase clears the marker, another window—or a restarted manager bootstrapped from the dirty journal—has an empty set and can expose/switch the row before the saved phase and index flush. Derive pending restore state from authority working-vs-saved evidence (dirty marker-cleared working document while the saved file still carries the marker), and revalidate at switch admission. Add a second-manager and restart/working-only regression.

2. **Scope exact fail-closed persistence to the feature-owned operations.** Preserve the exact path for canonical merge, duplicate retirement, and consolidated restore. Restore the predecessor `setWorkspaceHiddenFromSnapshot` behavior for ordinary MCP hide/unhide, using a distinct save source for consolidated restore. Delete the new generic ordinary-save payload substitution at 9325–9333 unless it is split into a separately reproduced/tested fix. This is Fable's brown-field finding and remains the smallest safe shape.

3. **Fix storage compatibility without relocating history.** Preserve the pre-change Agent and Chat roots; change only mutable-name directory selection to UUID discovery. Resolve before creating any root so an explicit `customStoragePath` cannot fail because an unrelated global root is unavailable. Root-scope the catalog memo, remove the arbitrary 1,024-entry all-workspaces failure, delete dead custom/discovered memo state, and allow a single UUID-bearing legacy directory that contains only sidecars (no `workspace.json`) to be reused. Revert `WorkspaceDirectoryName.swift`: its builder has no current callers, so that cross-module/comment rewrite is not part of the causal edit.

4. **Close the remaining sidecar safety holes, but do not build a speculative transaction framework.** Keep the prepare-both-before-commit coordinator. Add an early physical source/destination alias rejection so migration cannot rewrite the recovery source. At commit, verify the destination state observed during preparation (or create absent destinations with no-overwrite semantics) so a canonical update between prepare and commit is not replaced. Retain explicit partial-commit reporting; do not add a broad retry queue or migration framework. Replace the current preparation-only test with a manager-level cleanup regression proving Chat preflight failure writes no Agent destination and retires nothing.

5. **Repair stale-index MCP recovery.** The focused test deliberately rewrites `workspacesIndex.json` to contain only the canonical row. Cleanup re-plans from authority, but authority mode never repairs that index, while MCP listing/UUID resolution still calls `loadWorkspaceSnapshotFromDisk()`. Either atomically refresh the compatibility index from one bootstrapped authority snapshot at the cleanup seam, or route MCP inventory to the owning authority. Test `include_hidden` list plus UUID unhide/restore against the stale-index setup.

6. **Preserve backup crash integrity.** Keep the UUID filename, but restore atomic writing (`[.atomic, .withoutOverwriting]` or the equivalent). The unique suffix solves collisions; it does not justify dropping atomicity.

### Fable reductions to keep

- Simplify string-prefix UI guidance to one accurate generic sentence.
- Move pending restore routing into the manager and revert duplicated view routing.
- Delete ordinary-path consolidation-marker clears that are unreachable except for stale/harmful input.
- Delete the redundant pre-flush `canonical_changed` guard; later exact checks retain the terminal state.
- Remove dead resolver provenance/store/testing surface and duplicate sidecar enumeration where practical.
- Revert the unnecessary `private`→`fileprivate` change to `document(for:fileURL:)`.

### Recommendations not adopted

- **Do not** refactor existing `save` into a shared phased helper merely to remove the new `saveFailClosed` duplication; that rewrites predecessor control flow and violates the brown-field rule.
- **Do not** weaken fail-closed destructive deletion to save parameter-threading lines. Ambiguous storage must not be guessed for deletion.
- **Do not** replace the focused suite with Sol's proposed 15-test/transaction-framework expansion. Keep tests 1/2/3/4/6, add the switch/cross-window/restart/stale-index assertions, and replace test 5 with the coordinator-level regression.
- Treat non-authority retirement persistence as a **supported-mode clarification/follow-up** unless production evidence shows a user-reachable nil-authority path; do not expand this already-large PR on inference alone.

### Target footprint

Current: **+2,277 / -220**. A credible target remains close to Fable's **~+2,050–2,100 / -200** range after the required correctness additions—not Sol's speculative +1,540 target. Structural complexity should fall by returning ordinary hide/save paths to predecessor behavior, simplifying resolver state, and avoiding new migration abstractions.

### Maintainer-guidance check

- **User impact/invariant:** consolidation must preserve recoverable workspace/history state across windows and restart, while retired copies remain discoverable but unswitchable until the exact restore is saved and indexed.
- **Root-cause confidence:** confirmed for marker-blind switching, local-only restore fencing, stale MCP index, custom-root relocation/precreation, aliasing, 1,024-entry failure, and ordinary hide/save broadening.
- **Authority:** `DomainWorkspaceContextAuthority` owns workspace truth; MCP compatibility inventory and presentation state must project it without becoming competing authorities.
- **State-safety risks:** working-only partial success, source-sidecar mutation, stale destination overwrite, stale compatibility index, and non-atomic backup write.
- **Scale/observability:** remove the hard all-or-nothing root cap; retain explicit partial-save/migration reasons.
- **Recommended scope:** fix the six required items now; stage unsupported direct-mode behavior and broader transactional migration as follow-up investigation.
- **Validation boundary:** focused cleanup tests, coordinated product build/lint, then live CE MCP `include_hidden` list/switch/unhide against a stale-index retired workspace and a two-window working-only restore probe.

crstdr:
Reverted to draft as I found some issues with the PR while testing

baron (dismissed):
Re-reviewed current PR head `f08f538f59322f90f72b9a57f29293333922e9ad` in hosted merge `67cf3e1a44c45e824db111e2e63cba09b85cc9f8`, whose first parent is current `origin/main` `96b9c9cb4fa9e4f53712df3796fc454c60828a68`.

The restore-before-switch issue is fixed: switch/activation now revalidates retired and working-only restore state across managers and restart. The fail-closed two-phase save issue is also fixed for reporting: the working and saved outcomes remain distinct, working-only success is propagated, and cleanup/UI disclose partial completion.

One P2 remains before approval: the two sidecar families are prepared before either writes, but they are still committed serially—Agent first, then Chat (`WorkspaceManagerViewModel.swift:7052-7068`). Each family revalidates only at its own commit boundary (`WorkspaceSessionSidecarMigration.swift:221-278`). If the Chat destination changes after preparation, Chat correctly rejects the stale batch, but Agent files and its rebuilt index have already been published and are not rolled back. The same partial state can result if Agent file publication succeeds and its later reconciliation/index rebuild throws. This is independently tracked by the current unresolved thread at https://github.com/repoprompt/repoprompt-ce/pull/915#discussion_r3923267747.

Please provide a rollback/staged publish boundary across the two families, or otherwise make the partial state non-visible and recoverable before treating the duplicate as skipped. Add a coordinator-level regression that forces the later Chat commit to fail after Agent commit and proves no Agent destination remains visible.

Fresh local validation: `make dev-test FILTER=WorkspaceDuplicateCleanupTests` passed 6/6 on the exact merge. Those tests cover initial cross-family preflight and per-batch stale destination rejection, but not the later-Chat-failure-after-Agent-commit path. `git diff --check HEAD^1..HEAD` also passed. Hosted CI is still in progress.

crstdr:
Updated in `a27b7645`, addressing the remaining P2 from the review of `f08f538f` (and the same finding on the inline thread).

### The family commit boundary

The two families still commit serially, but the Agent batch is now withdrawn when the later Chat commit fails, before the duplicate is reported as skipped. A committed batch is rolled back from the data it already carries: destinations this run created are removed, and destinations it overwrote are restored to the bytes captured during preparation.

Only bytes this batch published are withdrawn. A destination holding anything else belongs to a newer writer, so it is left alone and returned to the coordinator, which reports `sidecar_migration_partial:` instead of claiming a clean skip. The withdrawal is a compensating undo of one prepared batch — no transaction framework, no retry queue, and source sidecars, backups, idempotent retry, and partial-success reporting are unchanged.

The same withdrawal also covers the second case in the review: a failure inside the Agent commit's own post-publication steps, where the deletion checks or the metadata index rebuild can throw once the files are already on disk.

### Regression

`testChatCommitFailureWithdrawsTheAlreadyCommittedAgentBatch` drives `consolidateDuplicateWorkspaces` with a canonical `Chats` path that passes Chat preflight and fails the Chat commit, which is the exact ordering in question. It asserts the destination Agent session file is gone, that `AgentSessionIndex.json` no longer advertises it, that the source survives, and that the duplicate is neither retired nor removed from detection.

I verified it is not vacuous: with the withdrawal disabled it fails on both the published file and the metadata index.

### Validation

Full `make dev-test` (0 failures), `make dev-lint` strict, `make dev-swift-build PRODUCT=repoprompt-mcp`, `git diff --check`. Diff for this commit is +209/−13 across three sources and one test; the suite stays at 7 tests.

### Remaining scope

To avoid discovering the boundary incrementally, here is what I believe is deliberately **not** in this PR, and I would like to confirm it is the complete remaining set:

- Intra-family commit-phase I/O partials, which remain recoverable, non-destructive, and disclosed.
- Workspace storage identity resolution for renamed workspaces (the reverted rename expansion). This is a separate PR; on this branch storage stays name-derived exactly as on `main`, so the behaviour is unchanged rather than fixed here.
- Non-authority (legacy/direct-mode) retirement persistence, per the earlier suggestion to treat it as a supported-mode clarification.

### Note on the earlier synthesis review

Several items in the 2026-09-01 review were made moot by `567ca49e`, which removed the rename storage expansion before that review's follow-ups landed. `WorkspaceStorageDirectoryResolver.swift` no longer exists on this branch, so the 1,024-entry cap, the memo/provenance state, the Agent/Chat storage-root relocation, and the `WorkspaceDirectoryName.swift` rewrite are all gone with it. Of the rest: the consolidated restore now uses its own save source, and the cleanup backup is written atomically without overwriting. Happy to walk through any item that still looks open.

PR #924, opened 2026-09-01
Reuse matching workspaces when opening folders

### LINKED ISSUE #917: Open Folder from workspace landing creates a duplicate instead of reopening a matching workspace
## Problem

`Open Folder` on the workspace landing screen always creates a workspace, even when an existing workspace already uses the selected folder as one of its roots.

This produces duplicates such as `Project` and `Project (1)`. The original workspace keeps its compose tabs, selections, prompts, presets, and settings. The user lands in a blank workspace and may think that state was lost.

`Add Folder` inside an active workspace is a separate action and is not part of this report.

## Reproduction

1. Create a workspace for a folder.
2. Add some workspace state, such as compose tabs or file selections.
3. Exit to the workspace landing screen.
4. Click `Open Folder`.
5. Select the same folder.

RepoPrompt creates a numbered duplicate instead of reopening the existing workspace.

## Expected behavior

`Open Folder` should reopen an eligible workspace when the selected folder exactly matches one of its roots. It should create a persistent workspace only when no eligible match exists.

If several workspaces match, use the same ordering as Recent Workspaces and resolve equal sort values deterministically.

Users can still create an intentional duplicate through Manage Workspaces → Create Workspace.

## Source confirmation

The landing action passes `.createNewWorkspace`:

- `Sources/RepoPrompt/Features/Workspaces/Views/WorkspaceEntryRootView.swift`

That branch always calls `createWorkspace`:

- `Sources/RepoPrompt/Features/Workspaces/ViewModels/WorkspaceManagerViewModel.swift`

The programmatic folder-opening path already searches workspace `repoPaths` before creating one:

- `Sources/RepoPrompt/App/WindowState.swift`

The interactive and programmatic paths therefore disagree. This report is confirmed by source reading. The macOS UI was not launched for reproduction.

## Acceptance criteria

- [ ] A workspace matches only when the selected folder equals one complete workspace root after canonicalization.
- [ ] Parent and child paths do not match each other.
- [ ] A folder matches a multi-root workspace when it equals any one of that workspace's roots.
- [ ] Eligible matches are persistent, non-system, and visible workspaces. Hidden and ephemeral workspaces do not match.
- [ ] When several workspaces match, RepoPrompt uses the Recent Workspaces ordering with a deterministic tie-breaker.
- [ ] Opening a match uses the normal workspace-switch path and does not create another workspace.
- [ ] Reopening a workspace preserves its compose tabs, selections, prompts, presets, and settings.
- [ ] If switching to a match fails or is cancelled, RepoPrompt does not create a replacement workspace.
- [ ] When no eligible workspace matches, the landing flow creates and activates one persistent workspace.
- [ ] Interactive and programmatic folder opening share path matching and candidate ranking while keeping their existing no-match persistence rules.
- [ ] Landing copy, picker messaging, tooltip text, and icon no longer imply that Open Folder always creates a workspace.
- [ ] Focused tests cover no match, one match, several matches, multi-root workspaces, excluded candidates, normalized path variants, parent and child rejection, switch failure, and workspace-count stability when reopening.

## Scope

This change prevents new duplicates. It does not merge, remove, or choose among duplicates created by the old behavior. Existing cleanup remains a separate operation.

Symlink aliases are outside this issue. Matching should follow the existing canonical root-path policy.

This is also separate from #829, which covers workspaces becoming unavailable after an app restart.

aAtila:
## Summary

Opening a folder now reuses an eligible workspace with an exact matching root, selecting between multiple matches using the same ordering as Recent Workspaces. Interactive landing and programmatic folder-command flows share the same matching policy while preserving their existing no-match persistence behavior.

## Changes

- Add canonical exact-root matching with system, hidden, and ephemeral eligibility rules plus deterministic Recent Workspaces ordering.
- Reuse matching workspaces from both Open Folder entry points without creating replacements after blocked or cancelled switches.
- Preserve command-specific ephemeral creation behavior and allow payloads to apply when the matching workspace is already active.
- Update landing and picker presentation so Open Folder no longer implies that it always creates a workspace.
- Cover resolver policy, interactive orchestration, and programmatic routing with focused regression tests.

## Why

Repeatedly opening a folder could create duplicate persistent workspaces because the landing flow never looked for an existing exact-root match, while programmatic commands used a separate, less precise matcher. A shared policy keeps both paths consistent without changing workspace persistence formats or consolidating existing duplicates.

Closes #917


crstdr:
**Potential blocker: exact-root selection does not use the same ordering as Recent Workspaces.**

Issue #917 explicitly requires multiple matches to use the **same ordering as Recent Workspaces**. The new `WorkspaceExactRootCandidateRank` instead sorts by `dateModified`, then lowercased/exact name, then UUID (`WorkspaceExactRootCandidateRank.swift:16-28`). The landing/menu list sorts by `lastUsed`, then `localizedCaseInsensitiveCompare`, then UUID (`WorkspaceManagerViewModel.swift:13308-13313`).

That means two workspaces sharing a root can appear in one order under Recent Workspaces but `Open Folder` can reopen the other one—for example, when A was opened more recently (`lastUsed`) while B was edited more recently (`dateModified`). The name tie-breakers can also disagree.

Could the authority/app matching rank share the actual Recent Workspaces fields/comparator, and add a regression where `lastUsed` and `dateModified` prefer opposite candidates? `DomainWorkspaceMetadata` currently omits `lastUsed`, so that looks like part of the boundary that needs aligning rather than substituting modification time.

aAtila:
@crstdr Fixed in 453d0691. Open Folder and Recent Workspaces now share one comparator: `lastUsed` descending, localized case-insensitive name, then UUID. The runtime reads `lastUsed` from domain metadata too.

Added regressions where `lastUsed` and `dateModified` favor different workspaces, plus name ties and selection after restart. For older documents, both decoders fall back to the stored modification date, then `.distantPast`, rather than the current time. Original document bytes are preserved.

crstdr:
Re-reviewed the latest head (453d0691) after the follow-up. The shared Recent Workspaces comparator, lastUsed metadata handling, and backward-compatible fallback address the blocker I raised. I found no remaining issues in this review; this is ready to merge from my perspective.

PR #934, opened 2026-09-02
Add Claude Fable Latest and normalize model picker fonts

w-winter:
RPCE's Claude model pickers now track the Fable family consistently with Opus, Sonnet, and Haiku.  In addition to the pinned Fable 5.1 option added by @datawisebets in #932 , this adds Claude Code's rolling `fable` alias, letting the installed Claude Code runtime resolve the current model in that family.

The Model Preset picker also lets every menu row inherit the standard compact macOS menu font, which fixes the inconsistent larger labels.

I've chosen not to add an additional "Fable Latest (1M)" because both Fable 5 and 5.1 run with native 1M context on the Anthropic API and RPCE already records Fable Latest as a 1M model.  Also, although Claude Code accepts `fable[1m]`, that adds nothing on the normal path and would only add clutter to the picker, IMO.

## Summary

- Add Fable Latest across the Claude model pickers
- Clarify pinned Fable 5 + 5.1 metadata, remove deprecated or prone-to-be-deprecated language
- Normalize Model Preset menu-item fonts

## Validation

- `make dev-format`, `make dev-lint`, and the usual preflight checks passed
- E2E validation in the Debug app of selecting and messaging Fable Latest, and verifying `"model":"claude-fable-5-1"` in the corresponding Claude Code session JSONL

### Before
old
old2

### After
new2
new


w-winter:
Realized after submitting that https://github.com/repoprompt/repoprompt-ce/pull/932 was already open.  That LGTM and I approved it.  I'll follow up in this PR with a reduced scope (Fable Latest + picker font size normalization + model description polish) after #932 is merged.

PR #936, opened 2026-09-02
Fix (MCP) - align context routing with active workspace

w-winter:
Found today that an explicit `bind_context` could succeed but then leave routed tools such as `file_search`, `read_file`, and `get_file_tree` failing with an apparently impossible error saying the requested tab was different even though both tab IDs were identical.

The regression came from the active-workspace routing fix in #812 .  `bind_context` correctly resolved `context_id` against active workspaces, while ordinary tool pre-resolution still searched inactive saved workspaces.  When an inactive workspace contained the same tab UUID, RPCE could combine that workspace identity with the explicitly bound window and reject the resulting hybrid route.

This patch gives one-shot routing the same active-workspace authority while continuing to honor an existing explicit binding after its visible workspace changes.  It also restores focused coverage for the binding contracts removed during the broad test cleanup in #908 .

## Summary

- Resolve ordinary `context_id` tool calls from active workspaces only, matching `bind_context`
- Preserve the existing explicit `(window_id, workspace_id, context_id)` binding as the connection authority
- Report the actual conflicting `context_id`, `workspace_id`, or `window_id` instead of reducing every mismatch to a tab conflict
- Restore three deterministic active-workspace binding tests
- Add a socket-backed regression covering a real `bind_context` followed by `workspace_context` on the same MCP connection

## Implementation Notes

- `MCPBindingResolver` now passes the connection ID to candidate collection; its window-affinity rules are unchanged
- `storedBindingCandidate(forContextID:)` remains available for the separate Oracle continuation flow that intentionally searches saved workspaces
- No workspace activation, retry, compatibility path, persistence change, recovery state, or new routing mechanism was added
- The production change is limited to the existing candidate injection point and conflict diagnostic

## Review Approach

- The patch received one focused maintainability review and two complete-patch correctness reviews
- Findings around DEBUG-only test hooks, singleton lifecycle restoration, deterministic fixture failures, and workspace initialization ordering were fixed within the focused test file
- A PR review finding about preserving explicit bindings across visible workspace switches was fixed by extending the existing socket-backed regression

## Validation

Passed locally:

- `make dev-test FILTER=RepoPromptTests.BindContextRoutingAuthorityTests` — 4 tests
- `make dev-test`
- `make dev-format`
- `make dev-lint`
- `make dev-build`
- Standard live CE MCP smoke
- Commit safety preflight, including staged-index secret scanning and repository guardrails

I also reproduced the original end-to-end sequence in the Debug app using the same tab context that previously failed.  Explicitly rebinding it by `context_id` now leaves `file_search`, `read_file`, and `get_file_tree` working while preserving the same binding.

An isolated release-configuration test build produced no error from the new guarded test file, then stopped on existing non-DEBUG compilation failures elsewhere in `RepoPromptTests`.

### DIFF
--- a/Sources/RepoPrompt/Infrastructure/MCP/MCPBindingResolver.swift
+++ b/Sources/RepoPrompt/Infrastructure/MCP/MCPBindingResolver.swift
@@ -35,7 +35,7 @@ struct MCPBindingResolver {
         }
     }
 
-    let collectMatchesForContextID: (UUID) async -> [MCPContextBindingMatch]
+    let collectMatchesForContextID: (UUID, UUID) async -> [MCPContextBindingMatch]
     let collectMatchesForWorkingDirs: ([String]) async -> [MCPContextBindingMatch]
     let collectActiveMatchForWindowID: (Int) async -> MCPContextBindingMatch?
     let existingWindowIDForConnection: (UUID) async -> Int?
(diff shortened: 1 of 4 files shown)


baron (approved):
Approved. The patch aligns one-shot context routing with active-workspace authority, preserves explicit bindings across ordinary workspace switches, and adds appropriate focused regression coverage.

Non-blocking follow-up note: if an explicitly bound workspace is deleted while the MCP connection remains alive, and another active window contains the same context UUID, the resolver may fall back to that window rather than fail closed. This requires an unusual combination of conditions and does not block the observed regression fix, but it would be worthwhile to harden the stale-binding/deletion path separately.

PR #982, opened 2026-09-07
Preserve file authority across context binding

### LINKED ISSUE #962: Successful bind_context can be followed by an empty root projection after app restart
## Impact

Immediately after RepoPrompt CE restarts, an MCP client can receive a successful explicit `bind_context` response for a configured workspace, then have the next file tool resolve that same context as an empty workspace. This breaks file discovery and downstream Context Builder or Agent workflows even though the client was told the bind succeeded.

## Environment

- RepoPrompt CE debug app on macOS and Apple silicon
- Fresh bundled `rpce-cli-debug` client
- Configured workspace with one or more loaded roots

## Reproduction

1. Configure a workspace with at least one root and restart the RepoPrompt CE debug app.
2. As the restored workspace becomes available, connect a fresh CE MCP client.
3. Bind the client to the intended context with `bind_context`.
4. Immediately call `get_file_tree` with `type: "roots"`.

## Expected behavior

A successful bind establishes usable file-tool authority for the selected context. The following root query returns the configured roots.

If restoration is not ready within the bounded wait, the bind returns a typed, retryable readiness failure and preserves any existing usable binding.

## Actual behavior

`bind_context` reports success, but the following root query reports zero roots and “No workspace loaded.” A later request can succeed after workspace restoration finishes.

The failure occurs in the post-restart readiness window: the logical context is available before its root catalog and binding projection are ready, and that incomplete state is exposed as a valid empty scope.

## Acceptance criteria

- An explicit bind during workspace restoration either succeeds with usable root authority or returns a typed, retryable readiness failure.
- A successful bind and the immediately following file-tool request use the same workspace and root projection.
- A configured workspace waiting for root restoration is never reported as an authorized zero-root workspace.
- A genuinely rootless workspace remains distinguishable from a workspace whose root projection is not ready.
- Failed or superseded binds do not replace an existing usable binding.
- Readiness waits and cancellation remain bounded and leave no retained routing or provider work.
- Deterministic coverage exercises immediate post-restart binding, delayed root restoration, supersession, cancellation, and genuinely rootless workspaces.
- A packaged CE live check verifies that the first root query after a successful post-restart bind returns the configured roots.

## Related reports

- #944 concerns root-removal persistence and projection convergence. This report concerns initial post-restart restoration exposing a configured context as rootless.
- #862 concerns persisted numeric window IDs routing to the wrong workspace after restart. This report concerns a correct explicit context bind exposing incomplete root authority as an empty workspace.
- PR #889 concerns qualified reads stalling on unrelated roots and replay-token identity. This report concerns post-restart readiness and root projection for basic file tools after a successful bind.
- #949 concerns an admitted `get_file_tree` request exhausting the outer watchdog. This report returns promptly with an incorrect zero-root result, so the failure stage and response contract are different.

mplibunao:
## Summary

- Freeze file-tool authority when an explicit context bind succeeds, so the immediately following file request uses the same workspace and root projection.
- Treat configured workspaces whose root catalogs are still restoring as retryable readiness failures instead of authorized empty scopes.
- Preserve prior usable bindings across failed or superseded binds while keeping genuinely rootless workspaces distinct.
- Bound readiness waits, cancellation, and physical file-provider settlement.

## Root cause

Context binding could publish the logical workspace before its root catalog and file-tool projection were ready. The bind therefore reported success, while the next file request independently resolved the incomplete projection as an empty workspace.

## Verification

- Binding authority suite — 34 tests passed
- Context Builder workspace-resolution suite — 4 tests passed
- MCP domain read-provider suite — 4 tests passed
- Full Swift test suite passed
- SwiftFormat and strict SwiftLint passed
- `RepoPrompt` and `repoprompt-mcp` product builds passed
- Repository guardrails and contribution preflights passed
- Packaged live checks passed for atomic post-restart roots, worktree-scoped search, exact worktree identity, and bounded cancellation with immediate tree responsiveness and zero retained settlement debt

Closes #962

### DIFF
--- a/Sources/RepoPrompt/Features/AgentMode/Services/AgentWorkspaceLookupContextResolver.swift
+++ b/Sources/RepoPrompt/Features/AgentMode/Services/AgentWorkspaceLookupContextResolver.swift
@@ -45,6 +45,11 @@ struct AgentWorkspaceLookupContextSource: Equatable {
         )
     }
 
+    /// Canonical routed contexts have authority without an Agent binding identity.
+    var authorityIdentity: AgentWorkspaceLookupContextIdentity? {
+        activeAgentSessionID == nil ? nil : identity
+    }
+
     static func worktreeBindingFingerprint(_ bindings: [AgentSessionWorktreeBinding]) -> String {
         worktreeBindingFingerprint(.hydrated(bindings))
     }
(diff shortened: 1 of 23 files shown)


mplibunao:
Thanks for the updates, and sorry I haven't been able to update the PR myself. I reviewed the latest commits and found one gap they don't cover: file tools called with a one-shot tab hint on a connection that never ran `bind_context` still fail with `workspace_authority_superseded`. For example, `rpce-cli -w 1 -c read_file`, or any MCP client that sends `_windowID` or a per-call `context_id` without binding. `get_file_tree` reports 0 roots on that path.

`requiredFileToolLookupContext` passes `metadata.connectionID` into `fileToolLookupSnapshotIsCurrent` and the hydration check, and both require a stored `tabContextByConnectionID` entry, which a hint route never creates. The read-file auto-selection enqueue has the same requirement and throws on it.

One option: skip only the connection check for hint routes that carry no run (`.explicitHint` with `runID == nil`), keep their live-tab and root-catalog checks, and have those routes skip read-file auto-selection. Hints that carry a run would keep the connection check, so they still fail closed.

baron:
Thanks for catching this. I reproduced the unbound one-shot hint path and fixed it in `13eb2635` (now at PR head `cc80a5ff`). Only runless explicit hints bypass the absent connection-binding check; live tab/root-catalog authority still applies, and those routes skip automatic selection. Run-carrying hints remain fail-closed. The exact-head binding suite (20/20) and watchdog suite (30/30) pass locally; the product build and hosted CI are still running. I have updated the PR description with the validation boundary.

PR #1005, opened 2026-09-13
Make the subagent wait configurable in Settings

crstdr:
Adds a **Default subagent wait** setting so you can choose how long an idle wait on a sub-agent lasts before it reports back — without ever delaying completions, questions, or your own steering.

## Why

When an orchestrator waits on a sub-agent, an otherwise-quiet wait eventually times out, returns a snapshot, and the orchestrator burns an inference round deciding to wait again. With a two-minute default that is thirty rounds an hour of "still running?" — expensive on a high-tier orchestrator model, and most of those rounds learn nothing.

But the right interval is a judgement call, not a constant. A hands-on orchestrator benefits from checking in during the work; a light-touch one mostly cares about completion. So this makes it configurable rather than picking a new fixed number.

## What changes

**A setting.** Settings → Agent Mode → Overview → **Sub-Agent Supervision** → **Default subagent wait**: 2, 5, 10, 20, 30, or 60 minutes. Factory default is **5 minutes**. Also readable and writable via the `app_settings` MCP tool (`agent_mode.subagent_default_wait_seconds`).

**Omitted timeouts use it.** When an `agent_run` / `agent_explore` caller omits a timeout on `start`, `wait`, or an opted-in `steer`, the wait resolves to the configured value instead of a hardcoded constant. Each call captures the preference once and freezes it, so changing the setting never reshapes a wait already in flight.

**Guidance, not a mandate.** Tool descriptions and workflow prompts no longer advertise a fixed default or model short waits in their examples. They point at the configured wait and say to shorten it for closer supervision or lengthen it for well-scoped independent work — deliberate checks stay explicitly allowed.

## What does not change

This is a **maximum** wait, not a notification delay. Completion, questions, failures, and parent steering still interrupt immediately.

Explicit timeouts, `timeout: 0` polling, `detach`, `wait: false` steering, and cancellation all behave exactly as before.

## CLI deadline alignment

`repoprompt-mcp` imposes its own client-side deadline, so a longer server wait would otherwise be cut off. Before an implicit blocking lifecycle call it now performs one bounded `app_settings` read, pins the resolved value into the request, and derives its deadline from it.

Failing to pin a preference is never a reason to abandon the caller's wait. Budget expiry, tool errors, and older servers without the setting all fall back to a compatibility guard; only caller cancellation and connection loss abort. The bounded read covers registration and send rather than the response alone, and `start` / `steer` deadlines include an allowance for the session setup that runs before the server begins its wait.

One adjacent fix came out of this: cancelling a task parked in the request send barrier now abandons its registration instead of waiting indefinitely for a delivery report that may never arrive.

## Validation

Full CI green on this head: all four Build and Test shards, Provider Tests, Sentry-enabled Build, Style, and Secret Scan.

Locally, against the exact PR content (not a personal branch): `RepoPrompt` and `repoprompt-mcp` both build, the full test suite runs clean, and the focused suites pass — `MCPExportWatchdogContractTests` (34), `AgentLifecycleExecutionContractTests` (4), `AppSettingsMCPServiceAgentModeSettingsTests` (9), `DirectHeadlessCompositionTests` (16), `AgentControlToolCardPresentationTests` (7), plus `OracleGroupBoundaryTests` (12) and `InteractiveREPLSingleShotRoutingTests` (4) for the shared send-barrier change. The generated canonical tool-definition snapshot was regenerated and matches.

**Still outstanding:** live end-to-end validation. Nobody has yet rebuilt the app, selected a non-default wait in Settings, and confirmed on a real sub-agent that an omitted timeout honours the setting while completion, questions, and steering still wake the wait immediately. Worth doing before merge.

## Review history

Reviewed in three passes — Claude Opus, Cursor Grok 4.6, and Codex GPT-5.6 Sol — each of which found defects the previous pass missed. Fixed along the way: a trap on out-of-range `app_settings` doubles, a preflight budget that did not cover send, preflight failures aborting instead of falling back, a routing race across the new suspension point, blank `op` strings bypassing preflight, missing setup headroom on `start`/`steer`, and the send-barrier cancellation hole above. Duplicated lifecycle operation rules were consolidated behind a single intent type.


baron (approved):
Approved after exact-current integration with merged #1009. The repair restores the existing 120-second factory/default behavior for absent or unsupported settings; 300/600/1200/1800/3600 remain explicit opt-in choices. The #1009 ACP parameter initializer remains intact. Local exact-head validation passed 47 focused tests with 0 failures, plus commit/push preflights and independent review. Non-blocking follow-up: if the MCP SDK or underlying transport ignores cancellation while app_settings is being sent, the preflight budget may not unwind until transport failure; this does not change ordinary request behavior and is not a merge blocker.

baron:
@aAtila exact repaired head is `f633ded4be6ac32eaa8c2943e3307fb870974194`, integrated with merged #1009/current main. Local exact-head validation: 47 focused tests, 0 failures; independent final review found no blocker. The repository requires post-push approval from someone other than the pusher, so please review when available. Hosted CI is running.

PR #1025, opened 2026-09-19
Fix composer cursor jumping during typing

jsmcx:
## Summary
- Prevent stale SwiftUI text bindings from overwriting active composer edits
- Preserve intentional external updates such as draft restoration, send clearing, and tagged-file removal
- Defer external synchronization during IME composition
- Add regression tests for focused text synchronization

## Validation
- `make dev-test FILTER=ResizableTextFieldTests`
- `make dev-lint`
- `make dev-build`
- `.agents/skills/rpce-contribution-check/scripts/preflight.sh pr-ready`


baron:
@jsmcx sorry, once this passes review we'll add you to contributors. my apologies.

baron (approved):
Reviewed exact head `6e6248f1b4a06b6b050effe89d6dc115065fa66e`.

The change correctly prevents stale SwiftUI binding snapshots from overwriting active native composer edits, while preserving intentional programmatic updates through the external revision tick. Draft restoration, submit clearing, tagged-file removal, and tab loading all advance that revision, and marked-text composition is deferred rather than overwritten. The latest complete CI rerun is green.

No blocking findings — approved.

Non-blocking follow-up: an AppKit-backed regression fixture that directly asserts typed order, caret position, undo preservation, and IME completion would strengthen the current policy-level tests, but I do not consider that required for this fix.

Issue #810, opened 2026-08-12
Context Builder header displays live settings instead of the active run’s frozen behavior


baron:
## Impact

Context Builder execution correctly freezes behavior when a run begins, but the running header continues to display live global enhancement/question settings. Edits from Settings or another window can therefore change what the UI reports without changing the active run.

For MCP-controlled runs, the question indicator uses the UI clarifying-question preference instead of the captured MCP permission.

## Affected version

Introduced by `0b4d738a`; still present at `a49cdfc1`.

### Cross-window drift

1. Start a Context Builder run using enhancement mode A and clarifying-question setting A.
2. In another window, change the global Context Builder settings to mode B/setting B.
3. Return to the running view.
4. Observe that the header reflects B while execution retains captured behavior A.

### MCP mismatch

1. Configure UI clarifying questions on and MCP clarifying questions off.
2. Start an MCP-controlled Context Builder run.
3. Observe the header/question indicator.

The UI can report questions enabled even though the captured MCP run does not permit `ask_user`.

## Expected behavior

While a run is active, every displayed behavioral value should come from the same immutable run snapshot used by execution.

When idle, the header may display live settings for the next run.

## Actual behavior

Execution snapshot:

- `Sources/RepoPrompt/Features/ContextBuilder/ViewModels/ContextBuilderRunLifecycle.swift:135-169`
- `Sources/RepoPrompt/Features/ContextBuilder/ViewModels/ContextBuilderAgentViewModel.swift:3677-3690`

The header receives only the active token budget from the snapshot; other bindings remain live:

- `Sources/RepoPrompt/Features/ContextBuilder/Views/ContextBuilderAgentView.swift:866-906`

The tooltip and icon render the live UI question value:

- `Sources/RepoPrompt/Features/ContextBuilder/Views/ContextBuilderAgentView.swift:1250-1262`
- `Sources/RepoPrompt/Features/ContextBuilder/Views/ContextBuilderAgentView.swift:1316-1319`

Cross-window store changes immediately republish the view model:

- `Sources/RepoPrompt/Features/ContextBuilder/ViewModels/ContextBuilderAgentViewModel.swift:960-968`

## Acceptance criteria

- Active-run header, tooltip, mode, budget, timeout, question permission, and follow-up status derive from the captured `ContextBuilderRunBehavior`.
- MCP runs display the captured MCP question permission, including inactive-target suppression.
- Global setting changes during a run affect only subsequent runs.
[...]

PR #1028, opened 2026-09-19
Show captured Context Builder behavior during active runs

bcgalvin:
Refs #810

**Summary**
While a Context Builder discovery run is active, the compose-tab header's enhancement mode and question readouts (title, mode tooltip, question indicator, settings tooltip, and the Auto follow-up status) read the live preference bindings, so changing a preference in another window could contradict the behavior the run had already captured for execution. The active token budget already used the capture (or "Unavailable") and is preserved. This change unifies all current-run readouts on the run's captured `ContextBuilderRunBehavior` through one read-only projection, keeps idle readouts on the desired next-run settings, shows an explicit unavailable state instead of falling back to live values, displays the effective (workspace-based) MCP question permission as captured, makes the Auto control tri-state (unavailable is never drawn as Off), and dismisses an open preferences popover when the displayed tab enters active discovery. MCP response type/model metadata remain separately owned and are kept when discovery settings are unavailable.

**Cause and scope**
Execution already captured behavior at admission and used it for the agent message, tool policy, and question timeout; only presentation authority was inconsistent. Commit 1 (`ContextBuilderAgentView.swift` + projection tests) adds the internal `ContextBuilderBehaviorPresentation` projection (idle / active / unavailable) and routes the header and current-run readouts through it. Commit 2 is tests-only. Capture timing, admission, registry ownership, settings persistence, permission policy, snapshot lifetime, and provider/model/response authority are unchanged; `Package.resolved` is unchanged.

**Validation**
- `ContextBuilderGlobalBehaviorSettingsTests` at `ed7f5128`: 16 tests / 0 failures (8 new projection regressions).
- `ContextBuilderMultiRootDiscoveryTests` at `97b820c5`: 27 tests / 0 failures, including three added methods: real UI and MCP admissions capture A, desired settings change to B before the agent message is built, the provider-bound message / run-bound timeout / presentation stay on A while desired getters show B, settlement by cancellation, idle B, and a second real admission capturing B; plus a driver cleanup regression.
- Second local `pr-ready` at `97b820c5`: 3013 tests / 0 failures, Swift lint, guardrails, RepoPrompt product build. The first full run had one failure in an unchanged AgentMode admission test (`ConcurrentSameWorkspaceAgentRunAdmissionTests/testCanonicalAdmissionRefreshRejectsDuplicateTabIDWithoutMutationOrDispatch`); an isolated rerun passed and the cause remains undetermined.
- Manual checks on a debug build of `97b820c5`: popover dismissal on admission and when switching to an already-running tab; active UI mode/question/Auto readouts retained after preferences changed from another window; cancellation returning idle readouts to the new preferences with no automatic follow-up; an active-workspace MCP run retaining captured permission and MCP response/model metadata. Both live journeys settled by cancellation; no discovery or Oracle response completed.

**Limitations**
- Not observed live: the active UI numeric budget and the header/question/Auto tooltips (a real question's countdown is not evidence of the header timeout summary), and an inactive-workspace MCP run displayed while active; inactive suppression has factory/projection test coverage, not an end-to-end routing demonstration. Unavailable-capture rendering was not exercised live.
- An earlier live attempt stopped at a provider capacity error before any settings change and is not evidence either way.
- The validation session's normal debug-app shutdown timed out and required forced termination; the cause and its relationship to this change are undetermined (lifecycle code is unchanged here, which does not prove independence).
- This bounded pre-publication refresh found no evidence establishing an integration requirement against current `main`; compatibility beyond disjoint changed paths is not proven.


baron (approved):
Reviewed at dc18d4f79517cd54159edb3f3a7fe00b774c7095.

The active/idle Context Builder presentation authority is coherent, preserves captured UI and MCP behavior, and keeps response/model metadata separately owned. Focused projection and real-admission coverage passed locally; no correctness blockers found.

The currently red root shard fails in an untouched Oracle-group test and passed on focused rerun, but the required CI check should still be rerun to green before merge.

PR #1029, opened 2026-09-20
Fix Devin ACP permission gaps: strict MCP auto-approval and configured launch mode

dsebban:
## Summary

Implements the fixes from `docs/investigations/devin-acp-approval-prompts-2026-09-19.md`. Devin was excluded from strict RepoPrompt-MCP auto-approval, and both unattended entry points hardcoded `--permission-mode auto`, ignoring an explicitly configured Full Approval.

- **Strict per-request auto-approval for Devin** in `ACPAgentSessionController.autoApprovalSelection`, restricted to allow-once options and matched on `rawInput` server/tool identifiers. Fixes RepoPrompt MCP tool prompts (e.g. context_builder) under Safe Managed.
- **Devin option denylist** in `ACPPermissionOptionPolicy` — `switch_*`/`plan_*` mode switches (would drift the launch-mode reuse key in `isCompatibleWith`) plus `*_global`/`all_fetches`/server-wide/`*_always` grants are never auto- or fallback-selected, enforced by both explicit IDs and pattern rules for unlisted variants (e.g. `switch_smart`, `net_grant_always`).
- **Kind-aware allow fallback** — an accept decision can no longer submit a `reject_*` or empty option ID when only denylisted allows were offered; the prompt is cancelled instead. Session-scoped decisions never widen into the persistent `allow_always` grant (`exceedsSessionScope`).
- **Configured level threaded into unattended launches** via `PermissionLevel.unattendedCLIPermissionMode`: Full Approval → `dangerous`, every other level → `auto` (unattended runs can't surface prompts, so only the explicit maximum grant escalates). Applies to `DevinACPHeadlessAgentProvider.makeRunRequest` and the `DevinCLIProvider` one-shot; model discovery (no injected MCP server) keeps the provider default.
- **Pending-prompt settlement** — escalating to Full Approval while a Devin prompt is pending settles it with `.acceptForSession` (→ `allow_session`), never `switch_bypass`, and never mutates the running process's launch flag. Settings copy now notes Full Approval also covers unattended runs.

#### Test plan
- [x] `swift build --product RepoPrompt` — clean
- [x] `make dev-test FILTER=DevinPermissionLevelTests` — 39 tests, 0 failures (incl. denylist explicit IDs + unlisted pattern variants, denylisted-`allow_once` disguise, strict auto-approval, session-fallback, unattended-mode, and CLI argv coverage)
- [x] `make dev-test FILTER=GrokBuildPermissionAndIdentityTests` — 6 tests, 0 failures (shared policy enum)
- [x] `make dev-lint` — SwiftFormat + SwiftLint strict clean
- [x] Oracle review (2 reviewers); P1 findings addressed (kind-aware fallback, pattern deny rules, session-scope cap, settings copy)

Generated with [Devin](https://devin.ai)


crstdr:
Hi — testing this on my side, so a few data points that might be useful.

**Yes, I reproduce the approval prompts in full approval mode.** Running an agent with `Permissions · Full Approval` (which the settings panel describes as launching Devin with `--permission-mode dangerous`), a plain `get_file_tree` call still stopped for approval. Model was `deepseek-v4-1-flash-high`, launched via `agent_run`, so this isn't specific to interactive use.

**On `preferredAllowOptionID` — I looked at this independently last week and reached the same design you've landed on here.** The motivating case is that Devin sends six options for one command, of which **three share `kind: "allow_always"`**:

| optionId | kind | name |
|---|---|---|
| `allow_once` | allow_once | Allow |
| `allow_session` | **allow_always** | Yes, allow `python3` commands (this session) |
| `allow_always` | **allow_always** | Yes, always allow `python3` commands in `<project>` |
| `allow_always_global` | **allow_always** | Yes, always allow `python3` commands in all projects |
| `switch_bypass` | **allow_always** | Yes, switch to bypass mode |
| `reject_once` | reject_once | Reject |

Matching on `kind` alone meant "Always Allow" could resolve to `allow_always_global` or `switch_bypass` depending on array order — so your `exceedsSessionScope` denylist plus dropping the `?? ""` fallback looks right to me, and the `String?` return is the important half.

**One adjacent gap, and I want to be clear it's pre-existing on `main`, not something this PR introduced** — but you're one file away from it and it undercuts the work here. The option `name` is parsed and discarded:

```swift
// ACPAgentSessionController.swift:1757-1763
guard
    let optionID = optionDictionary["optionId"] as? String,
    let kind = optionDictionary["kind"] as? String
else { return nil }
return PermissionOption(optionID: optionID, kind: kind)   // `name` dropped
```

`PermissionOption` (line 199) only carries `optionID` and `kind`, so the UI can't label anything. The result is that `approvalDetails` dumps the raw options array as a JSON blob (line 3353) and the user gets three generic buttons instead of your six clearly-worded ones. Other providers send two or three terse options so nobody noticed; Devin is the first provider where it's obvious.

Not asking you to take that on in this PR — just flagging it since you're deep in this code and it's the other half of the same user-facing problem.

dsebban:
Thanks for the review, I will be off for a couple of days and I see your agent already knows what to do, so if you don't mind and if you have time ofc, could you take over the branch and push the fix you mentioned to this branch and then double check the approvals issues are gone and then make the PR ready and request that it's merged in the channel . If you don't have time or can't spare the tokens I understand. 🙏

crstdr:
Quick update for anyone testing or reviewing this: the readable option labels in `937ca376` are verified locally, but a plain `get_file_tree` call still prompts with **Full Approval** selected.

We have two leads:
- Devin's log records `AcceptEdits` for a prompting call, despite launches specifying `--permission-mode dangerous`. We're checking how the effective mode is established during ACP session setup.
- RepoPrompt's permission-request path may miss Devin's machine-readable tool identity in metadata. We're confirming the actual permission-request payload before changing the matching logic.

These are promising findings, not a fully confirmed root cause yet. I'll follow up with a narrowly scoped fix, local validation and regression coverage, then update this branch. Keeping the PR in draft until the remaining approval behaviour is verified.

